PR #289, opened 2014-04-02
Add additional config to the locations resource and fix set_header in vhost resource

b4ldr:
Following parameters where added to the location resource.  These
already existed in the vhost resource so the same logic and defaults
where used
- proxy_set_header
- proxy_connect_timeout
- proxy_redirect

The set_header variable expectes a hash of headers to set.  Puppet is
does not allow hash keys to have a hyphen in the name or begin with an
capital leter.  As (i belive) all headers start with a
capital letter this limitation means you can not set any header.
further more the majority of headers also contain an underscore.

I have included a patch which allows a key to be passed all in lower case and
use underscores in place of hyphens.  The template will then make the
conversion.  e.g. to set the header Accept-Charset: utf-8.  one would
pass

set_header => { accept_charset => 'utf-8' }

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -20,8 +20,14 @@
 #   [*proxy*]                - Proxy server(s) for a location to connect to.
 #     Accepts a single value, can be used in conjunction with
 #     nginx::resource::upstream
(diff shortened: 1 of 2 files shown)


ZhukV:
:+1:

PR #367, opened 2014-06-27
Add all gzip params for http context

richdel:
Added all the available gzip params for the nginx.conf template and blank defaults which won't get picked up if not set therefore not written to .conf and nginx will use internal defaults. gzip_types param can be either a string or array.


jg-development:
+1 ... will this be merged?

3flex:
Not as is... there are merge conflicts and a few updates to be made. If someone wants to create an updated PR we can look at merging that instead, if @richdel doesn't want to update this.

PR #398, opened 2014-08-12
Add uwsgi support

mvintila:
Add uwsgi support, very similar to fastcgi. Actually most of the code is just copied over from the fastcgi functionality. This is tested and currently running in a production environment.


mvintila:
Tests are fixed, would be nice if this was merged eventually.

jfryman:
@mvintila thanks for the code.

Yeah, I'm good with this. I know it can be done with custom directives, but this is cool to merge.

3flex:
Why doesn't uwsgi_params match upstream? Was was committed contains this extra line:

`uwsgi_param SERVER_ADDR $server_addr;`

mvintila:
What do you mean? The file is identical to the example here:
http://uwsgi-docs.readthedocs.org/en/latest/Nginx.html

3flex:
This is the current file in nginx source: http://trac.nginx.org/nginx/browser/nginx/conf/uwsgi_params

You wouldn't include `uwsgi_param  HTTPS              $https if_not_empty;` from upstream because that's only supported in nginx 1.1.1 and higher, but otherwise I'd expect the file in this module to match upstream.

3flex:
@jfryman what do you want to do here? I'm trying to avoid a repeat of https://github.com/jfryman/puppet-nginx/pull/500

PR #402, opened 2014-08-14
Added add_header to nginx::resource::location

chaosmail:
I made add_header available also in location, and added a test for add_header in vhosts

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -15,6 +15,8 @@
 #   [*location_deny*]        - Array: Locations to deny connections from.
 #   [*www_root*]             - Specifies the location on disk for files to be
 #     read from. Cannot be set in conjunction with $proxy
+#   [*add_header*]          - Hash: Adds headers to the HTTP response when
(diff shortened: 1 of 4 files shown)


cyberious:
Any possibility of writing acceptance tests as well for this?

chaosmail:
Its been 2 years i submitted this PR :) think this is slightly outdated

wyardley:
@chaosmail @cyberious: Does anyone want to update this and resolve the conflicts? As @chaosmail says, the PR is pretty out of date now.

If there are no more changes or comments in the next month, we may close this for inactivity.

wyardley:
There was also some discussion in #899  about supporting the 'always' directive. I'm not sure how that would be implemented, would probably have to be a hash with an option of having an array as the hash value. For now, I'm going to close this out (I think people can use the location append / prepend features if necessary), feel free to open a new PR if someone wants to implement this. I think it would be smart to do it as above, though it would probably be a good idea to add the same functionality backwards (in a backwards compatible way) to vhost as well.

PR #421, opened 2014-08-22
added ssl_session_timeout as vhost parameter

fpizzurro:


3flex:
Fixed in https://github.com/jfryman/puppet-nginx/commit/50f9e19770c5b856757a9173508d66e9f1ad8d68

PR #431, opened 2014-09-02
Support for RHEL/CentOS 7 which uses FHS-3.0. 

esbjerg:
FHS-3.0 dictates that pid files lives in /run.

NB. This is a suggestion for a change. If the code I have written does not fit well feel free to change it around or suggest changes.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -78,6 +78,10 @@ class nginx::params {
 
   $nx_pid = $::kernel ? {
     /(?i-mx:linux)/   => $::osfamily ? {
+        /(?i-mx:redhat)/ => $::operatingsystemmajrelease ? {
+            /7/     => '/run/nginx.pid',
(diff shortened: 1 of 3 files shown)


3flex:
I've reviewed the Red Hat documentation at https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Migration_Planning_Guide/sect-Red_Hat_Enterprise_Linux-Migration_Planning_Guide-File_System_Layout.html

It says
> Applications can now use /run the same way they use the /var/run directory.

As the directories are bind mounted together it's not important from this module's perspective whether the file gets created in /var/run or /run. Once created it will be in both locations anyway.

In future if a complete migration occurs and all PIDs are created in /run we can update this module accordingly.

A new PR should be opened for the `fastcgi_index` addition.

Cheers!

PR #451, opened 2014-09-20
(maint) Better docs

3flex:
Moves all docs to the README, using the standard module documentation template recommended by Puppetlabs.

All public classes and defined types now say "refer to README" in the documentation section. All private classes had the documentation comments removed completely.

First step to fixing https://github.com/jfryman/puppet-nginx/issues/401


wyardley:
@3flex: totally support this idea! The fact that there are currently no docs for the nginx and nginx::config is terrible! Not sure what the plans are in terms of using puppet-strings or similar things for self-documentation, but sounds like it's not totally ready.

Is there anything I or "we" can do to help with improving the docs? Is the work in this PR still useful with a rebase, or is it so old that re-using it is going to be hard?

3flex:
I see many conflicts; updating this is probably more trouble that it's worth.

IMHO we should use puppet-strings as the documentation based going forward, making most of this PR redundant.

I think the best way to improve docs is to open a new PR to format the existing inline documentation so it's compatible with puppet strings 0.99.0 which is likely to drop very very soon, and use puppet-strings for documentation going forward (I think this is voxpupuli's direction, see https://github.com/voxpupuli/plumbing/issues/19)

Not sure what the best approach is for adding documentation to nginx and nginx::config classes, other than someone digging in to document it all...

wyardley:
I agree on both counts.

If the version of puppet strings that allows inline docs is really imminent, I'm totally in favor.

I'm going to leave this open for now simply for ease of finding it -- even with all the conflicts, it should still be a good starting point for the descriptions of the various knobs / whistles, at least the ones that existed when this was written.

wyardley:
Opened #941 for this

PR #471, opened 2014-10-07
issue 470 allow creation of location with only try_files

### LINKED ISSUE #470: cannot create location with only try_files defined
The check for creating a resource does not check to see if try_files is defined which alone could be used to create a valid location reference.

For example in wordpress if you want to forward rewrites to index.php you would need the following

    location / {
	  try_files $uri $uri/ /index.php?$args;
    }

buckstr:
This commit will fix #470.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -308,8 +308,8 @@ define nginx::resource::location (
   if ($vhost == undef) {
     fail('Cannot create a location reference without attaching to a virtual host')
   }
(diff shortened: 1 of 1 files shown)


3flex:
@buckstr the proposed PR won't allow you to create a location with only `try_files`.

As is, if you only set `try_files` and don't set `proxy`, `location_alias`, `stub_status`, `fastcgi` or `www_root` then the template `vhost/locations/empty.erb` will be rendered. This doesn't include `try_files`.

wyardley:
Hi --
There's been no further activity on this since the 'needs Rebase' and 'needs tests' labels were set. If we don't have any further comments in the next month, this PR will be closed.

wyardley:
Addressed in #834 already.

PR #487, opened 2014-10-21
Allow multiple proxy_cache_paths and allow users to be more specific.

actown:
In the past the module has only allowed for one proxy_cache_path to be
set. I might have a use case coming up where I will need a few caches
with different requirements so I wanted to be able to define both of
those within puppet. Another thing about breaking this into a hash
instead of using settings for each option is that now the user can
define the minimum options which are path and keys_zone as defined in
the docs [1].

The only thing I can think that this is missing is if we'd want the
option to be called proxy_cache_path still as then its a 1 to 1 matching
to the option within nginx. I only changed it because it does allow a
hash now and is a collection of proxy_caches. I can easily change this
back and make a new PR if that's what we want to go with.

[1] http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_path

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -27,11 +27,7 @@ nginx::config::proxy_set_header:
   - 'Host $host'
   - 'X-Real-IP $remote_addr'
   - 'X-Forwarded-For $proxy_add_x_forwarded_for'
-nginx::config::proxy_cache_path: false
-nginx::config::proxy_cache_levels: '1'
-nginx::config::proxy_cache_keys_zone: 'd2:100m'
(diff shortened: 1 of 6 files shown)


wyardley:
I see some mentions in #529 (also not merged yet). Is this going anywhere, or should we close it? At the least, it needs a rebase. @actown: do you want this one to be closed, or do you want to rebase it and resolve the conflicts?

actown:
If it seems like a worthwhile change I could attempt a rebase, but I haven't looked at this code in a while so I'm leaning towards just closing the PR. Thoughts?

PR #500, opened 2014-11-08
fastcgi: sync fastcgi_params with upstream fastcgi.conf

### LINKED ISSUE #499: Issues with fastcgi_params
Hi!

The shipped `fastcgi_params` file is not generic enough. It would be nice to change `SCRIPT_FILENAME` to `$document_root$fastcgi_script_name` for better flexibility. However, this change may be disruptive. This is how the file shipped with nginx is.

Currently, `SCRIPT_FILENAME` is not correct for many setup and it may not be overridable because the order of the `fastcgi_param` directives are important. For stuff like `php-fpm`, the last entry wins, so it is easy to override `SCRIPT_FILENAME`. However, for `fcgiwrap`, this is the first one that will win.

vincentbernat:
This is a risky change for `SCRIPT_FILENAME` but it handles more
situation than fastcgi_params. Upstream still ships the previous
fastcgi_params file.

Fixes #499 

As explained in #499, the change is disruptive. I am doing the pull request for comments only. Only `SCRIPT_FILENAME` is changed (and `HTTPS` a bit).

### DIFF
--- a/templates/vhost/fastcgi_params.erb
+++ b/templates/vhost/fastcgi_params.erb
@@ -1,27 +1,26 @@
 # This file managed by puppet on host <%= @fqdn %>
 
-fastcgi_param	QUERY_STRING		$query_string;
-fastcgi_param	REQUEST_METHOD	  	$request_method;
-fastcgi_param	CONTENT_TYPE		$content_type;
-fastcgi_param	CONTENT_LENGTH	  	$content_length;
+fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;
+fastcgi_param  QUERY_STRING       $query_string;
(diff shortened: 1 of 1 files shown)


wyardley:
@3flex: what are your thoughts about this one?

3flex:
I think it's a bigger question than whether to merge or not merge... I think we need consensus on https://github.com/voxpupuli/puppet-nginx/issues/862 before proceeding.

wyardley:
Decision on #862 has basically been made, is someone willing to rebase this PR? I know it's ancient now...

wyardley:
I like this idea, but no response in the comments, so I'm going to close this for now; would welcome a new or updated PR that does the same thing.

PR #545, opened 2015-01-09
Prevent missing resource errors if custom configuration is used without default location

SteveMaddison:
Could possibly be simplified by passing these four parameters directly to the resource definition for the default location.


3flex:
@SteveMaddison would you be able to simplify as you suggested? We can merge as is but there's no reason to have the overrides here instead of setting the parameters directly on the default resource as you suggest.

SteveMaddison:
Sure, I thought I must've been overlooking something so just went for a minimal change. New commit pushed with all params going directly to the resource.

3flex:
Actually this might need to be rebased after https://github.com/jfryman/puppet-nginx/pull/529 is merged... I think @jfryman wants to hold off on changes to the code that will be affected.

jfryman:
Yeah, this will probably need to be re-based. I'm not traveling this week, so I'll aim to finish up #529 this week.

3flex:
@SteveMaddison could you rebase, and squash this to one commit? We'll get it merged.

SteveMaddison:
I Think that's it, have a good look. CI failed but it appears to be unrelated.

3flex:
Looks good! And you're right, the CI issue isn't related.

Could you please squash it to a single commit to keep the history clean? There's a helpful guide here if you're not sure how to do it: https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

SteveMaddison:
Ended up resetting the branch but this looks like it has the same effect.

3flex:
Looks much better, thanks for the code!

PR #546, opened 2015-01-13
allow listen_ip and ipv6_listen_ip to contain a String or Array

b4ldr:
This allows one to pass multiple ipv4/ipv6 addresses to a specific vhost.  I also added ipv6 and a few ssl options to the www redirect block.


b4ldr:
As far as i can tell the test in spec/defines/resource_mailhost_spec.rb would have been failing before this change.

jfryman:
Awesome. Sorely needed. Thanks for this.

PR #549, opened 2015-01-19
do www-rewrite with params

paschdan:


jfryman:
:cool:! Thanks for the code.

PR #551, opened 2015-01-23
sort add_header values for vhost

sbaryakov:


jfryman:
:+1: Thanks for the contribution!

PR #552, opened 2015-01-23
Allow to use OpenBSD specific service_flags and package_flavors.

buzzdeee:
Give example for passenger in README.md and add OpenBSD to the
supported OSs in metadata.json.


jfryman:
Neat! Thanks for the code!

PR #553, opened 2015-01-24
closes #541 (maybe: needs user feedback)

### LINKED ISSUE #541: Symlink happening after service refresh
It appears that the symlink from sites-available to sites-enabled is happening after the service gets refreshed.  This is problematic as we need to bounce nginx manually (unless we make another subsequent change to the configs and re-run puppet)

steakknife:
Perhaps there's a subtle gotcha in how Puppet processes conflicting notifications to contained resources.


jfryman:
I'm not sure if this closes the loop, but it's certainly better to call the containing class vs the resource directly. So, :+1:

PR #557, opened 2015-01-29
Set up relationships for nginx::config even when overridden

radford:
This keeps the user from having to duplicate them when defining nginx::config externally.


jfryman:
:metal:

PR #560, opened 2015-02-03
changed $::operatingsystemmajrelease to $::lsbmajdistrelease for Debian

janschumann:
The fact operatingsystemmajrelease does not exists in facter < 1.7 wich is installed by default on debian wheezy. That comes along with puppet 2.7.23 wich should be supported. The resolution is to use lsbmajdistrelease fact.

I only adopted the code for debian, just to here what you think. If you agree, I would also update the code for redhat etc.


jfryman:
Yeah, I'm down with this. If you wanna PR for RH too, that would be most appreciated.

3flex:
@jfryman do you want to revert this? See discussion at https://github.com/jfryman/puppet-nginx/pull/562

PR #561, opened 2015-02-05
Sort fastcgi params to have stable ordering

mlafeldt:
The hash sorting was started in #532 and this fixes it for fastcgi as well.

/cc @s0enke


jfryman:
:cool:

mlafeldt:
Thanks, looking forward to the next release. Any plans on when that's going to be?

PR #562, opened 2015-02-06
Addition to #560: use lsbmajdistrelease instead of operatingsystemmajrelease

janschumann:
Added fix for redhead

### DIFF
--- a/manifests/package/redhat.pp
+++ b/manifests/package/redhat.pp
@@ -33,7 +33,7 @@ class nginx::package::redhat (
     case $package_source {
       'nginx', 'nginx-stable': {
         yumrepo { 'nginx-release':
-          baseurl  => "http://nginx.org/packages/${_os}/${::operatingsystemmajrelease}/\$basearch/",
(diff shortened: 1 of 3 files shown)


3flex:
Be wary of merging this. The LSB facts require a certain package (redhat-lsb) to be installed on RH systems which pulls in several dependencies which is not desirable on cloud/minimal servers. When that package is not installed the fact will be undefined (I'm not sure if a better fallback is available in newer Facter versions but historically this was the case).

Facter 1.7 is a dependency of this module according to the README for the very reason that the `operatingsystemmajrelease` fact is required, so the redhat-lsb package doesn't have to be installed.

I don't think the situation is as bad on Debian since the required LSB package is smaller and has fewer dependencies so https://github.com/jfryman/puppet-nginx/pull/560 probably won't be much of an issue. But this PR needs more thought.

janschumann:
Right. Facter >=1.7 is a requirement. And this will be installed through the puppet deb repo for wheezy!

So #560 should be reverted and this PR not merged in. Sorry!

PR #569, opened 2015-02-12
Flags parameter supported only on OpenBSD

Zophar78:
Flags parameter is only supported on OpenBSD and breaks on some other systems like RH (on puppet 3.4.x)
See the discussion here: https://github.com/jfryman/puppet-nginx/commit/2960c10cd

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -41,7 +41,7 @@ class nginx::service(
         ensure     => $service_ensure_real,
         name       => $service_name,
         enable     => $service_enable,
-        flags      => '',
+        flags      => $service_flags,
         hasstatus  => true,
         hasrestart => true,
       }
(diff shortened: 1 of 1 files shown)


manifests/service.pp
```diff
@@ -35,14 +35,28 @@
...
-    flags      => $service_flags,
-    hasstatus  => true,
-    hasrestart => true,
+  case $::osfamilly {
+    'OpenBSD': {
+      service { 'nginx':
+        ensure     => $service_ensure_real,
+        name       => $service_name,
+        enable     => $service_enable,
+        flags      => '',
```
buzzdeee: Haven't tested, but this line should be:

 flags => $service_flags,

otherwise looks exactly as what I would have done.

Zophar78:
Will do a branch next time (i did that a bit too fast)

jrgarcia:
:+1: I'd love to see this get pulled in. This module no longer works on our CentOS systems.

jrgarcia:
Thanks so much!

PR #570, opened 2015-02-12
update yumrepos to use $releasever

bastelfreak:
The yum team introduced this feature several years ago and it is currently their prefered way for setting up the release number

### DIFF
--- a/manifests/package/redhat.pp
+++ b/manifests/package/redhat.pp
@@ -33,7 +33,7 @@ class nginx::package::redhat (
     case $package_source {
       'nginx', 'nginx-stable': {
         yumrepo { 'nginx-release':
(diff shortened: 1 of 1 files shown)


3flex:
According to Red Hat docs

> The value of $releasever typically consists of the major release number and the variant of Red Hat Enterprise Linux, for example 6Client, or 6Server.

but `6Client` would not match the directory tree (see http://nginx.org/packages/rhel/) since it only has directories named 5, 6, and 7.

Did you test this works? While it may work for CentOS (I'm not sure) according to the docs it won't work for Red Hat.

bastelfreak:
Ah interesting, I don't have any RHEL around to test it, but this works perfectly on CentOS 6,7 and CloudLinux.

3flex:
Closing. Unless the current code is not working correctly on a given distribution I'm happy to set the yumrepo path like this.

PR #574, opened 2015-03-05
Prepend to the nginx config block

prachetasp:
#### What's this PR do?
Adds functionality and tests for prepending key/value pairs to the top of the nginx configuration. Tests mimicking the http_cfg_append tests have been added since this functionality is identical albeit in a different place. All rspec tests are currently passing. Furthermore, this change is currently working on a CentOS 6.6 (Final) staging box at a client site. I did not add to the README documentation since there doesn't seem to be a good place to put this. Let me know if there is somewhere you want me to add it.

#### Where should the reviewer start?
Begin in the manifests/config.pp file with the $nginx_cfg_prepend variable validation. Then check the templates/conf.d/nginx.conf.erb for the added nginx_cfg_prepend block.

#### How should this be tested?
By running `bundle exec rake spec` for the unit tests. Tests have been added for this change.

#### Any background context you want to provide?
At Railsdog we support several sites that run Nginx/Passenger and Rbenv on RHEL derivatives. One of the requirements is to feed environment variables to the Passenger process via the Nginx config. This is easily accomplished by using http://nginx.org/en/docs/ngx_core_module.html#env. The only issue is that this variable can only occur at the top of the main nginx config block. Figured others could use this feature as well!

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -62,6 +62,7 @@ class nginx::config(
   $multi_accept                   = 'off',
   $names_hash_bucket_size         = '64',
   $names_hash_max_size            = '512',
+  $nginx_cfg_prepend              = false,
   $proxy_buffers                  = '32 4k',
   $proxy_buffer_size              = '8k',
   $proxy_cache_inactive           = '20m',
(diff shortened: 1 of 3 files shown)


jfryman:
Nice. Thank you very much for the code! :bow:

PR #575, opened 2015-03-10
Phusion changed the directory for passenger

lukesmith:
I'm assuming this module was always installing passenger 4 previously? Phusion have [changed the location](https://blog.phusion.nl/2015/03/08/passenger-4-apt-repository-now-available/) of the passenger 4 apt repository which breaks installing nginx with passenger.

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -50,7 +50,7 @@ class nginx::package::debian(
       }
       'passenger': {
         apt::source { 'nginx':
-          location          => 'https://oss-binaries.phusionpassenger.com/apt/passenger',
+          location          => 'https://oss-binaries.phusionpassenger.com/apt/passenger/4',
           repos             => 'main',
           key               => '561F9B9CAC40B2F7',
           key_source        => 'https://oss-binaries.phusionpassenger.com/auto-software-signing-gpg-key.txt',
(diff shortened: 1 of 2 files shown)


3flex:
Is this module incompatible with Passenger 5? If not, I'd be inclined to install the current version from the apt repository rather than keep everyone on 4.

If you need 4, I'd suggest adding a new parameter to allow people to choose between 4 and 5 instead of hard coding it.

lukesmith:
> Is this module incompatible with Passenger 5? If not, I'd be inclined to install the current version from the apt repository rather than keep everyone on 4.

I've no idea if the module is compatible with Passenger 5. We hit this issue when we were in the process of provisioning a new machine and couldn't, not even 5 would be installed as I suspect the url would need to change to `/latest`. I expect anyone using this module would encounter an issue without this change?

Agree that if the module is compatible with Passenger 4 or 5 then it should be configurable. I want repeatability in provisioning a new machine and wouldn't want a new version to be installed without opting in to it.

3flex:
Looking at the docs the repo location is still https://oss-binaries.phusionpassenger.com/apt/passenger

Though it looks like it should be `/latest` or `/4` when you visit that URL, according to discussion at https://github.com/phusion/passenger/issues/1458 they do an internal redirect from /apt/passenger to /apt/passenger/latest

If you can change the PR so it gives the option to install 4 or latest then I can merge that in, but I don't want to hard code the older version's repo.

3flex:
I'm closing this for now, as the original PR isn't suitable for the module.

If you'd like to add a new variable that allows you to specify the path then that may be accepted. Alternately you could set `manage_repo => false` in this module and declare the `apt::source` resource in your own custom manifests.

PR #578, opened 2015-03-17
feat: add an easy maintenance page support

brunoleon:
This adds a parameter to easily set a vhost in maintenance mode.


jfryman:
:cool: Thanks for the code!

PR #579, opened 2015-03-19
apt::key: puppetlabs-apt check now the full GPG fingerprints.

sbadia:
Since commit 531ef9 of puppetlabs-apt, the gpg key (apt::key) is
checked, and should use the long format.

This value is validated with a regex enforcing it to only contain valid
hexadecimal characters, be precisely 8 or 16 hexadecimal characters long
and optionally prefixed with 0x for key IDs, or 40 hexadecimal
characters long for key fingerprints.

https://github.com/puppetlabs/puppetlabs-apt/commit/f588f2651a68c5863aac7dac910d96bbc7531ef9

This patch fixes the puppet warning below:

```puppet
Warning: /Apt_key[Add key: 7BD9BF62 from Apt::Source nginx]: The id should be a full fingerprint (40 characters), see README.
```

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -36,7 +36,7 @@ class nginx::package::debian(
         apt::source { 'nginx':
           location   => "http://nginx.org/packages/${distro}",
           repos      => 'nginx',
-          key        => '7BD9BF62',
+          key        => 'ABF5BD827BD9BF62',
           key_source => 'http://nginx.org/keys/nginx_signing.key',
         }
       }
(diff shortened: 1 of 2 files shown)


jfryman:
Excellent. Thanks!

errygg:
I'm still getting a warning with the 16 character gpg key:
> Warning: /Apt_key[Add key: ABF5BD827BD9BF62 from Apt::Source nginx]: The id should be a full fingerprint (40 characters), see README.
The fix is put in was to add the 40 character key instead of the 16 character key.

jfryman:
@errygg Ok. Great... misunderstood. Can you re-open and add tests then?

sbadia:
erf… sorry for that…

Issue #308, opened 2014-05-05
Problem adding if blocks inside a location using location_cfg_append/prepend


mezezo:
Trying to add an if block inside an ginx::resource::location using location_cfg_append/prepend does not work as a trailing ; is appended to the closing }.

<pre>
nginx::resource::location{ 'resources':
    location            => '/assets/',
    vhost               => $webserver_hostname,
    location_alias      => $versioned_assets,
    location_cfg_append => {
      'if'              => '($args ~ "debug=y") {
        rewrite .* /index.debug.html last;
      }',
      'try_files'       => '$uri /index.html',
      'expires'         => 'epoch',
    }
}
</pre>
Looking at the generated conf file I can clearly see that a ; is after the closing } and nginx complains about it

Error: /Stage[main]/Nginx::Service/Service[nginx]: Could not restart Service[nginx]: Execution of '/sbin/service nginx restart' returned 1: nginx: [emerg] unexpected ";" in /etc/nginx/sites-enabled...

What is the right way to add an if block inside a location element?

PR #589, opened 2015-04-01
Add missing semicolon to custom location settings

Skyr:
The custom prepend/append headers in location sections miss the trailing semicolon (as described in issue #308).


3flex:
Use `location_cfg_prepend` and `location_cfg_append` instead. It adds the trailing `;`

PR #592, opened 2015-04-02
Allow placement of include files to be a the top or the bottom

cwood:
Its handy at times to have includes run before the rest of the config. So this allows to set them to the top. It defaults to bottom.

### DIFF
--- a/manifests/resource/vhost.pp
+++ b/manifests/resource/vhost.pp
@@ -433,6 +433,9 @@ define nginx::resource::vhost (
     validate_array($include_files)
     validate_bool($include_files_bottom)
     validate_bool($include_files_top)
+    if ($include_files_bottom == $include_files_top) {
+      fail('$include_files_bottom can not be the same as $include_files_top')
+    }
   }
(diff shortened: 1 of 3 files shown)


3flex:
Hi @cwood,

Any reason you can't use the `$vhost_cfg_prepend` or `vhost_cfg_append` (or similar) parameters? We already have many ways to add directives to either the top or bottom of the vhost block and I'm  reluctant to add any more that have even more specific/niche uses.

cwood:
I mean I guess I can use that. But just seems weird since some options are paramaters and it has support for include files. So why not offer placement details?

3flex:
@cwood sorry I'm going to close this, because as I said we already have too many ways to inject custom content into the config files and some of the existing ones will work for you.

Ideally we'd like to move from using these configuration options to using puppetlabs-concat as the standard way to add custom content to the configuration files. If you're feeling ambitious it would be great to see a PR to add that functionality instead.

Thanks for the contribution though!

PR #595, opened 2015-04-08
package/redhat: correct dependency on package

3flex:
When $package_name is not 'nginx' the dependency on the package resource would be incorrect.

Probably never an issue in practice since 'nginx' is the package name in the nginx.org repo, but addresses my OCD


jfryman:
:cool:

PR #596, opened 2015-04-09
create a location

vladpanainte:
You should be able to create a location when $try_files is not empty

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -308,7 +308,7 @@ define nginx::resource::location (
   if ($vhost == undef) {
     fail('Cannot create a location reference without attaching to a virtual host')
   }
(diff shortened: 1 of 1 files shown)


wyardley:
@vladpanainte: are you willing / able to add tests for this? Obviously, it's been quite some time, but we need to either have the tests added (and the code rebased against master), or we'll need to close it. If there's no activity in about a month, we may close this.

Let me know if you need help figuring out what to test / how.

wyardley:
I looked and not sure that too much testing is necessary. I will add this in to #923, since I've re-worked the conditional there to be easier to read, so you'd have a conflict anyway..

wyardley:
D'oh, #834 had already addressed this anyway.

PR #598, opened 2015-04-10
metadata: add Puppet version compatibility

3flex:
Also:
* Remove "description" field (is deprecated)
* Declare support for Debian 7
* Remove the ".0" part of the version number for Debian & Red Hat


jfryman:
@3flex killin it!

PR #603, opened 2015-04-13
location: make auth_basic_user_file resource unique

### LINKED ISSUE #572: Can't have more than 1 password protected location
**tl;dr** I want to have more than one application served from `/` via nginx with basic auth on different ports. At the moment it doesn't seem to be possible.

**Details**

Given this setup: 

```puppet

 ::nginx::resource::vhost { 'app-1':
   listen_port => 1000,
 }
                                                     
 ::nginx::resource::location { 'app-1-proxy':
   vhost                => 'app-1',
   location             => '/',
   auth_basic_user_file => '/etc/common-htpasswd',
   auth_basic           => 'app-1',
   proxy                => 'http://app-1-upstream',
 }
                                                     
 ::nginx::resource::upstream { 'app-1-upstream':
   members => [ 'http://localhost:3000' ]
 }
                                                     
                                                     
 ::nginx::resource::vhost { 'app-2':
   listen_port => 2000,
 }
                                                     
 ::nginx::resource::location { 'app-2-proxy':
   vhost                => 'app-2',
   location             => '/',
   auth_basic_user_file => '/etc/common-htpasswd',
   auth_basic           => 'app-2',
   proxy                => 'http://app-2-upstream',
 }
                                                     
 ::nginx::resource::upstream { 'app-2-upstream':
   members => [ 'http://localhost:4000' ]
 }
                                                     

```

I'm getting an error saying that `File[/etc/nginx/__htpasswd]` is being created twice. Of course puppet doesn't like that.

The problematic line is here: https://github.com/jfryman/puppet-nginx/blob/a19cb94f2398d0b9609a04db2ec9a0071c565004/manifests/resource/location.pp#L381

Basically location of `/` gets escaped as `_`

Seems like the easiest fix would be to just use the path to original password file and use that in generated nginx config.

Is there a reason why  it's done this way? Or am I misunderstanding the usage of the resources?

3flex:
Before this change there would be conflicts if multiple vhosts contained locations with the
same name, and those locations set auth_basic_user_file.

Fixes #572

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -378,7 +378,7 @@ define nginx::resource::location (
 
   if ($auth_basic_user_file != undef) {
     #Generate htpasswd with provided file-locations
-    file { "${::nginx::config::conf_dir}/${location_sanitized}_htpasswd":
+    file { "${::nginx::config::conf_dir}/${vhost_sanitized}-${location_sanitized}_htpasswd":
       ensure => $ensure_real,
       mode   => '0644',
       source => $auth_basic_user_file,
--- a/spec/defines/resource_location_spec.rb
+++ b/spec/defines/resource_location_spec.rb
@@ -668,7 +668,7 @@ describe 'nginx::resource::location' do
       context 'when auth_basic_user_file => true' do
         let :params do { :auth_basic_user_file => '/path/to/file', :vhost => 'vhost1', :www_root => '/', } end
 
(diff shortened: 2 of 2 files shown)


3flex:
Don't merge, tests need updating.

3flex:
Closing, as this is the wrong approach.

Currently the module copies the file to `"${::nginx::config::conf_dir}/${location_sanitized}_htpasswd"`, but the template references the original file location. So the code being updated currently creates a new file then does nothing with it.

New PR will stop creating that file in the first place.

PR #605, opened 2015-04-14
location: move auth_basic directives to header

### LINKED ISSUE #260: fastcgi location does not support auth_basic
Based on https://github.com/jfryman/puppet-nginx/blob/master/templates/vhost/vhost_location_fastcgi.erb, the fastcgi location type does not support auth_basic.

nginx *does* support this configuration, however getting it in via location_cfg_prepend/append is harder than it should be.

I will try and knock together a patch in a branch for you to pull in.

### LINKED ISSUE #600: Unable to set auth_basic for "alias" location type
I've been working on an nginx host that has a default auth_basic for the vhost.  However, I am needing to add a single location with auth_basic off so that it can have a health check for a load balancer infront of it.

However, when setting auth_basic to 'off' I get the following error due to puppet treating 'off' as 'false':

```
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: false is not a string.  It looks to be a FalseClass at /etc/puppet/modules/modules/nginx/manifests/resource/location.pp:255 on node mynode.mydomain.net
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

Due to this check in location.pp:

```
  if ($auth_basic != undef) {
    validate_string($auth_basic)
  }
```

3flex:
Allows setting these directives on any location:
* auth_basic
* auth_basic_user_file

Fixes #600 
Fixes #260

### DIFF
--- a/spec/defines/resource_location_spec.rb
+++ b/spec/defines/resource_location_spec.rb
@@ -148,6 +148,18 @@ describe 'nginx::resource::location' do
           :value    => [],
           :notmatch => /rewrite/
         },
+        {
+          :title => 'should set auth_basic',
+          :attr  => 'auth_basic',
+          :value => 'value',
+          :match => '    auth_basic           "value";',
+        },
+        {
+          :title => 'should set auth_basic_user_file',
+          :attr  => 'auth_basic_user_file',
+          :value => 'value',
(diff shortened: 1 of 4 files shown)


jfryman:
:metal:

PR #606, opened 2015-04-14
vhost: simplify maintenance variable code

### LINKED ISSUE #602: redundant "maintenance" code applied to every vhost
This code is unnecessarily added to every vhost:
```
  if ($maintenance = "on") {
      <%= @maintenance_value %>;
  }
```

But it only does anything when `maintenance => on` is on. Introduced by #578 & #584.

3flex:
Fixes #602


jfryman:
:+1:

PR #613, opened 2015-04-23
travis: test on Puppet 4 for real

3flex:
No need to test with future parser, the real thing is here.


3flex:
I expect test failures as Travis will use puppetlabs-apt 2.0.0 module because the version's not pinned in .fixtures.yml. Merge #612 to address that.

PR #616, opened 2015-04-28
fix docs

Cofyc:
Currently, nginx::geos should be nginx::geo_mappings.


jfryman:
Good find.

Very weird build errors, though. Wonder if it is transient or not. :/

3flex:
You can review then merge https://github.com/jfryman/puppet-nginx/pull/612 to fix the build failures. There are backwards incompatible changes introduced by puppetlabs-apt 2.0.0

PR #618, opened 2015-04-29
Add passenger_set_header and passenger_env_var parameters for Passenger 5.0+

mmarod:
Passenger 5.0 deprecates passenger_set_cgi_param and replaced it with passenger_set_header and passenger_env_var. This allows you to specify those parameters.

https://blog.phusion.nl/2015/03/04/phusion-passenger-5-0-1-released/

### DIFF
--- a/templates/vhost/vhost_header.erb
+++ b/templates/vhost/vhost_header.erb
@@ -94,6 +94,11 @@ server {
   passenger_set_cgi_param  <%= key %> <%= @passenger_cgi_param[key] %>;
   <%- end -%>
 <% end -%>
+<% if @passenger_header -%>
+  <%- @passenger_header.keys.sort.each do |key| -%>
+  passenger_set_header  <%= key %> <%= @passenger_header[key] %>;
+  <%- end -%>
+<% end -%>
 <% if Array(@resolver).count > 0 -%>
   resolver                  <% Array(@resolver).each do |r| %> <%= r %><% end %>;
 <% end -%>
--- a/templates/vhost/vhost_header.erb
+++ b/templates/vhost/vhost_header.erb
@@ -99,6 +99,11 @@ server {
   passenger_set_header  <%= key %> <%= @passenger_header[key] %>;
   <%- end -%>
 <% end -%>
+<% if @passenger_env_var -%>
+  <%- @passenger_env_var.keys.sort.each do |key| -%>
+  passenger_set_env_var  <%= key %> <%= @passenger_env_var[key] %>;
+  <%- end -%>
+<% end -%>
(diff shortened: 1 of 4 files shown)


templates/vhost/vhost_header.erb
```diff
@@ -94,6 +94,16 @@ server {
...
   <%- end -%>
 <% end -%>
+<% if @passenger_header -%>
+  <%- @passenger_header.keys.sort.each do |key| -%>
+  passenger_set_header  <%= key %> <%= @passenger_header[key] %>;
+  <%- end -%>
+<% end -%>
+<% if @passenger_env_var -%>
+  <%- @passenger_env_var.keys.sort.each do |key| -%>
+  passenger_set_env_var  <%= key %> <%= @passenger_env_var[key] %>;
```
3flex: The directive name according to passenger docs is `passenger_env_var`.

templates/vhost/vhost_header.erb
```diff
@@ -94,6 +94,16 @@ server {
   passenger_set_cgi_param  <%= key %> <%= @passenger_cgi_param[key] %>;
   <%- end -%>
 <% end -%>
+<% if @passenger_header -%>
```
3flex: Please call this parameter `passenger_set_header` so it matches the nginx directive name.

templates/vhost/vhost_ssl_header.erb
```diff
@@ -152,6 +152,12 @@ server {
 <% Array(@passenger_cgi_param).each do |key,value| -%>
   passenger_set_cgi_param  <%= key %> <%= value %>;
 <% end -%>
+<% Array(@passenger_header).each do |key,value| -%>
```
3flex: Please call this parameter `passenger_set_header` so it matches the nginx directive name.

templates/vhost/vhost_ssl_header.erb
```diff
@@ -152,6 +152,12 @@ server {
...
+  passenger_set_header  <%= key %> <%= value %>;
+<% end -%>
+<% Array(@passenger_env_var).each do |key,value| -%>
+  passenger_set_env_var  <%= key %> <%= value %>;
```
3flex: The directive name according to passenger docs is `passenger_env_var`.

templates/vhost/vhost_header.erb
mmarod: I was following the pattern for 'passenger_set_cgi_param' which only uses '_set_' in the template when it is actually writing the config.

3flex:
Yeah, several existing parameters don't match the name of the nginx directives, but it's certainly something we'd like to avoid in future. Thanks for updating!

PR #622, opened 2015-05-04
fix rewrite_www_to_non_www vhost

andrenarchy:
The www-to-non-www vhost actually does not need any further listen_options (e.g., see [here](http://serverfault.com/a/548942)).

### DIFF
--- a/templates/vhost/vhost_ssl_header.erb
+++ b/templates/vhost/vhost_ssl_header.erb
@@ -2,19 +2,19 @@
 server {
   <%- if @listen_ip.is_a?(Array) then -%>
     <%- @listen_ip.each do |ip| -%>
-  listen       <%= ip %>:<%= @ssl_port %> <% if @ssl_listen_option %>ssl<% end %><% if @spdy == 'on' %> spdy<% end %><% if @listen_options %> <%= @listen_options %><% end %>;
+  listen       <%= ip %>:<%= @ssl_port %>;
     <%- end -%>
   <%- else -%>
-  listen       <%= @listen_ip %>:<%= @ssl_port %> <% if @ssl_listen_option %>ssl<% end %><% if @spdy == 'on' %> spdy<% end %><% if @listen_options %> <%= @listen_options %><% end %>;
+  listen       <%= @listen_ip %>:<%= @ssl_port %>;
   <%- end -%>
 <%# check to see if ipv6 support exists in the kernel before applying -%>
   <%- if @ipv6_enable && (defined? @ipaddress6) -%>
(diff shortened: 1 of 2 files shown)


andrenarchy:
Oh, and I should mention that the generated config is broken without the fix. Error message was:

```
[...] duplicate listen options for [::]:80 in /etc/nginx/sites-enabled/[...]
```

3flex:
I'm happy to merge this, but could you rebase and remove that second commit? I'm not sure I see any benefit to moving the rewrite server below the regular server block. Thanks!

andrenarchy:
Without moving the rewrite server block below the regular block, nginx was able to load the config but I got this error when accessing the www-version via https *and* ipv4:
```
*138 no "ssl_certificate" is defined in server listening on SSL port while SSL handshaking, client: [...], server: 0.0.0.0:443
```
However, you're absolutely right that the order shouldn't matter. The clean solution to this problem is adding `listen_options => 'default'` to the `nginx::resource::vhost`. The ipv6 option includes this by default (`ipv6_listen_options` is set to `default ipv6only=on`) but the ipv4 option `listen_options` is `undef` by default. Maybe the default for `listen_options` should also be changed to `default` in order to be consistent and allow an easy www-to-non-www config?

andrenarchy:
Any updates?

3flex:
There are some listen options that are safe to duplicate between resources, so I'd like to see the `ssl` and `spdy` flags remain in the SSL header template since it will be a little easier to tell what's going on in the generated template.

Otherwise it looks good! We'll keep the `listen_options` as `undef` for now, so no need to change that.

andrenarchy:
What's the progress here? 

@3flex I'm not sure what exactly you like to see fixed...

templates/vhost/vhost_ssl_header.erb
```diff
@@ -2,19 +2,19 @@
 server {
   <%- if @listen_ip.is_a?(Array) then -%>
     <%- @listen_ip.each do |ip| -%>
-  listen       <%= ip %>:<%= @ssl_port %> <% if @ssl_listen_option %>ssl<% end %><% if @spdy == 'on' %> spdy<% end %><% if @listen_options %> <%= @listen_options %><% end %>;
```
3flex: Don't remove `ssl` or `spdy` flags here. They can be duplicated across listen directives.

This applies to the next 3 changed lines as well.

andrenarchy:
I switched all my hosts to ansible and won't have time to pursue this PR. Feel free to adapt the code to your needs.

PR #636, opened 2015-05-29
Vhost and loation proxy_cache_key and proxy_cache_use_stale

jacobmw:


manifests/config.pp
```diff
@@ -66,6 +66,7 @@
   $proxy_buffers                  = '32 4k',
   $proxy_buffer_size              = '8k',
   $proxy_cache_inactive           = '20m',
+  $proxy_cache_key                = '$scheme$proxy_host$request_uri',
```
3flex: I'm not seeing where this is referenced from... can it be removed?

3flex:
Apart from my comment, looks good. Can you address that feedback, then squash the submission down to one commit? Then we can get it merged.

If you're not sure how to do this, there's a useful guide here: https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

manifests/config.pp
jacobmw: $scheme$proxy_host$request_uri is the actual default value:
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_key

3flex: Yep, but the nginx::config::proxy_cache_key parameter you created here isn't referenced anywhere else.

jacobmw:
I see what you mean about the proxy_cache_key.  I've removed it and attempted to squash my submission.  Please let me know if there's anything else I need to do.

PR #640, opened 2015-06-07
Update vhost_header.erb

SoldierCorp:
www_root should to be the correct variable not root. Fixing this line my Laravel project is working correctly.

### DIFF
--- a/templates/vhost/vhost_header.erb
+++ b/templates/vhost/vhost_header.erb
@@ -86,8 +86,8 @@ server {
 <% Array(@raw_prepend).each do |line| -%>
   <%= line %>
 <% end %>
-<% if @root -%>
-  root <%= @root %>;
+<% if @www_root -%>
+  root <%= @www_root %>;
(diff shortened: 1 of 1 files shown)


wyardley:
IIRC, this is intentional, no?

`$root` is set from `$www_root` here in the vhost manifest (Puppet manifests won't let you reassign variables):

See between:
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/vhost.pp#L604-L651

If `$use_default_location` is true, `$www_root` gets set in the default location block (though I doubt it would hurt to set it in the server {} block as well?); if it's false, `$root` gets set to `$www_root` in the server {} block.

wyardley:
ps - I don't know that it's harmful to have root set in both the default location and the `server {} ` block, but I don't think it should be necessary? Can you explain your use case where it was failing?

wyardley:
I think this works as intended, if this is still a problem, please re-open with an example use case.

PR #647, opened 2015-06-17
Added parameters daemon_group and set_daemon_group

jakm:
Allow specify daemon's group as documented at http://nginx.org/en/docs/ngx_core_module.html#user

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -19,6 +19,8 @@ class nginx::config(
   $confd_purge                    = false,
   $conf_dir                       = $::nginx::params::conf_dir,
   $daemon_user                    = $::nginx::params::daemon_user,
+  $daemon_group                   = $::nginx::params::daemon_group,
+  $set_daemon_group               = $::nginx::params::set_daemon_group,
   $global_owner                   = $::nginx::params::global_owner,
   $global_group                   = $::nginx::params::global_group,
   $global_mode                    = $::nginx::params::global_mode,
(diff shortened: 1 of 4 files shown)


3flex:
@jakm thanks for the PR!

I'd ask that you don't add the `$set_daemon_group` parameter and always populate the daemon_group in the template, there's no need to make it conditional as the defaults you've set will be safe.

manifests/config.pp
```diff
@@ -216,11 +218,13 @@
   file {$client_body_temp_path:
     ensure => directory,
     owner  => $daemon_user,
+    group  => $daemon_group,
```
3flex: You can remove this. The group is 'root' but with permissions 0644 it will be readable even if you change the daemon user.

manifests/config.pp
```diff
@@ -216,11 +218,13 @@
...
   }
 
   file {$proxy_temp_path:
     ensure => directory,
     owner  => $daemon_user,
+    group  => $daemon_group,
```
3flex: You can remove this. The group is 'root' but with permissions 0644 it will be readable even if you change the daemon user.

jakm:
Parameter `$set_daemon_group` is added because of backward compatibility. If $daemon_group is set to, for example, web-nginx-14 then generated nginx.conf will contain `user web-nginx-14;` and nginx's processes will have set both user and group to web-nginx-14. Patch without $set_daemon_group parameter would break configurations with non-standard user as described above. Because $daemon_group's default is set to www-data (www, ...).

wyardley:
@jakm: are you able / willing to add tests to this request? Do you need any help determining what / how to test? The commits would also need to be squashed.

At the least, there should be more or less a copy / paste of https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/config_spec.rb#L95 

As far as the true / false `$set_daemon_group` param, why not just make it default to undef, and leave out all the params for each OS, that way, people can set it explicitly if they need to override the default. Do you know if nginx always supported the group as well as the user, or is that a change that was added at some point?

wyardley:
No response, going to close this one for now, though I would like to see a reworked version of this at some point as well.

PR #649, opened 2015-06-19
printing www_root in vhost_header

sirgt:
update vhost_header.erb, www_root variable is printed in server section. I'm aware that this might not be the best way to do this but i modified the module on my puppet server and now nginx vhost is working good. Also I'm not fully aware about how to create tests for a puppet module so this is lacks of tests, maybe someone else can help us with that.


wyardley:
See my comment in https://github.com/voxpupuli/puppet-nginx/pull/640#issuecomment-252109325

wyardley:
As mentioned in #640, I'm not sure that this is a bug, but either way, the two are more or less the same, so I'm going to close this one, and we can have any further discussion there.

PR #652, opened 2015-06-24
update default SSL ciphers

pulecp:
suggested from https://wiki.mozilla.org/Security/Server_Side_TLS#Nginx_configuration_details and https://weakdh.org/sysadmin.html


3flex:
Looks good, thanks!

PR #657, opened 2015-07-03
fix circuler dependency

b4ldr:
See issue https://github.com/jfryman/puppet-nginx/issues/656


3flex:
Did you test this on FreeBSD?

I'm not sure how this would solve the issue. `nginx::config` is a dependency of `nginx::service` so the service shouldn't start before the config is in place, including the creation of the log directory. While `nginx::package` also notifies `nginx::service`, the service start/restart won't happen until all dependencies have been resolved, including `nginx::config`.

It may be that the error log is being created in the wrong location in FreeBSD.

b4ldr:
Yes i did test this in freebsd as soon as i applied the patch the policy ran without error.   You say that nginx::config is a dependency of nginx::service however it is not.  nginx::config is set to notify nginx::service it does not require it and therefore there is not guaranteed to run before nginx::config.  however nginx::config does require nginx::package which means everything in nginx::package including the notify to nginx::service needs to apply before nginx::config.  

At least that is my understanding of how the dependency resolution works.  Regardless without this patch nginx fails to run on a fresh freeBSD 10.1, with the patch t runs without issue

3flex:
In Puppet when one resource notifies another, it still applies the standard ordering rules. Since `nginx::config` notifies `nginx::service`, all the resources in `nginx::config` will be applied before `nginx::service`.

While the patch might fix things in your environment I'm concerned about changing this behaviour as it may have unintended consequences on other platforms. We had a problem on DragonflyBSD/FreeBSD (https://github.com/jfryman/puppet-nginx/issues/619) and I thought https://github.com/jfryman/puppet-nginx/pull/635 might have solved it, but that might not be the case.

I'd like to know what the root cause is before applying any change and I'm not sure we know what that is yet. It may be related to #619 and possibly the change in #635 did not address it properly.

b4ldr:
ok no problem, ill try and set up a vagrant image to better troubleshoot the issue however im away at conferences for the next two weeks so it might take a while

3flex:
Closing for now. If we get more information we can reopen. Thanks!

PR #658, opened 2015-07-06
sort add_header values for ssl vhost

cgroschupp:


3flex:
Sweet!

If you could please squash to a single commit we'll get this merged right in. Thanks! (and thanks especially for adding tests)

cgroschupp:
Done

PR #666, opened 2015-08-07
Don't qualified call to defined resource type

PierreR:
I don't think this is such an improvement (over non qualification).

My main motivation comes from the fact that it is not currently supported so well by "language-puppet".


3flex:
Thanks, and my mistake for doing this in the first place as resource names are not limited by scope so as you say, don't need to be fully qualified in this manner.

PR #667, opened 2015-08-07
adding semi-colon to end of location_custom_cfg_append options.

kwolf:
This already exists on the prepend, just missing on the append.


3flex:
@kwolf thanks for the contribution, but you should be aware that this is by design.

If you want the versions with a trailing semi-colon you should use `location_cfg_append` instead.

kwolf:
Sorry about that.  Thanks for pointing out the other option!

Issue #536, opened 2014-12-23
Hiera lookup


PierreR:
I have spent a couple of hours to migrate our config to use hiera as indicated by https://github.com/jfryman/puppet-nginx/blob/master/docs/hiera.md

While I am testing this, I see no param change when I use `nginx::config::parameterx`.

Here is an example:

```
nginx::config::client_max_body_size: '100m'
nginx::config::worker_processes: 2
```
using:
```
class { 'nginx':
}
```
Of course It works when I use:
```
class { 'nginx':
  worker_processes => 2
}
```

Note also that if I use in hiera:
```
nginx::worker_processes: 2
```
then it works ...

Am I missing something ?

I am using the latest tag `0.2.1` and puppet 3.7.3

Thanks for your help.

PierreR:
@jfryman  I have tried your suggestion:
```
  class { 'nginx': }
  include nginx::config

  nginx::resource::upstream { $upstream:
    members => ['localhost:7990', ],
  }

  nginx::resource::vhost { $vhosts:
    proxy   => "http://${upstream}",
  }
```
But I get this complain from [language-puppet](https://github.com/bartavelle/language-puppet) :
```
Can't include class nginx::config twice when using the resource-like syntax (first occurence at # "./modules/nginx/manifests/init.pp" (line 217, column 13)) at # "./modules/application/manifests/profile/stash.pp" (line 10, column 3)
```
What wrong with using `nginx::worker_processes: 2` instead of `nginx::config::worker_processes: 2`

Thanks for your help.

Hufschmidt:
@PierreR Was having the same issue. The problem is that you need to make sure that "your" settings for the nginx::config class are proccessed before nginx is evaluated/compiled, because it includes its own nginx::config class (from parameters given to the nginx-class) IFF its non is defined.

The problem is you cannot do Class['nginx::config'] -> Class['nginx'] because this would lead to a cyclic dependency.
Because inside init.pp:
    Class['::nginx::package'] -> Class['::nginx::config'] ~> Class['::nginx::service']
    [...]
    anchor{ 'nginx::begin':
        before => Class['::nginx::package'],
        notify => Class['::nginx::service'],
      }

Luckily this also "contains" the solution for the ordering problem:
      Anchor['nginx::begin']
      ->
      class { 'nginx::config' :
        confd_purge   => true,
        vhost_purge   => true,
      }
      class { 'nginx' :
        manage_repo   => false,
      }
[...]

PR #669, opened 2015-08-11
Update non-hiera usage (see #536)

Hufschmidt:
As described in issue #536, the example on how to use nginx correctly without hiera needs to be updated to fix parse-order dependency with defined(nginx::config) with puppet 3.x.


jfryman:
Thanks for the code!

PR #673, opened 2015-08-28
Add 'ensure' parameter to resource::upstream::member.

kwolf:
I'm using stored configs to create upstream::member on servers that should belong to up upstream.  Removing them while developing the puppet manifests if something goes wrong is troublesome without the ability to ensure => absent.

All tests are passing under ruby 2.1.2, though I didn't add new ones.


jfryman:
Looks great. Thanks for the code!

Issue #381, opened 2014-07-19
vhost: $rewrite_www_to_non_www


guyhughes:
Using latest github release https://github.com/jfryman/puppet-nginx/commit/61b569ca02e11a86c1abf647490b5f98a4aee03c on ruby 2.1.2p95 when `$rewrite_www_to_non_www => true,` in a vhost, having an array in `$server_name` goes wonky and only the first element of the array is handled.

Haven't investigated further. Only discovered that unsetting `$rewrite_www_to_non_www` removed the issue.

May be related to:
https://github.com/jfryman/puppet-nginx/pull/114
https://github.com/jfryman/puppet-nginx/pull/310
https://github.com/jfryman/puppet-nginx/pull/113

3flex:
@guyhughes I'm closing this as you said it seems to be resolved. If not please comment and I'll reopen. Cheers!

PR #683, opened 2015-09-15
iterate server_name when rewrite_www_to_non_www is used

kronos-pbrideau:
Was not working when server_name was an array

with
```
server_name            => [ 'a.example.com', 'b.example.org' ],
rewrite_www_to_non_www => true,
```

Before patch:
```
server {
  [...]
  server_name www.a.example.com;
  return 301 http://a.example.com$request_uri; 
}
[...]
server {
  [...]
  servername a.example.com
  [...]
}
```

now:
```
server {
  [...]
  server_name www.a.example.com;
  return 301 http://a.example.com$request_uri; 
}
server {
  [...]
  server_name www.b.example.org;
  return 301 http://b.example.org$request_uri; 
}

[...]
server {
  [...]
  servername a.example.com b.example.org
  [...]
}
```
```

### DIFF
--- a/spec/defines/resource_vhost_spec.rb
+++ b/spec/defines/resource_vhost_spec.rb
@@ -711,8 +711,8 @@ describe 'nginx::resource::vhost' do
           }
         end
 
-        it "should set the server_name of the rewrite server stanza to the first server_name with 'www.' stripped" do
-          is_expected.to contain_concat__fragment("#{title}-ssl-header").with_content(/^\s+server_name\s+foo.com;/)
+        it "should set the server_name of the rewrite server stanza to every server_name with 'www.' stripped" do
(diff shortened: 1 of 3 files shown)


kronos-pbrideau:
Was it intended to show _only_ the first server of the list?  The comment in the tests suggest it is intended, but I don't understand the reason...

in the spec file:
`should set the server_name of the rewrite server stanza to the first server_name with 'www.' stripped"`

I don't seem to be the only one to detect this bug though: #381

@kalmanolah beeing the one who submit PR #310 do you see something bad about my PR?

kalmanolah:
@kronos-pbrideau The PR looks alright to me. How I missed/ignored the fact that only the first server name was being handled is still a mystery to me. Thanks for the fix :+1:

3flex:
@kalmanolah thanks for reviewing!

PR #687, opened 2015-09-18
Ensure priority is an integer 

erikanderson:
Ensure that priority is an integer before comparisonWith stringify_facts set to false puppet fails with this error: 

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Comparison of: String < Integer, is not possible. Caused by 'A String is not comparable to a non String'. at /srv/puppetmaster/current/modules/nginx/manifests/resource/location.pp:238:17 on node sa-sand-logyard1.sa.moneydesktop.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

This ensures that $priority is an integer

https://docs.puppetlabs.com/puppet/latest/reference/lang_data_number.html#converting-strings-to-numbers

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -308,7 +308,7 @@ define nginx::resource::location (
     fail('$priority must be an integer.')
   }
   validate_array($rewrite_rules)
-  if ($priority < 401) or ($priority > 899) {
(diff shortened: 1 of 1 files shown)


erikanderson:
travis is erroring for no reason, will recreate

Issue #684, opened 2015-09-17
Error: Comparison of: String < Integer, is not possible


erikanderson:
We are getting ready to upgrade from puppet 3.8 to 4.2 so we have set `stringify_facts = false`. Once we did that an error popped up from this module. The full error: 

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Comparison of: String < Integer, is not possible. Caused by 'A String is not comparable to a non String'. at /modules/nginx/manifests/resource/location.pp:238:17
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run

and the section in question: 

```
  $priority             = 500
...
  validate_array($rewrite_rules)
  if ($priority < 401) or ($priority > 899) {
    fail('$priority must be in the range 401-899.')
  }
```

 To force puppet to see the $priority as an integer by adding zero to it: 
(https://docs.puppetlabs.com/puppet/latest/reference/lang_data_number.html#converting-strings-to-numbers)
``` 
  validate_array($rewrite_rules)
  if (($priority + 0) < 401) or (($priority + 0) > 899) {
    fail('$priority must be in the range 401-899.')
  }
```

erikanderson:
Created PR: https://github.com/jfryman/puppet-nginx/pull/687

erikanderson:
Updated PR to use str2num

erikanderson:
Created PR https://github.com/jfryman/puppet-nginx/pull/689

PR #689, opened 2015-09-22
Convert $priority to integer before comparison

erikanderson:
Comparison was causing an error with parser = future and stringify_facts = false. This ensures that the comparison is done between two integers

Fixes: #684


jfryman:
Very nice. Thank you for the code!

erikanderson:
Thanks for merging this in but we were looking in our code and found that we were mistakenly setting this value to an string. The code I added will still work but I think it is unnecessary.

PR #693, opened 2015-10-05
Adding more configuration variables with default values

victorgp:
Some of this variables already existed in "server" section, but according to Nginx docs they are also allowed in "http" section, so adding them with the default values.

Also, supporting error_page that was missing

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -44,6 +44,9 @@ class nginx::config(
   ### START Nginx Configuration ###
   $client_body_buffer_size        = '128k',
   $client_max_body_size           = '10m',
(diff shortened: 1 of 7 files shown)


wyardley:
Following. I need a couple of these (though I've just passed them in freeform instead).
Since this is pretty old, I can try to add some tests if I've got time.

wyardley:
@3flex @ffrank: #881 should address the feedback here (added tests and rebased against recent-ish master).

bastelfreak:
This got merged as https://github.com/voxpupuli/puppet-nginx/pull/881

PR #698, opened 2015-10-09
Allow map item input order to be preserved

btisdall:
The template used by `nginx::resource::map` sorts map items lexically on the key, presumably to provide sane behaviour on Ruby 1.8.7; this does not work for situations where multiple map keys match a given input and first-match-wins behaviour is desired. This change allows mappings to be supplied as an array of hashes where preservation of item order is important.

### DIFF
--- a/manifests/resource/map.pp
+++ b/manifests/resource/map.pp
@@ -28,6 +28,16 @@
 #    }
 #  }
 #
+# Sample Usage (preserving input of order of mappings):
+#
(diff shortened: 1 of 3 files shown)


wyardley:
Hi @btisdall: are you still interested in having this merged, and if so, can you rebase it against master (and squash any additional commits)?

wyardley:
replacing with #934

PR #700, opened 2015-10-19
Update vhost proxy_set_header defaults to match location

alext:
`nginx::resource::location` defaults this value to
`$::nginx::config::proxy_set_headers`, whereas `nginx::resource::vhost`
defaults it to `[]`. This means that you get a different set of defaults
if you use an explicit `nginx::resource::location` resource as opposed to
setting it up via an `nginx::resource::vhost` with a proxy attribute.

Updating this default value to be consistent will be less surprising.


jfryman:
:+1:

PR #701, opened 2015-10-19
Make the name of the upstream that appears in the conf.d file configurable

jharlow1:
Make the name that appears in the nginx configuration independent of the
namevar known to puppet. This is useful when using exported resources,
if you want to have the same upstream names in the config file on multiple independent
clusters exporting to the same puppetdb.

Also, modify spec for upstreams to test for proper string after upstream_name param addition


jharlow1:
Retracting this for now...the file names in the conf.d directory still end up not matching the upstream names, which is technically solves the problem I'm facing but ends up looking a bit confusing...the hack to get around that adds a fair amount of ugliness so not sure it is a good idea....going to see if I can fix my own downstream dependency that raised this issue (some LUA scripts that parse the nginx config and need the upstream name) rather than try and add spaghetti here.

PR #707, opened 2015-10-30
Add unix socket for listening.

werekraken:
nginx can listen on unix sockets in addition to v4 and v6 addrs. This adds parameters to permit this use case.
http://nginx.org/en/docs/http/ngx_http_core_module.html#listen

The parameters emulate the ipv6 implementation. The enable parameter only affects listen in the unix socket case (not things like proxy_pass), so I called this listen_unix_socket_enable rather than unix_socket_enable.


jfryman:
Very nice. Thanks for the code!

PR #709, opened 2015-11-01
Add parameter to allow setting error_log severity level

Phil-Friderici:
This patch allows to set the severity level for error_log.
If not set it will use the default severity level of 'error'.
$nginx_error_log_severity is validated to only allow valid values.


Phil-Friderici:
If you don't like that this patch will add the severity level to nginx.conf, I can modify it to only add it if severity level is set by the user intentionally. Personally I think setting default values in configurations files enhance readability.

ghoneycutt:
Hi @jfryman 

We have a dependency on this as we transition to using your awesome module. Could you please take a look at this PR?

jfryman:
LGTM. Thanks for the code!

ghoneycutt:
Thank you James and Phil!

PR #711, opened 2015-11-02
added package_name to config class

anthonyclarka2:
Allows nginx-plus to be specified in Hiera (along with manage_repo = false)

I haven't added an "if package_name is nginx-plus then fail if manage_repo is true" statement but I can if required for this to be accepted.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -35,6 +35,7 @@ class nginx::config(
   $super_user                     = $::nginx::params::super_user,
   $temp_dir                       = $::nginx::params::temp_dir,
   $vhost_purge                    = false,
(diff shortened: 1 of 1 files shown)


3flex:
Did you try setting in hiera with `nginx::package::package_name`? That should allow you to overwrite the package name.

anthonyclarka2:
No, I hadn't tried nginx::package at all, just nginx::config.  I'll test that out on Monday.  Thanks!

3flex:
@anthonyclarka2 I didn't hear back, so will close this for now. If you're still having trouble please post and I'll reopen. Thanks!

PR #712, opened 2015-11-02
Ensure configtest still runs before reload

jfryman:
Accompanies https://github.com/jfryman/puppet-nginx/pull/702

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -110,7 +110,7 @@ class nginx (
   $configtest_enable              = false,
   $service_ensure                 = running,
   $service_flags                  = undef,
-  $service_restart                = '/etc/init.d/nginx reload',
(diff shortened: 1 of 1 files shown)


3flex:
Did I miss something? The docs say reload "checks the syntax validity of the new configuration file"

So why run configtest manually?

jfryman:
@3flex that is what I thought as well, but some older versions with unpatched systems do not handle that properly, so I went with safe over sorry. Maybe it truly isn't needed.

pallavjosh:
Again in file puppet-nginx/spec/classes/service_spec.rb, restart command is being used. Shouldn't there too we need to make this change.

As in case upstream is wrong still shoots error as execution of /etc/init.d/nginx restart command failed. That pointed out issue.

Please have a look.
Thanks,
Pallav Joshi.

PR #716, opened 2015-11-11
Provide more info on deprecations for user

erikanderson:
This will provide more info for the user on what they need to change


erikanderson:
I realized this is not going to work with puppet 3 and travis seems to not like the heredoc

PR #719, opened 2015-11-18
allow reverse proxy in config

maces:
Change Rule to allow Ngnix to be used as reverse Proxy


maces:
Failed because of: ```should raise Puppet::Error with message matching /Cannot define both directory and proxy in a virtual host/ (FAILED - 1)```
But that's exactly what this PR allows.

PR #721, opened 2015-11-21
File[undef] error: re-allowed vhost's www_root with location's proxy

dguhl:
Hello there.

This PR fixes a circumstance in Nginx' module configuration. The removed code did not allow a www_root for a vhost with a proxy in one of its locations, although it is allowed in nginx configurations. Configuring the vhost without a www_root to satisfy the location spec leads to an error in Puppet-Nginx main class, where an undefined File reference is not allowed. (The message is `Error: Parameter path failed on File[undef]: File paths must be fully qualified, not 'undef' at /tmp/vagrant-puppet/manifests-[hash]/nodes/Nginx.pp:187`


bastelfreak:
I'm going to close this PR since @herzmut deleted his branch. Please let me know if you are still interested in this PR.

PR #724, opened 2015-11-30
improved location ordering

vicinus:
This change orders locations the same way if a virtual host supports both http and https. Also if two virtual hosts have the same locations the ordering is identical.


jfryman:
Ah. Nice. Thanks for the contribution!

PR #725, opened 2015-12-03
restart replaced with reload in service_spec.rb file

pallavjosh:
@jfryman related with pull request 702 comment from me. review ad confirm please.
https://github.com/jfryman/puppet-nginx/pull/702

### DIFF
--- a/spec/classes/service_spec.rb
+++ b/spec/classes/service_spec.rb
@@ -3,7 +3,7 @@ describe 'nginx::service' do
 
   let :params do {
       :configtest_enable => false,
-      :service_restart => '/etc/init.d/nginx relaod',
+      :service_restart => '/etc/init.d/nginx reload',
       :service_ensure => 'running',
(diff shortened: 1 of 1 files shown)


3flex:
@pallavjosh this PR is only updating the tests, you would also need to update the Puppet manifests themselves.

Also please be careful with your spelling, the command would be `reload` not `relaod`

pallavjosh:
@3flex Sorry for the typo, corrected. Though in this case manifest is already changed with pull request #702 #712 , test was unchanged. So changing test file only with this PR
 
If In case upstream was wrong still shooting error as execution of "/etc/init.d/nginx restart" command failed. That pointed out issue.

pallavjosh:
@jfryman Please review, merge and confirm on reload part. Thanks

jfryman:
@pallavjosh Thanks so much for the contribution! 🎩🙇🏽

PR #732, opened 2015-12-28
new variable: http_raw_append for nginx.conf

grzybniak:
I added a new variable for special configuration (raw). An example for ldap config:

#### Hieradata ###
nginx::http_raw_append:
  - 'ldap_server my_ldap {'
  - 'url ldap://ldap.server/dc=my_dc?uid?sub?(|(description=*admin*)(description=*options*));'
  - 'binddn "cn=cn_auth,dc=my_dc";'
  - 'binddn_passwd ***password***;'
  - 'group_attribute_is_dn on;'
  - 'require valid_user;'
  - 'satisfy any;'
  - '}'

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -63,6 +63,7 @@ class nginx::config(
   $gzip_types                     = 'text/html',
   $gzip_vary                      = 'off',
   $http_cfg_append                = false,
+  $http_raw_append                = false,
   $http_tcp_nodelay               = 'on',
   $http_tcp_nopush                = 'off',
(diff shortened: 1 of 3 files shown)


3flex:
My personal belief is that instead of adding more `raw` parameters, that we should take further advantage of puppet-concat. Additional support for that would be preferable over adding more of these parameters (since we'll get another for prepend_cfg, then prepend_raw, and it grows from there).

The concat module is perfect for these cases and if support for that was added it would be very beneficial. I'm going to close this because I'd much rather concat support be expanded instead of using this approach.

PR #738, opened 2016-01-17
Added locations paramater to use it in hiera.

jkroepke:
Hi,

I add the paramter locations. With this, you can write more cleaner hiera.

Also, puppetlabs/apache has this option, too.

An exmaple based from the example on the README:
```
nginx::nginx_vhosts:
  'www.puppetlabs.com':
    www_root: '/var/www/www.puppetlabs.com'
    locations:
      'static':
        location: '~ "^/static/[0-9a-fA-F]{8}\/(.*)$"'
      'userContent':
        location: /userContent
  'rack.puppetlabs.com':
    proxy: 'http://puppet_rack_app'
```


jfryman:
:+1: Thanks for the code.

PR #744, opened 2016-01-20
Remove extra white space

gerases:
This is what I was able to correct this far based on the vhosts I've been creating.


jfryman:
Nice! Thanks!

PR #745, opened 2016-01-20
Make it possible to not include ssl_ciphers in the resulting vhost.

gerases:


templates/vhost/vhost_ssl_settings.erb
```diff
@@ -15,7 +15,9 @@
   ssl_buffer_size           <%= @ssl_buffer_size %>;
 <% end -%>
   ssl_protocols             <%= @ssl_protocols %>;
+<% if @ssl_ciphers != '' -%>
```
jfryman: This wouldn't be `if @ssl_ciphers`?

gerases: I might be mistaken, but since `@ssl_ciphers` defaults to the long list of ciphers, setting it in the configuration to an empty string would work. Like so:

`ssl_ciphers   => ''`

What do you think?

gerases: But ultimately, whatever you think would work best

gerases: I think it's safe to ignore my pull request because I think a more comprehensive solution is needed. I will try to come up with an idea and let you know.

3flex:
Closing in favor of #794

PR #746, opened 2016-01-21
Add expires option to the location class

Snowmean:
I've added "expires" option to the locations class. 

We are using that option in our servers configurations, and it helps us to reduce amount of traffic significantly due to different cache expiration time for static content like jpg, css, js etc... So I think it would be nice to have it in public module as well.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -90,6 +90,7 @@
 #     used for mp4 streaming. Default: false
 #   [*flv*]             - Indicates whether or not this loation can be
 #     used for flv streaming. Default: false
+#   [*expires*]         - Setup expires time for locations content
 #
 #
 # Actions:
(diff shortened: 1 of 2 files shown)


3flex:
@Snowmean sorry for the huge delay. Can you please rebase on master, and squash this down to a single commit for merging?

Adding new tests would also be very useful and speed up the merging process once updated. Thanks!

wyardley:
Hi --
This still hasn't had a rebase, and has no tests. If there's no activity in a month, we'll close this PR.

Snowmean:
You can close it now, I have no time for that now, and I doubt I will have it anytime soon.

wyardley:
Ok, thanks. It's pretty simple and seems like it could be useful (even though there are workarounds), so I will take a stab at updating it if I get some time.

wyardley:
Replacing this for you in #924 (with test / rebase).

PR #752, opened 2016-01-28
Issue 751 Add an example for a simple reverse proxy to the README

### LINKED ISSUE #751: README not helping as much as it could to create a reverse proxy
I had a simple requirement to proxy port 80 through to a Kibana 4 application running on localhost:5601.

The README seems to imply that I need to do something like this:

```
include nginx
nginx::resource::upstream { $::fqdn:
  members => [ 'localhost:5601' ],
}
nginx::resource::vhost { $::fqdn:
  proxy => "http://${::fqdn}",
}
```

(By the way, this led to a very unhappy Nginx that had lots of `(24: Too many open files)` messages when I tried to connect to it.)

I found the answer in the tests directory:

```
include nginx
nginx::resource::vhost { "${::fqdn}:80":
  ensure      => present,
  listen_port => 80,
  server_name => ["${::fqdn} ${::hostname}"],
  proxy       => 'http://localhost:5601',
}
```

which can be simplified to:

```
include nginx
nginx::resource::vhost { "${::fqdn}:80":
  proxy       => 'http://localhost:5601',
}
```

I would recommend making the documentation clearer for this simple use-case.

alexharv074:
Add an example for a simple reverse proxy to the README.

Fixes #751

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -33,7 +33,16 @@ This module manages NGINX configuration.
 class { 'nginx': }
 ```
 
-### Creating a new virtual host
+### A simple reverse proxy
+
+```puppet
+nginx::resource::vhost { 'kibana.myhost.com':
+  listen_port => 80,
+  proxy       => 'http://localhost:5601',
+}
+```
+
+### A virtual host with static content
 
 ```puppet
 nginx::resource::vhost { 'www.puppetlabs.com':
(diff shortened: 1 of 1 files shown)


jfryman:
Fantastic! Thank you so much! :heart_eyes:

alexharv074:
Not at all.  Thanks for writing and maintaining the module.  4 lines of code for a working Nginx reverse proxy is great value. :)

PR #753, opened 2016-01-29
Add nginx_version fact

jyaworski:


3flex:
Please add a test for this. Can use https://github.com/puppetlabs/puppetlabs-mysql/commit/3ce8ad3799803cbd34d42a4b39ec27b914e9abd6 as a template.

jyaworski:
@3flex done. Thanks.

jyaworski:
Hmm. I'm not familiar with testing this particular fact type. Let me look into it; the test fails, but the fact works.

jyaworski:
OK, I got it. I forgot to stub :which.

3flex:
Looks good. Hopefully obvious, but did you confirm it works on a real machine?

jyaworski:
Using it in production now, actually.

3flex:
@jyaworski I'm proposing this be reverted (see #760) because it's not supported by earlier versions of Facter which are still supported by this module.

Are you able to make a version that is compatible back to Facter 1.7? If so, please submit a new PR. Thanks!

PR #762, opened 2016-02-05
nginx version fact compatibility with Facter 1.7

alexharv074:


alexharv074:
cc @3flex @jfryman @jyaworski

3flex:
Thanks @alexharv074.

I was going to ask for the Travis config to be updated so we can test on older Facter versions, but it seems that Puppetlabs aren't doing that in their own modules anymore. Looks good as is.

alexharv074:
:+1:

PR #763, opened 2016-02-05
travis: test with strict_variables on puppet 4

3flex:


3flex:
I expected something to break enabling this, just because it hadn't been enforced before and there are lots of variables in the module. But apparently the module is healthy which is great!

jfryman:
Woo hoo! :metal:

PR #765, opened 2016-02-08
Allow removal of gzip_types from the config

### LINKED ISSUE #748: duplicate MIME type "text/html" when starting nginx
When I restart Nginx:

```
[root@centos-66-x64 ~]# /etc/init.d/nginx restart
nginx: [warn] duplicate MIME type "text/html" in /etc/nginx/nginx.conf:38
Stopping nginx:                                            [  OK  ]
Starting nginx: nginx: [warn] duplicate MIME type "text/html" in /etc/nginx/nginx.conf:38
                                                           [  OK  ]
```

Which I think is [this issue](http://stackoverflow.com/questions/6475472/duplicate-mime-type-text-html).

3flex:
The current default value is not ideal, as 'text/html' is always included whether specified in the config or not. If it's specified, nginx shows a warning on start up.

Fixes #748

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -60,7 +60,7 @@ class nginx::config(
   $gzip_min_length                = 20,
   $gzip_http_version              = 1.1,
   $gzip_proxied                   = 'off',
-  $gzip_types                     = 'text/html',
+  $gzip_types                     = undef,
   $gzip_vary                      = 'off',
   $http_cfg_append                = false,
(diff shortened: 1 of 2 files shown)


ctavan:
Thanks for this patch! I have applied this patched locally to get rid of the warning and I would highly appreciate if this would be merged and released as part of the next puppet-nginx version.

ctavan:
That was quick, thank you!

PR #769, opened 2016-02-22
Adds new SSL && protocol specific directives to mailhost setup

dol:
The SSL setup customisation of a mailhost is very limited. To allow
more directives to be set, the ssl configuration was extracted and
arranged like the setup for a vhost.
The setup was also expanded to customize protocol specific authentication
method and capabilities.


bastelfreak:
Hi @dol, can you please rebase?

wyardley:
btw, I also made a change in #909 that updates the default ciphers for the module, and has the default ciphers in the mailhost manifest inherited from `nginx::config`, which I think it is how it should be. Can you integrate that change once it's merged?

dol:
@wyardley @bastelfreak Will perform an adjustment and rebase in the upcoming week.

wyardley:
Thanks @dol 
Note also that we prefer to have the commits squashed down to one (though if you want to leave a commit for the other author, that's fine).

wyardley:
#909 is now merged. See also #930, which was also merged, and takes `$ssl_protocols`'s default from `nginx::config`.

wyardley:
@dol: just a reminder to rebase / squash if you've got the time.

dol:
@wyardley I started with the rebase. Due to the divergency of the repo since February a rebase isn't that easy. Most of the work is done. But the test cases need some rearrangement. I'll keep you posted.
I also have to take care that the RP #909 and #930 aren't lost due to rearrangement/split up of the template. See the code comments in the stated PR's.

wyardley:
Thanks for looking at it, yeah, there have been a lot of changes, so sorry that the conflicts will be a bit of a pain to resolve!

It looks like I didn't notice that mailhost.erb had the SSL thing in #930 for the case where STARTTLS is enabled; that's a good catch. Seems almost like (even if they're under separate listeners), the SSL parameters should just be generated once (maybe an additional template that gets included twice?) to keep things in sync and to make it more DRY, but it's fine to just sync them up for now.

wyardley:
ps - Feel free to ping me or @bastelfreak on #voxpupuli in Puppet Slack or on IRC if you want to chat about it more.

dol:
I merged and updated the PR. Next steps are:
- Rebase into one commit
- Check the code and review it with my coworker

@wyardley The unification of the common parts are part of the initial PR. See https://github.com/voxpupuli/puppet-nginx/pull/769/files#diff-a9378fe4f575a8c57d256e37d135f6ccR38

wyardley:
@dol Thanks for the hard work! Will take a look over the code as well and see if anything sticks out.

Note that the tests are passing _except_ the rubocop; there are some warnings there that you'll need to resolve before we can merge. 
https://github.com/voxpupuli/puppet-nginx/blob/master/.github/CONTRIBUTING.md has a bit more info on that, but basically, `bundle exec rake rubocop` shouldn't report any failures.

Looking at the commits, it looks like maybe an unrelated commit made it in?  If it ends up being too difficult to squash, you can also make a new branch and create a new PR off of that one.

spec/defines/resource_mailhost_spec.rb
```diff
@@ -48,31 +48,31 @@
...
-          match: '  listen [::]:80 default ipv6only=on;'
+          match: '  listen                [::]:80 default ipv6only=on;'
         },
         {
           title: 'should not enable IPv6',
           attr: 'ipv6_enable',
           value: false,
-          notmatch: %r{  listen \[::\]:80 default ipv6only=on;}
+          notmatch: %r{  listen                \[::\]:80 default ipv6only=on;}
         },
```
wyardley: Is it possible to keep the spacing as it was before? In general, the module seems to use just a space or two after parameters, so I think it will be a little inconsistent if the spacing changes here. If we change the way we handle spacing, we'll probably need to do it on a module-wide basis.

spec/defines/resource_mailhost_spec.rb
```diff
@@ -136,29 +178,152 @@
...
+              lines = catalogue.resource('concat::fragment', "#{title}-header").send(:parameters)[:content].split("\n")
+              expect(lines & Array(param[:match])).to eq(Array(param[:match]))
             end
           end
         end
       end
     end
 
-    describe 'mailhost template content (SSL enabled)' do
+    describe "mailhost template content for imap" do
```
wyardley: I think Rubocop will also flag this, but use single quotes if there are no interpolated variables.

wyardley:
> @wyardley The unification of the common parts are part of the initial PR. See https://github.com/voxpupuli/puppet-nginx/pull/769/files#diff-a9378fe4f575a8c57d256e37d135f6ccR38

This is also great, thank you for making that change. Overall, I think if we can get this with the rubocop tests passing and commits squashed, I think we can merge this after one other person doing a quick review.

However, it does look like there have been some merges vs. rebases in the branch, so may be easier to just do a diff, and apply it to a new branch. Let us know if you need help getting it squashed. I know it's possible to do a squash merge in, but generally, Voxpupuli requires that the commits be squashed prior to merging.

bbriggs:
Looking much better. Can you please also address the complaints from Rubocop? IMO, it's easier to clean those up as you go.

wyardley:
@dol: Do you think you'll have time to rework this a bit more... re-rebase, sqash commits, etc?

Fixing the rubocop warnings isn't too hard.

If not, let me know and I can give it a shot.

I like breaking the common template out, but I'm thinking we should maybe use concat rather than embedding the template in another template where possible? @bastelfreak: thoughts?

I have a sort-of in-progress change going here:
https://github.com/wyardley/puppet-nginx/tree/cyon-fix/ssl_settings_for_mail
Specifically, these are some changes I applied on top, though a couple tests are still failing.
https://github.com/wyardley/puppet-nginx/commit/4bb90905ec3bdd04bfef95810dea45807938886b

dol:
@wyardley I found time to resolve the issues and performed some minor glitches that where introduced into the master during the last rebase.
About the single commit: I prepared an additional branch that contains all the changes in one commit. Squashing/rebasing was not easy possible. I'll push force the changes if you like. Normally the Github comments are lost when push forcing a branch.

dol:
@wyardley :+1:

wyardley:
@dol Thanks so much for persisting with this and making all those changes. I think this will really improve things in that section of the module!

PR #772, opened 2016-02-25
Add "satisfy" option to the location section

bernhardjt:
Add "satisfy" option to the location section 

https://www.nginx.com/resources/admin-guide/restricting-access/
http://nginx.org/en/docs/http/ngx_http_core_module.html?&_ga=1.238118336.1949645469.1456390584#satisfy


3flex:
Thanks for adding a test!

PR #775, opened 2016-03-02
Remove SSLv3 as it is insecure

ghoneycutt:


ghoneycutt:
Found some references to SSLv3 that were statically defined and not removed with the previous efforts to disable SSLv3

jfryman:
:heart:s and :lock:!

PR #767, opened 2016-02-15
uwsgi: allow custom uwsgi_param directives

darken99:
This patch allows custom uwsgi params to be added with uwsgi_param, same way fastcgi_param work.

PR #777, opened 2016-03-04
Added missing uwsgi_params HTTPS

ajcrowe:
This is in the default uwsgi_params file but is missing from the template. I see #767 will add custom params, but I think this should be in the defaults


bastelfreak:
Hi @ajcrowe, thanks for the PR. are you able to add spec tests for it?

wyardley:
Note that #855 also deals with this, specifically, it adds
```
+uwsgi_param REQUEST_SCHEME  $scheme;
+uwsgi_param HTTPS           $https if_not_empty;
```

wyardley:
Going to close this in favor of #855

PR #779, opened 2016-03-07
Upstream member array converted to hash to allow additional upstream member options

gtufte:
The following change gives an opportunity to define if an upstream member should be a backup member or not.

### DIFF
--- a/manifests/resource/upstream.pp
+++ b/manifests/resource/upstream.pp
@@ -40,7 +40,7 @@
 #    upstream_cfg_prepend => $my_config,
 #  }
 define nginx::resource::upstream (
-  $members = undef,
+  $members = {},
   $ensure = 'present',
   $upstream_cfg_prepend = undef,
   $upstream_fail_timeout = '10s',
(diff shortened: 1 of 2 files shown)


3flex:
Hi @gtufte, thanks for the contribution. Either the tests will need to be updated to reflect the new template, or the template should be tweaked so the existing tests pass.

Thanks!

wyardley:
Not sure if #755 affects this or creates any conflicts.
@gtufte: are you interested in creating tests for this, and if so, do you need assistance figuring out how to modify the tests? If not, this may be closed for inactivity after a month or so.

gtufte:
@wyardley It seems that #755 is more of a way to collect upstream backends based on a tag. My PR is more a way of adding additional upstream member settings that can be useful in some scenarios. The reason to why i made the member variable into a hash is that this scales and the nginx template will, in theory, only be the thing to update to add future upstream member settings provided by nginx.

I'm quite new to tests, but I've written some not long ago, I'll give it a shot by my own in the first instance, ty! :)

wyardley:
@gtufte: Fantastic.
Couple of notes:
1. There have been a lot of changes merged recently, so please rebase as well (your own commits should be squashed down to one, so needs to be a rebase instead of a merge).
1. Preserving backwards compatibility isn't super important, but you might consider following the pattern that much of the module follows where the parameter can be a hash _or_ an array; it will make the template a little more complex, but there are a lot of examples.

vinzent:
closing because of no updates for a longer time.

@gtufte feel free to re-open of you find time to rebase!

Issue #780, opened 2016-03-08
include /etc/nginx/streams-available|enabled not in nginx.conf.erb


alwiesner:
When attempting to build a configuration for a stream host, I am able to successfully build a  stream{} section in the main nginx.conf file, which correctly includes:  /etc/nginx/conf.stream.d/*.

When actually building the streamhost, it correctly gets placed in the /etc/nginx/streams-available directory/.  However, the stream{} section within nginx.conf, and how it is configured in the .erb template, does not include this directory (include etc/nginx/streams-available/*) when setting up the streamhost.

Can we add support for this?  Is there another way to do this that I am missing?

astrostl:
Looks like a working fix has been awaiting a merge for months over at https://github.com/jfryman/puppet-nginx/pull/784

wyardley:
I've just commented in #784. There is one open PR (#879) by someone else that was already merged that does a partial fix, and I did some other stuff related to this that may yet be merged (#878), but it does seem like support for this is pretty broken as-is.

wyardley:
See #933

wyardley:
This should be now be resolved in #933, which is in the most recent release (0.5.0)

PR #784, opened 2016-03-08
Fixing Streamhost implementation support

alwiesner:
There were issues with the streamhosts implementation for TCP load balancing. Changes made to support successful implementation of a stream host within NGINX configurations:

1) added "include <%= @conf_dir %>/streams-enabled/.conf;" to stream {} section of nginx.conf.erb. In order to enable nginx to listen on the defined TCP port, the stream{} configuration needs to include the /etc/nginx/streams-enabled/.conf files.

2) Removed 'server_name' from streamhost.erb template. The 'server_name' parameter is not valid for a stream host implementation. Removed line 20: server_name <%= @server_name.join(" ") %>;

3) nginx::resource::streamhost.pp does not ensure that the /etc/nginx/streams-enabled and /etc/nginx/streams-available directories exist before creating the stream host files. Added a file resource to ensure these directories are created, and adjusted the file creation resource to require this resource when creating the files.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -86,11 +86,12 @@ class nginx::config(
   $proxy_read_timeout             = '90',
   $proxy_redirect                 = 'off',
   $proxy_send_timeout             = '90',
-  $proxy_set_header               = [
-    'Host $host',
-    'X-Real-IP $remote_addr',
-    'X-Forwarded-For $proxy_add_x_forwarded_for',
-  ],
+  #$proxy_set_header               = [
+  #  'Host $host',
+  #  'X-Real-IP $remote_addr',
+  #  'X-Forwarded-For $proxy_add_x_forwarded_for',
+  #],
+  $proxy_set_header                = undef,
   $sendfile                       = 'on',
   $server_tokens                  = 'on',
   $spdy                           = 'off',
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -91,7 +91,7 @@ class nginx::config(
   #  'X-Real-IP $remote_addr',
   #  'X-Forwarded-For $proxy_add_x_forwarded_for',
   #],
-  $proxy_set_header                = undef,
+  $proxy_set_header               = [],
   $sendfile                       = 'on',
   $server_tokens                  = 'on',
   $spdy                           = 'off',
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -206,16 +206,14 @@ class nginx::config(
     ensure => directory,
   }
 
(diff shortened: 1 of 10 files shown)


alwiesner:
I believe this build failure is expected, since the server_name parameter was removed from streamhosts.erb, as it is an looks to be an invalid parameter for a tcp load balancing implementation with the stream directive.

Example error from nginx when attempting to start when a conf.stream.d conf file has a server_name parameter defined:

nginx: [emerg] "server_name" directive is not allowed here in /etc/nginx/streams-enabled/sdk.conf:4

astrostl:
Can this get a merge? Was about to develop and submit a similar fix!

jfryman:
@astrostl If you're willing to pick up the PR and fix up the tests, we can absolutely get this merged in.

There might be a bad test. We should fix that, not merge in code that isn't passing.

astrostl:
Gotcha. Are passing PRs usually quick-merged?

3flex:
Just some feedback on this PR - there seems to be several changes unrelated to the original intent of fixing the streamhost support. These should be removed from this PR and added to a new PR as it changes several long-standing behaviours of the module. Thanks!

bastelfreak:
HI @alwiesner, we recently merged https://github.com/voxpupuli/puppet-nginx/pull/790 which a bit of the functionality from this PR. Please have a look and rebase.

wyardley:
#879 did some of this; I also played around with adding some more tests and doing this in a slightly different way in #878, however, as issue #780 points out, it doesn't look as if the streams related directories currently get included.

I got rid of the '$stream' variable in #878, though maybe it should be reintroduced if it's actually going to be implemented as expected (I'd think it should be sufficient to just include the relevant directories, but maybe some other things could drop stuff in there, and that's why there was a `$stream`?

wyardley:
@astrostl: are you still interested in working on stream support? As of now, there are some additional changes which should fix some of the problems mentioned here (see above), but not all. However, there have been enough changes since then that I'm not sure if it's easier to start over, or rework this one.

astrostl:
I'm inactive on that feature and active on other things at work, so I'd pass for now.

wyardley:
@astrostl, no worries, I've got the same thing going on a lot of the time.
@alwiesner: I think I intended that for you anyway.

For the folks who understand Nginx stream support, do you think there's any reason to have the $streams option that I got rid of recently? Or should we just include the streams directory in `/etc/nginx/conf.d` always, and have streams exist if someone creates the appropriate resource?

wyardley:
Replacing with #933 
Would _love_ if anyone has time to test 933 + current master, but I think this hits most of the major points.

PR #786, opened 2016-03-17
add support for passenger on CentOS/RHEL

### LINKED ISSUE #633: Passenger Packages for CentOS/RHEL!
I am excited this morning. I got an email which says that Phusion is now producing packages for RHEL and CentOS. See blog post here: https://blog.phusion.nl/2015/05/20/phusion-passenger-5-0-8-new-debian-packages-red-hat-packages-bug-fixes/

Tommy

jlambert121:
Closes #633

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,6 @@ Gemfile.lock
 spec/fixtures/modules
 spec/fixtures/manifests
 vendor
+.bundle
 .ruby-version
 .vagrant/
--- a/README.markdown
+++ b/README.markdown
@@ -146,9 +146,7 @@ nginx::nginx_mailhosts:
 
 ## Nginx with precompiled Passenger
 
-Currently this works only for Debian family and OpenBSD.
-
-On Debian it might look like:
+On Debian and CentOS it might look like:
 ```puppet
 class { 'nginx':
   package_source  => 'passenger',
--- a/manifests/package/redhat.pp
+++ b/manifests/package/redhat.pp
@@ -53,8 +53,24 @@ class nginx::package::redhat (
           before   => Package['nginx'],
         }
       }
+      'passenger': {
+        yumrepo { 'passenger':
+          baseurl  => "https://oss-binaries.phusionpassenger.com/yum/passenger/el/${::operatingsystemmajrelease}/\$basearch",
+          descr    => 'passenger repo',
+          enabled  => '1',
+          gpgcheck => '1',
+          priority => '1',
+          gpgkey   => 'https://packagecloud.io/gpg.key',
+          before   => Package['nginx'],
+        }
+
+        package { 'passenger':
+          ensure  => 'present',
+          require => Yumrepo['passenger'],
+        }
+      }
       default: {
-        fail("\$package_source must be 'nginx-stable' or 'nginx-mainline'. It was set to '${package_source}'")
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -6,5 +6,6 @@ Gemfile.lock
 spec/fixtures/modules
 spec/fixtures/manifests
 vendor
+.bundle
```
3flex: Shouldn't need any changes to gitignore for this update

3flex:
We'll need some additional logic, since passenger is only supported on RedHat/CentOS 6 & 7. As is, this would allow installation on unsupported OSs (e.g. Scientific Linux, Amazon...) that share `osfamily` of RedHat.

I'd only want this to be automatically installed on platforms that Passenger officially supports.

The README should then also state that's automatic installation is supported on Debian family, RedHat and CentOS only.

3flex:
Looks good otherwise!

wyardley:
@jlambert121: Are you planning to make the changes suggested by @3flex? This PR would be very useful for us as well.

.gitignore
alexjfisher: `.gitignore` is also managed centrally using modulesync

wyardley:
I don't know if they were signing their packages before, but I'm getting this error when using it:

```
Error: Execution of '/bin/yum -d 0 -e 0 -y install passenger' returned 1: Package passenger-5.0.30-8.el7.x86_64.rpm is not signed
Error: /Stage[main]/Nginx::Package::Redhat/Package[passenger]/ensure: change from absent to present failed: Execution of '/bin/yum -d 0 -e 0 -y install passenger' returned 1: Package passenger-5.0.30-8.el7.x86_64.rpm is not signed
```

so might be good to add an acceptance test? 

https://oss-binaries.phusionpassenger.com/yum/definitions/el-passenger.repo
has:
```
name=passenger
baseurl=https://oss-binaries.phusionpassenger.com/yum/passenger/el/$releasever/$basearch
repo_gpgcheck=1
gpgcheck=0
enabled=1
gpgkey=https://packagecloud.io/gpg.key
sslverify=1
sslcacert=/etc/pki/tls/certs/ca-bundle.crt
```

It doesn't look to me like the package is signed at all, so I don't think it's just a matter of importing the package's GPG signing key vs. the repo metadata's. Not sure if it's easier to file a PR against the fork that this is coming from, but this patch seems to resolve the installation issue (at the cost of installing a package that's not signed).

[el_passenger_patch_1.txt](https://github.com/voxpupuli/puppet-nginx/files/473623/el_passenger_patch_1.txt)

wyardley:
With the change above and rebased against current master: https://github.com/voxpupuli/puppet-nginx/compare/master...wyardley:el_passenger

I can take a stab at doing the additional logic to making the other requested changes above, but could use some guidance as to where that logic should happen, i.e., does the logic need to happen in params and then set something that we can look at in `manifests/package/redhat.pp`, or if we should just do the case / if statements directly in `manifests/package/redhat.pp`.

OR, we could just remove support for EL5....

ffrank:
@jlambert121 this PR is about to be superseded by #876, which includes a commit with your changes, attributed to you. Will that be fine?

3flex:
Closed by #876

PR #787, opened 2016-03-18
Add support for 'default' vhosts

heliocentric:
Allow a user to specify an empty server_name array. This allows you to configure a 'fallback' vhost if no host header matches, which is useful if NGINX is an ssl termination point for something like varnish.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -49,6 +49,14 @@ nginx::resource::vhost { 'www.puppetlabs.com':
   www_root => '/var/www/www.puppetlabs.com',
 }
 ```
+### Creating a landing page if no other vhost matches:
+
+```puppet
+nginx::resource::vhost { 'fallback-host':
+  server_name => [],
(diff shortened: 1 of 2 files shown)


wyardley:
Just curious, I think '_' is accepted as a server_name (http://nginx.org/en/docs/http/server_names.html - see "catch-all"), and provides catch-all functionality. Is that a reasonable workaround?

Whether it's due to server_name '_' or an empty server_name, it could cause unpredictable behavior in either case if more than one vhost has this set. That said, maybe we could just set it to '_' if it's not set, or document that users should use '_' in the README or examples?

wyardley:
@heliocentric: Does using '_' as (one of) the server names work for you?
If so, maybe open a new PR that's a docs fix that explains this somewhere?

heliocentric:
It will, and I'll make a doc pr instead

PR #788, opened 2016-03-25
Add support for multiple 'proxy_cache_valid' directives

hbog:
Nginx supports multiple proxy_cache_valid directives.
This allows to specify a different caching time for different response codes.
(http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_valid)

This PR adds the possibility to specify an array of strings for the 'proxy_cache_valid' attribute, while preserving backwards compatibility by still accepting a string (which is the expected behavior today). 

Also, a unit test has been added that covers the new behavior, hence both the current and the new behavior is tested.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -296,7 +296,7 @@ define nginx::resource::location (
     validate_string($proxy_cache_use_stale)
   }
   if ($proxy_cache_valid != false) {
-    validate_string($proxy_cache_valid)
+    validate_slength(any2array($proxy_cache_valid),12,1)
   }
   if ($proxy_method != undef) {
     validate_string($proxy_method)
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -296,7 +296,9 @@ define nginx::resource::location (
     validate_string($proxy_cache_use_stale)
   }
   if ($proxy_cache_valid != false) {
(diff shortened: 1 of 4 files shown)


3flex:
Looks like by using `validate_slength` with `12` as the value of the second argument that strings are limited to a maximum length of 12 characters. Seems this could easily be exceeded.

Is there a better way to validate the array values are all strings? Sure you can increase the character limit but it seems like a workaround.

hbog:
The maximum length is indeed an arbitrary number and 12 may be too small. Increasing the number helps. A string longer than 64 characters for example is extremely unlikely for this directive.

An alternative is to stick to `validate_array`, which is also used elsewhere in the module to validate parameters containing an array of strings.  This has the benefit of consistency with the rest of the code.

If you agree, I will replace `validate_slength` by `validate_array`.

3flex:
`validate_array` doesn't add much value here though if you're using `any2array`, since we already know the output of that function is an array.

Ideally you would check the input is either a string or an array first, then either use `any2array` to ensure a string is turned into an array so the template works properly, or handle either a string or an array of strings in the template.

Whichever results in cleaner code is best but either way is OK with me.

hbog:
The validation now checks for string or array first, using a pattern that I borrowed from similar validations elsewhere in the module.

PR #794, opened 2016-04-02
Make certain SSL related parameters optional (to allow using an include)

gerases:


jfryman:
Can you also ensure that this passes test for Puppet 4?

gerases:
I'm afraid that would be difficult to set up in my environment :( I don't
even have puppet4

3flex:
I think failures are due to STRICT_VARIABLES on the Puppet 4 Travis config, not a problem with the PR.

I think STRICT_VARIABLES may not have been enforced properly, but it is now due to a new release of rspec-puppet.

Can either disable STRICT_VARIABLES or fix the module. I don't have time to fix it at the moment, so will leave it to @jfryman to either fix the errors or stop checking strict variables.

3flex:
Also I'm not a fan of the `defined_elsewhere` option. I think if ssl_cert and ssl_key are now optional that the `fail` check should be removed.

Makes it more flexible and means we can avoid adding another parameter which as designed wouldn't be used by many people.

gerases:
Are you saying you will remove the fail check?

gerases:
The other problem with validation is that I do define quite a few pieces in
an include deployed by a process over which I have no control. And it
includes quite a few SSL configuration pieces, including allowed ciphers,
protocols, etc. What's the best way to handle that if not through some
construct allowing exclude certain pieces from validation?

manifests/resource/vhost.pp
```diff
@@ -271,7 +271,8 @@
   $mode                         = $::nginx::config::global_mode,
   $maintenance                  = false,
   $maintenance_value            = 'return 503',
-  $locations                    = {}
+  $locations                    = {},
+  $defined_elsewhere            = [],
```
3flex: `defined_elsewhere` should be removed

manifests/resource/vhost.pp
```diff
@@ -536,7 +537,9 @@
 
   # Check to see if SSL Certificates are properly defined.
   if ($ssl == true) {
-    if ($ssl_cert == undef) or ($ssl_key == undef) {
+    $cert_elsewhere = member($defined_elsewhere, 'ssl_cert')
+    $key_elsewhere =  member($defined_elsewhere, 'ssl_key')
+    if ($ssl_cert == undef and !$cert_elsewhere) or ($ssl_key == undef and !$key_elsewhere) {
       fail('nginx: SSL certificate/key (ssl_cert/ssl_key) and/or SSL Private must be defined and exist on the target system(s)')
```
3flex: I think this `fail` should be removed, or downgraded to a warning with appropriate text (which we'll have to think about)

wyardley:
@gerases: This will still need some more review, but doing the following would help:
1) Give the PR a more descriptive subject
2) Rebase against current master
3) Add tests

Let us know if you need help knowing what to test or how (https://github.com/voxpupuli/puppet-nginx/blob/master/.github/CONTRIBUTING.md also has some details)

wyardley:
For now, I think #945 resolves this, or at least accomplishes the same goals. If you'd like to rework it to be more clever, please open a new PR (with tests) based on current master.

PR #798, opened 2016-04-11
add ssl_verify_client parameter

rchicoli:
the ssl_verify_client variable was hardcoded to 'on', now it is possible to pass other values like "on | off | optional | optional_no_ca".


3flex:
Thanks for the code @rchicoli!

Couple of things:

* Test at https://github.com/jfryman/puppet-nginx/blob/master/spec/defines/resource_vhost_spec.rb#L930 is now failing. See Travis results e.g. https://travis-ci.org/jfryman/puppet-nginx/jobs/122227324). This line needs to be updated since the spacing in the output of the file is changed, so the existing matcher isn't matching.
* Please try adding some tests for the new parameter. You could model after these tests: https://github.com/jfryman/puppet-nginx/blob/master/spec/defines/resource_vhost_spec.rb#L390-L401

Cheers!

rchicoli:
Thx for pointing out the error @3flex 
* that was pretty easy to fix
* the ssl_verify_client parameter has to belong to some contexts, because it will only be set if ssl_client_cert is given  

Glad to contribute.

PR #812, opened 2016-06-03
No reasons to manage separate files since confd_purge is available

vchepkov:
Module also prevents managing content of default files
if configuration files installed by rpm just removed, without supplying a sanitized safe content, they will be re-installed during rpm update,
this will leave site broken until next puppet run

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -227,14 +227,6 @@ class nginx::config(
     }
   }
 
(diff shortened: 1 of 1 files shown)


wyardley:
#900 is this, but with the tests for those resources being absent removed, should address the Travis failures.

wyardley:
Merged #900

PR #817, opened 2016-06-13
Add manage_service parameter

iain-buclaw-sociomantic:
Puppet managing the service is not desirable in every environment.

Use `?w=1` to omit whitespace changes.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -112,6 +112,7 @@ class nginx (
   $service_flags                  = undef,
   $service_restart                = '/etc/init.d/nginx reload',
   $service_name                   = undef,
+  $service_manage                 = true,
   ### END Service Configuration ###
 
   ### START Hiera Lookups ###
@@ -294,6 +295,7 @@ class nginx (
     service_restart   => $service_restart,
     service_name      => $service_name,
     service_flags     => $service_flags,
+    service_manage    => $service_manage,
   }
 
   create_resources('nginx::resource::upstream', $nginx_upstreams)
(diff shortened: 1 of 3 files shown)


manifests/service.pp
```diff
@@ -35,24 +36,26 @@
...
-  case $::osfamily {
-    'OpenBSD': {
-      service { 'nginx':
-        ensure     => $service_ensure_real,
-        name       => $service_name,
-        enable     => $service_enable,
-        flags      => $service_flags,
-        hasstatus  => true,
-        hasrestart => true,
+  if $manage_service == true {
```
3flex: Please use `if $manage_service {`, it's a little cleaner

manifests/init.pp
```diff
@@ -112,6 +112,7 @@
   $service_flags                  = undef,
   $service_restart                = '/etc/init.d/nginx reload',
   $service_name                   = undef,
+  $manage_service                 = true,
```
3flex: Please rename the parameter `service_manage` to follow the convention for the other service-related parameters.

3flex:
Thanks for the contribution!

I've added some notes inline. Tests for this functionality would also be very very useful, take a look at `/spec/classes/service_spec.rb` and see if you're able to add a new test. It would be similar to adding a block like this https://github.com/jfryman/puppet-nginx/blob/955cdc7aac744141c8be82e1f4b9e7efbbb9dadb/spec/classes/service_spec.rb#L24-L42

iain-buclaw-sociomantic:
Added test in `service_spec.rb`, and made changes as per notes.  Rebased it against master also.

I hope the test is fine, I wasn't able to check locally.

3flex:
Looks solid, thank you!

iain-buclaw-sociomantic:
Thanks!

PR #820, opened 2016-06-16
Don't generate http locations if rewrite_to_https

iain-buclaw-sociomantic:
The configuration I'm trying to produce being:

```
server {
  listen 80;
  server_name  host;
}

server {
  listen 443 ssl;
  server_name host;

  ssl on;

  location /foo {
  }
  location /bar {
  }
  # ...
```

Locations don't need to be present in both http and ssl servers.

### DIFF
--- a/manifests/resource/vhost.pp
+++ b/manifests/resource/vhost.pp
@@ -569,7 +569,7 @@ define nginx::resource::vhost (
     notify => Class['::nginx::service'],
   }
 
-  $ssl_only = ($ssl == true) and (($ssl_port + 0) == ($listen_port + 0))
+  $ssl_only = ($ssl == true) and (($ssl_port + 0) == ($listen_port + 0) or $rewrite_to_https)
 
   if $use_default_location == true {
     # Create the default location reference for the vHost


marcofl:
this is a pretty common use-case. I would +1 for such a implementation, but for now I'm just doing it by passing ssl_port = 443 and listen_port = 443 on the ssl only vhost. then just add a non ssl vhost with the redirect.

wyardley:
@marcofl the current $rewrite_to_https param does do the redirect, @iain-buclaw-sociomantic is just saying that the locations can be suppressed in the non-ssl listener, since all traffic is being redirected.

@iain-buclaw-sociomantic: do you know how to add tests for this? https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/CONTRIBUTING.md has some info, and specifically, I think you'd mostly be adding stuff in `spec/defines/resource_vhost_spec.rb` (and / or `resource_location_spec.rb`).

wyardley:
@iain-buclaw-sociomantic: I have reworked this somewhat in #957, hope that's Ok... if you get a chance, take a look and see if it looks reasonable to you.

wyardley:
Superceded by #957

PR #821, opened 2016-06-16
Add ssl_session_tickets and ssl_session_ticket_key parameters

iain-buclaw-sociomantic:
Support for http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_tickets

### DIFF
--- a/manifests/resource/vhost.pp
+++ b/manifests/resource/vhost.pp
@@ -79,6 +79,10 @@
 #     OCSP responses by the server. Defaults to false.
 #   [*ssl_session_timeout*] - String: Specifies a time during which a client
 #     may reuse the session parameters stored in a cache. Defaults to 5m.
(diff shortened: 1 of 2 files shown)


manifests/resource/vhost.pp
```diff
@@ -206,6 +210,8 @@
   $ssl_stapling_responder       = undef,
   $ssl_stapling_verify          = false,
   $ssl_session_timeout          = '5m',
+  $ssl_session_tickets          = 'on',
```
3flex: This should default undef, and be conditional in the template.

Since it was introduced in 1.5.9, and some distros support older versions than that, this directive would be unrecognized by nginx and prevent it from starting.

iain-buclaw-sociomantic: Rebased with suggested default.

PR #829, opened 2016-07-12
Pin rake to last supported version for ruby 1.8.7

samuelson:
The latest version of rake does not support ruby 1.8.7, so some of the tests were failing.  This pins rake to the last supported version: '10.5'.


3flex:
Thanks @samuelson but this module supports Ruby 1.9.3 and above, with 1.8.7 support as YMMV.

samuelson:
@3flex ok that's fine with me, can we remove it from the testing matrix?

samuelson:
Oh, nm, looks like it was already :)

PR #832, opened 2016-07-14
Remove storeconfigs warning on puppet apply

### LINKED ISSUE #720: "You cannot collect exported resources without storeconfigs being set" at manifests/resource/upstream.pp:89:5
Hello.

When I start my puppet script to create a virtual machine, I get the error message above. I feed your module with the following configuration:
```yaml
upstreams:
        nxv_394dk0458afs:
            name: django
            upstream_fail_timeout: 10s
            members:
                - '127.0.0.1:8000'
```
Yes, the YAML syntax derived from a generated template of PuPHPet (https://puphpet.com). I still think the configuration is valid as the other modules in your project work perfectly. The error messages mentions location.pp explicitly.

This error does not keep me from creating my machine and makes this a low-priority issue.

sorreltree:
This is an attempt to fix issue #720 .  I can only do minimal testing as I don't have a master environment to test with.  I believe this should fix the issue and retain functionality, based on other modules that have addressed the same problem.

Fixes #720

### DIFF
--- a/manifests/resource/upstream.pp
+++ b/manifests/resource/upstream.pp
@@ -97,8 +97,10 @@ define nginx::resource::upstream (
       content => template('nginx/conf.d/upstream_members.erb'),
     }
   } else {
-    # Collect exported members:
-    ::Nginx::Resource::Upstream::Member <<| upstream == $name |>>
+    class { 'nginx::resource::upstream::collect':
+      # Collect exported members
+      upstream_name => $name,
+    }
   }
 
   concat::fragment { "${name}_upstream_footer":
--- a/manifests/resource/upstream/collect.pp
+++ b/manifests/resource/upstream/collect.pp
@@ -1,0 +1,12 @@
+# Class to collect exported upstream members, if needed.
+#
+# This is separated into a separate class in order to remove errors if
+# storeconfigs are unavailable.
+
+class nginx::resource::upstream::collect (
+  $upstream_name,
+)
+{
+  # Collect exported members:
+  ::Nginx::Resource::Upstream::Member <<| upstream == $upstream_name |>>
+}


3flex:
Sounds reasonable, based on what I've seen in https://github.com/ssm/ssm-munin/pull/33 for example.

I'd really like someone to test this though! I don't have a master setup currently either. Will ask in thread for #720.

3flex:
In the meantime, can you please squash the commits to a single commit? Thank you!

sorreltree:
Sorry about the multiple commits -- I thought I had them squashed.

Should be squashed now.

sorreltree:
Since I can verify that the patch does in fact disable the error, and since upstream inclusion seems to be well tested in the Travis tests (see block below), I'm bumping this pull request for reconsideration.

```
nginx::resource::upstream

  os-independent items

    basic assumptions

      should contain Concat[/etc/nginx/conf.d/upstream-test-upstream.conf]

      should contain Concat::Fragment[upstream-test_upstream_header] with content =~ /upstream upstream-test/

      should contain Concat::Fragment[upstream-test_upstream_header] with target => "/etc/nginx/conf.d/upstream-test-upstream.conf" and order => 10

      should contain Concat::Fragment[upstream-test_upstream_members] with target => "/etc/nginx/conf.d/upstream-test-upstream.conf" and order => 50

      should contain Concat::Fragment[upstream-test_upstream_footer] with target => "/etc/nginx/conf.d/upstream-test-upstream.conf", order => 90 and content  supplied string

    upstream.conf template content

      when upstream_cfg_prepend is {"test3"=>"test value 3", "test6"=>{"subkey1"=>["subvalue1", "subvalue2"]}, "test1"=>"test value 1", "test2"=>"test value 2", "test5"=>{"subkey1"=>"subvalue1"}, "test4"=>["test value 1", "test value 2"]}

        should contain Concat[/etc/nginx/conf.d/upstream-test-upstream.conf] with mode => "0644"

        should contain Concat::Fragment[upstream-test_upstream_header]

        should contain ordered prepended directives

      when members is ["test3", "test1", "test2"]

        should contain Concat[/etc/nginx/conf.d/upstream-test-upstream.conf] with mode => "0644"

        should contain Concat::Fragment[upstream-test_upstream_members]

        should set server

      when ensure => absent

        should contain Concat[/etc/nginx/conf.d/upstream-test-upstream.conf] with ensure => "absent"
```

sorreltree:
Yay!

PR #833, opened 2016-07-19
Mitigation for US-CERT VU#797896

abrander:
US-CERT VU#797896 (and others) describes an attack where an end-user is capable of overriding the proxy setting for multiple FastCGI backends.

I propose we help mitigate this attack by removing the variable before reaching the backend as proposed by multiple vendors and industry players.

Sources:
https://www.kb.cert.org/vuls/id/797896
https://httpoxy.org/
https://www.nginx.com/blog/mitigating-the-httpoxy-vulnerability-with-nginx/

Relevant CVEs:
CVE-2016-5385: PHP
CVE-2016-5386: Go
CVE-2016-5387: Apache HTTP Server
CVE-2016-5388: Apache Tomcat
CVE-2016-1000109: HHVM
CVE-2016-1000110: Python

### DIFF
--- a/templates/vhost/fastcgi_params.erb
+++ b/templates/vhost/fastcgi_params.erb
@@ -25,3 +25,8 @@ fastcgi_param  HTTPS              $https;
 
 # PHP only, required if PHP was built with --enable-force-cgi-redirect
 fastcgi_param  REDIRECT_STATUS    200;
+
+# Workaround for US-CERT VU#797896. There appear to be no real reason
+# to forward the Proxy: header to FastCGI backends. We simply remove it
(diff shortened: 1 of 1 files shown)


FlorianSW:
This doesn't cover the case, where you proxy the request to another server, e.g. if you use nginx only for ssl termination and the actual execution of any application is made internally by Apache or another nginx server or whatever crazy setups there're used in the wild. So I would propose to set
`proxy_set_header Proxy "";`
just in case, too (in `templates/vhost/locations/proxy.erb`).  There is nearly no reason I can imagine, where you would need this header and if so, we probably can make it configurable but enabled by default.

abrander:
I think this is better solved by #835 - I'll close this PR.

PR #835, opened 2016-07-22
Mitigating Httpoxy 

marcofl:
Should possibly go together with https://github.com/jfryman/puppet-nginx/compare/httpoxy-test
If you guys are already working on a httpoxy fix, just reject this pull request. 
Thanks.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -92,6 +92,7 @@ class nginx::config(
     'Host $host',
     'X-Real-IP $remote_addr',
     'X-Forwarded-For $proxy_add_x_forwarded_for',
+    'Proxy ""',
   ],
   $proxy_hide_header              = [],
   $sendfile                       = 'on',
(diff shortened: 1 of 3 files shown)


dasrecht:
Looks good to me 👍

FlorianSW:
Looks good to me  :ok_hand:

jfryman:
Stellar. Tyvm for the code!

dominics:
Note that this fix was reverted as part of #862

That seems to be because Nginx recommend overriding the header just for PHP requests (https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/#connecting-nginx-to-php-fpm) - not a great solution IMO considering there is exactly zero legitimate use for a `Proxy` header in HTTP.

And the "upstream version" copied as part of that predates the (re)discovery of the vulnerability.

Upshot is that when people upgrade to 0.7.0 of this library, their mitigation will come undone.

PR #851, opened 2016-08-11
Conditionally adding the packages if they are not added previously

Faffnir:
We faced the problem that when using the puppet-nginx module and managing ca-certificates on our own it will bring up problems only because of assigning the same resource twice.

(ensure_resources() would be nice to use but I didn't want to add additional dependencies)

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -53,10 +53,9 @@ class nginx::package::debian(
           key      => '16378A33A6EF16762922526E561F9B9CAC40B2F7',
         }
 
-        package { ['apt-transport-https', 'ca-certificates']:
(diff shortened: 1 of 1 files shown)


3flex:
Would using ensure_packages work for you? If so I'd rather use that for apt-transport-https and ca-certificates. However the passenger package should not use ensure_packages because by installing it with this module the assumption is that the package would not be declared elsewhere.

Faffnir:
Absolutely correct and thank you for pointing me to it!

3flex:
That's great, thanks! Could you please squash the two commits to a single commit and I'll get this merged.

Faffnir:
All done!

PR #852, opened 2016-08-12
Merge Request #851 introduced a wrong order of installation

Faffnir:
Sorry... Introduced a wrong order with the fix/PR yesterday.


3flex:
I should have noticed that as well! Thanks for picking up on it.

PR #855, opened 2016-08-18
Adding some params to the uwsgi template that I needed. And some formating.

jostmart:
Added some params to the uwsgi template that I needed. And some simple formating.

### DIFF
--- a/templates/vhost/uwsgi_params.erb
+++ b/templates/vhost/uwsgi_params.erb
@@ -1,15 +1,19 @@
 # This file managed by puppet on host <%= @fqdn %>
 
-uwsgi_param QUERY_STRING $query_string;
-uwsgi_param REQUEST_METHOD $request_method;
-uwsgi_param CONTENT_TYPE $content_type;
-uwsgi_param CONTENT_LENGTH $content_length;
-uwsgi_param REQUEST_URI $request_uri;
-uwsgi_param PATH_INFO $document_uri;
-uwsgi_param DOCUMENT_ROOT $document_root;
+uwsgi_param QUERY_STRING    $query_string;
+uwsgi_param REQUEST_METHOD  $request_method;
+uwsgi_param CONTENT_TYPE    $content_type;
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @jostmart, could you add tests to the changes?

ffrank:
Is this safe? Can we be certain that this has no negative impact for existing setups?

jyaworski:
I don't see any problem with this. It looks like it's just adding:

```
+uwsgi_param REQUEST_SCHEME  $scheme;
+uwsgi_param HTTPS           $https if_not_empty;
```

ffrank:
Yes. Is it conceivable that a module user does specifically not want either of those?

wyardley:
@jostmart: will you be able to add tests? If there's no more activity in about a month, this will be closed. Would the fix in #767 work instead? It's generic, but allows adding arbitrary uwsgi_params, which might be more flexible.

I don't deal with nginx / uwsgi together much, but I agree with @ffrank that it might be better not to default to adding them if they're not explicitly set.

wyardley:
@jostmart I think in principle this is fine to add; are you able to add tests?

wyardley:
I'm not even sure that tests are needed here, now that I think about it, but the rename is causing a conflict.

ffrank:
Well, to make this easier: The three added parameters take very sensible values, so I think it should be fine to just merge this.
@jostmart will you be able to take care of those merge conflicts? I'm in favor of just merging once that is done.

wyardley:
@ffrank agreed, but so far, no response (I'm guessing it just needs to be moved to `templates/server`?

Does someone want to take a stab at reworking it and re-submitting (with attribution to the original author)?

jostmart:
Hi I will try to look at it in a couple of days. The change I made have been working great for me since I added it.

vinzent:
@jostmart  any chance to look into rebasing?

wyardley:
If you don't mind, can we pick this up in #1076?

wyardley:
closing via #1076

PR #864, opened 2016-08-31
Make uwsgi_params non-executable

ekohl:


igalic:
tests are failing because rspec update broke… everything.

Issue #538, opened 2014-12-30
Get more friendly with concat


jfryman:
>As an aside I think we should move away from these methods to insert custom code into the manifests, and promote puppetlabs-concat as the best way to do this since it allows for total customization of any code that will be added to the vhost/location.

Want to capture a comment from @3flex here https://github.com/jfryman/puppet-nginx/pull/532#issuecomment-68370797 to track for a later day. :grin:

wyardley:
I think we are now friendly with concat. 😉

PR #870, opened 2016-09-02
Add option http_cfg_prepend

abraham1901:
Add option http_cfg_prepend


ffrank:
Hi, thanks for the patch.

Are our tests generally broken?

What's the use case here? Is the append option insufficient?

In any case, the commits need squashing.

abraham1901:
Hi
  > Are our tests generally broken?
  
  Yes
  
  > What's the use case here? Is the append option insufficient?
  
  I'm have a trouble sort option in http_cfg_append.
  For example if i'm use include custom directory with config and need added limit_conn_zone used in included config - it's not posible.
  In config now:
  
  include                      /etc/nginx/customdir/*.conf;
  limit_conn_zone $server_name zone=perserver:100m;
  
  And receive syntax nginx error. But need another sequence:
 
  limit_conn_zone $server_name zone=perserver:100m;
  include                      /etc/nginx/customdir/*.conf;

>In any case, the commits need squashing.
Done.

manifests/init.pp
```diff
@@ -245,6 +246,7 @@
       fastcgi_cache_use_stale        => $fastcgi_cache_use_stale,
       gzip                           => $gzip,
       http_access_log                => $http_access_log,
+      http_cfg_prepend                => $http_cfg_prepend,
```
3flex: This is causing the Travis tests to fail, because the `=>` is not aligned with the rest. Puppet Lint is running in Travis and checks for this.

abraham1901:
Sorry, fixed

templates/conf.d/nginx.conf.erb
```diff
@@ -36,6 +36,14 @@ events {
 }
 
 http {
+<% if @http_cfg_append -%>
```
3flex: Should be checking for `http_cfg_prepend`

abraham1901: +

ffrank:
Thanks @3flex. Now that this is fix, would you give this a thumbs up? Because if so, I'm inclined to squash-merge this as is now.

3flex:
If there's a valid use case, I say go ahead and merge.

I have a personal issue with all the append/prepend parameters throughout the module, but until something is done about https://github.com/voxpupuli/puppet-nginx/issues/538 there's not really a better alternative. That's a big job though.

One other possibility is to stop sorting the directives in the templates like we have been, but apparently even on Ruby 1.9.3 there were issues in the past with non-stable ordering, I think Puppet itself might reorder sometimes and makes no guarantees about that unfortunately.

That said, :+1: from me!

ffrank:
Great, thanks for the summary.

Thanks to @abraham1901 for the contribution!

PR #873, opened 2016-09-10
[keepalive_requests] added keepalive_requests parameter in nginx.conf

shoeb751:
set it to default value of 100 as shown in nginx config
Require this parameter as default value results in a huge number of TIME_WAIT connections with high QPS applications hitting nginx
The check has been copied from the keepalive_timeout parameter

### DIFF
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -62,6 +62,7 @@ http {
   server_names_hash_max_size <%= @names_hash_max_size %>;
 
   keepalive_timeout  <%= @keepalive_timeout %>;
+  keepalive_requests  <%= @keepalive_requests %>;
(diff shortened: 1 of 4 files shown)


templates/conf.d/nginx.conf.erb
```diff
@@ -62,6 +62,7 @@ http {
   server_names_hash_max_size <%= @names_hash_max_size %>;
 
   keepalive_timeout  <%= @keepalive_timeout %>;
+  keepalive_requests  <%= @keepalive_requests %>;
```
alexjfisher: nitpick - indentation

alexjfisher:
LGTM Default is indeed `100` http://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_requests

PR #879, opened 2016-09-22
Add missing stream dirs and create streams from hiera

andybotting:
This creates the missing streams-available and streams-enabled
directories if stream is true and allow the streamhost resources
to be created from hiera.

Some simple tests are included.

This commit might replace PRs:
 - Fix streams support #790
 - Fixing Streamhost implementation support #784

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -264,6 +264,33 @@ class nginx::config(
     owner  => $daemon_user,
   }
 
+  if $stream {
+    file { "${conf_dir}/streams-available":
+      ensure => directory,
+      owner  => $sites_available_owner,
+      group  => $sites_available_group,
+      mode   => $sites_available_mode,
+    }
+
+    if $vhost_purge == true {
+      File["${conf_dir}/streams-available"] {
+        purge   => true,
+        recurse => true,
+      }
+    }
+
+    file { "${conf_dir}/streams-enabled":
+      ensure => directory,
+    }
+
+    if $vhost_purge == true {
+      File["${conf_dir}/streams-enabled"] {
+        purge   => true,
+        recurse => true,
+      }
+    }
+  }
+
   file { "${conf_dir}/sites-available":
     ensure => directory,
     owner  => $sites_available_owner,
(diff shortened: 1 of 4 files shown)


bastelfreak:
Thanks for the PR @andybotting!

wyardley:
@andybotting: This affects #878 somewhat. I've rebased against your PR, but would love it if you can take a look and see if my changes seem sane (and if you have any suggestions to make the code cleaner or less spaghetti like).

I'm adding some additional tests related to this in there as well.

wyardley:
What's the point of only creating certain dirs if `$stream = true`, when it will still try to create the resource when defined if `$stream = false`? Maybe the parameter needs to be changed, or just deprecate get rid of the $stream parameter entirely? @bastelfreak @andybotting 

When creating the streams-available dir is wrapped in an `if $stream` block
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb#L36

I get: ```
  1) nginx with defaults should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
       error during compilation: Could not retrieve dependency 'File[/etc/nginx/streams-available]' of Concat[/etc/nginx/streams-available/streamhost1.conf]
     # ./spec/classes/nginx_spec.rb:22:in `block (3 levels) in <top (required)>'
```
after adding a 'require' on that file (something I added in #878, modeling what we did in #906).

I'm kind of of a mind to just get rid of / deprecate the `if $stream` bits, or, if we really want this logic, maybe rename it to make it clear that it's just ensuring that directory that the parameter is doing?

andybotting:
@wyardley I was aiming for the least amount of changes to get the functionality I needed, so I just kept that conditional for the directory. Personally, I'm not a big fan of the Debian style *-available *-enabled directory thing, so I like your changes in #878.

I wouldn't change this now that its merged, but I would also prefer to just always create the stream directories in this case and forget about the $stream conditional.

wyardley:
_I wouldn't change this now that its merged, but I would also prefer to just always create the stream directories in this case and forget about the $stream conditional._

Yes, in the latest iteration of #878, I ended up just dropping `$stream` entirely, since it wasn't really doing much except controlling whether those dirs were ensured (and we're already ensuring a lot of other dirs anyway, so I don't see that behavior as being particularly harmful).

I don't use this feature much, so would be very appreciative if you could give the stream-related changes in #878 at some point (especially the spec tests I added) to make sure things look reasonable.

PR #881, opened 2016-09-22
Add additional config variables with default values (update of #693)

wyardley:
This is a rework of @victorgp's #693 that rebases against current master, and adds tests for the new parameters (the two reasons that PR seems to have been stalled).

I also removed what appear to be a few duplicate tests (let me know if I'm wrong), and realigned dashes in one section of the docs, though mostly tried not to change too much formatting.

I squashed his and my commits, though left attribution for @victorgp's original commit. I don't think I can commit back to the PR since I'm not in the org, but let me know if that's the preferred way to do it (vs. submitting a new PR).

I tried to follow the existing conventions, but It would be great if someone could review my changes to the tests, as well as review the formatting and ordering of directives within nginx.conf.


wyardley:
$error_pages should have defaulted to 'undef' instead of {}.
otherwise, since the template had
`<% if @error_pages -%>` instead of `<% unless @error_pages.empty? -%>`
the template would print extra lines even if no error pages were defined. I'm updating the fix for this in #888, even though it's not directly related.

wyardley:
@pschichtel: We have a few PRs in progress that should go out first, but shooting to cut a Forge release pretty soon. It'll be a big one, with a few breaking changes.

Hoping to get this out within the next week or so.

pschichtel:
Cool, thanks for the update!

Issue #882, opened 2016-09-22
acceptance tests with new(ish) Beaker version


wyardley:
- Puppet: 4.6.2
- Ruby: 2.3.1p112
- Distribution: Mac OSX 
- Module version: [master]

### How to reproduce (e.g Puppet code you use)
run `bundle exec rake acceptance`

### What are you seeing
`/usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/dsl/helpers/host_helpers.rb:81:in `block in on': DSL method `on` can only be called with a String or Beaker::Command object as the command parameter, not NilClass. (ArgumentError)`

### What behaviour did you expect instead
acceptance tests to run

### Output log
```
/usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/dsl/helpers/host_helpers.rb:81:in `block in on': DSL method `on` can only be called with a String or Beaker::Command object as the command parameter, not NilClass. (ArgumentError)
	from /usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/shared/host_manager.rb:127:in `run_block_on'
	from /usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/dsl/patterns.rb:37:in `block_on'
	from /usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/dsl/helpers/host_helpers.rb:63:in `on'
	from /Users/wby/git/puppet-nginx/spec/spec_helper_acceptance.rb:5:in `block in <top (required)>'
	from /Users/wby/git/puppet-nginx/spec/spec_helper_acceptance.rb:3:in `each'
	from /Users/wby/git/puppet-nginx/spec/spec_helper_acceptance.rb:3:in `<top (required)>'
	from /Users/wby/git/puppet-nginx/spec/acceptance/class_spec.rb:1:in `require'
	from /Users/wby/git/puppet-nginx/spec/acceptance/class_spec.rb:1:in `<top (required)>'
	from /usr/local/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.3/lib/rspec/core/configuration.rb:1435:in `load'
	from /usr/local/lib/ruby/gems/2.3.0/gems/rspec-core-3.5.3/lib/rspec/core/configuration.rb:1435:in `block in load_spec_files'
[...]

PR #883, opened 2016-09-23
Fix acceptance test failures with newer Beaker versions

wyardley:
(re #882): Still getting a failure, but putting in this PR for when we get acceptance tests working with newer Beaker version. Presumably we should also require a specific version more explicitly?


wyardley:
cc: @3flex

3flex:
Verified locally, thanks!

PR #884, opened 2016-09-24
remove unmanaged nodesets

3flex:
Remove all nodesets which aren't managed in voxpupuli's modulesync_config (https://github.com/voxpupuli/modulesync_config/tree/master/moduleroot/spec/acceptance/nodesets)

This means all will be updated automatically going forward.

Note that until a `default.yml` is added to modulesync_config, users running acceptance tests will have to specify a nodeset using `BEAKER_set=` e.g. `BEAKER_set=ubuntu-server-1204-x64 bundle exec rake beaker`


3flex:
To avoid issues with missing `default.yml` it might be preferred to wait until https://github.com/voxpupuli/modulesync_config/pull/229 or something like it is merged, and wait for the next module sync.

But given the acceptance tests haven't been run at all for many months at least, impact of merging before that happens will be minimal.

dhoppe:
We should also remove `default.yml`, because this file is not managed via `module_sync` and the user should be able to choose the operating system as he likes. You can not provide Ubuntu as default, if the user prefers CentOS or anything else.

3flex:
@dhoppe since I did remove`default.yml` from this repo in this PR, can we merge?

PR #885, opened 2016-09-24
Configure acceptance tests on docker on travis

3flex:
Depends on https://github.com/voxpupuli/modulesync_config/pull/230

It's safe to merge without that change and a new module sync, but the changes in `.sync.yml` would be redundant.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -59,6 +59,16 @@ HaHIWGMBuXApE7t4PNdYWZ5Y56tI+HT59yVoDjc1YSnuzkKlWUPibVYoLpX/ROKr
 aIZ8kxsBjLvpi9KQTHi7Wl6Sw3ecoYdKy+2P8S5xOIpWjs8XVmOWf7Tq1+9KPv3z
 HLw/FDCzntkdq3G4em15CdFlO9BTY4HXiHU=
 -----END CERTIFICATE-----" > /tmp/blah.cert'
+
+      #--- START SELINUX WORKAROUND ---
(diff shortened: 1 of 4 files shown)


petems:
Merging as this is good for Vagrant VM's as well as Docker. The discussion around enabling PR's for Travis is a larger one that requires consensus. 

This is Safe, Small and Sensible so merging after a local test with a Vagrant box:

```
Finished in 1 minute 40.02 seconds (files took 2 minutes 7 seconds to load)
43 examples, 0 failures
```

spec/spec_helper_acceptance.rb
```diff
@@ -59,6 +59,16 @@
 aIZ8kxsBjLvpi9KQTHi7Wl6Sw3ecoYdKy+2P8S5xOIpWjs8XVmOWf7Tq1+9KPv3z
 HLw/FDCzntkdq3G4em15CdFlO9BTY4HXiHU=
 -----END CERTIFICATE-----" > /tmp/blah.cert'
+
+      #--- START SELINUX WORKAROUND ---
```
wyardley: What about just dropping the certs into /etc in the first place, and then just using /etc/blah.{crt,key} as the path? Then we don't have to use this whole workaround at all, assuming selinux allows those files to be in /etc
Or symlink /etc/pki to /etc/ssl early in the process for Debian?

I also see some discussion about disabling SELinux with Beaker here: https://bugzilla.redhat.com/show_bug.cgi?id=1074384

petems: Confusingly, that is a different Beaker testing project, run by RedHat 😄 https://beaker-project.org/

PR #886, opened 2016-09-27
more test and docs fixes for acceptance tests for CentOS / Passenger

wyardley:
@3flex @ffrank 
Here's one more related to the now-merged #876 
@3flex says that #885 should address the other test failure (though there will probably be a conflict if one isn't rebased after the other is merged).

I also added the `--fail` (`-f`) option to curl for the vhost test calls that rely on exit status, which should return a non-zero exit code if there's a non-2XX response. Normally, curl will exit non-zero if there's a failure to connect in the first place, but if it connects and gets a 4XX or 5XX response, it will display the error but exit 0.
```
[root@centos-7-x64 tmp]# curl localhost
<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.10.1</center>
</body>
</html>
[root@centos-7-x64 tmp]# echo $?
0
[root@centos-7-x64 tmp]# curl -f localhost
curl: (22) The requested URL returned error: 502 Bad Gateway
[root@centos-7-x64 tmp]# echo $?
22
```

Should figure out if we want to explicitly add some kind of dependency on epel or just document it as I've done in the README.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -15,7 +15,7 @@ class{'nginx':
     package_source => 'nginx-mainline'
 }
 ```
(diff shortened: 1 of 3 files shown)


petems:
```
BEAKER_set='centos-72-x64' bundle exec rake acceptance
Finished in 3 minutes 11.8 seconds (files took 1 minute 52.78 seconds to load)
42 examples, 0 failures
```

PR #890, opened 2016-09-29
Custom fact fails when using openresty

shoikan:
When a modified Nginx (for instance Openresty: http://openresty.org/en/) is installed, the custom fact fails, regardless of the puppet code for that host actually uses any code of the module.

The regular Nginx responds to -v: 'nginx version: nginx/1.6.3'. The Nginx in Openresty responds: 'nginx version: openresty/1.11.2.1'.

### DIFF
--- a/lib/facter/nginx_version.rb
+++ b/lib/facter/nginx_version.rb
@@ -2,7 +2,7 @@ Facter.add(:nginx_version) do
   setcode do
     if Facter.value('kernel') != 'windows' && Facter::Util::Resolution.which('nginx')
       nginx_version = Facter::Util::Resolution.exec('nginx -v 2>&1')
-      %r{nginx version: nginx\/([\w\.]+)}.match(nginx_version)[1]
+      %r{nginx version: (nginx|openresty)\/([\w\.]+)}.match(nginx_version)[2]
     end
   end
 end
--- a/manifests/resource/vhost.pp
+++ b/manifests/resource/vhost.pp
@@ -152,6 +152,9 @@
 #     backend application (Passenger 5.0+)
 #   [*passenger_env_var*]       - Allows one to set environemnt variables to pass
 #     to the backend application (Passenger 5.0+)
+#   [*passenger_pre_start*]     - Allows setting a URL to pre-warm the host. Per
+#     Passenger docs, the "domain part of the URL" must match a value of
+#     server_name
 #   [*log_by_lua*]              - Run the Lua source code inlined as the
 #     <lua-script-str> at the log request processing phase.
 #     This does not replace the current access logs, but runs after.
@@ -272,6 +275,7 @@ define nginx::resource::vhost (
   $passenger_cgi_param          = undef,
   $passenger_set_header         = undef,
   $passenger_env_var            = undef,
+  $passenger_pre_start          = undef,
   $log_by_lua                   = undef,
   $log_by_lua_file              = undef,
   $use_default_location         = true,
@@ -510,6 +514,9 @@ define nginx::resource::vhost (
   }
   if ($passenger_env_var != undef) {
     validate_hash($passenger_env_var)
+  }
+  if ($passenger_pre_start != undef) {
+    validate_string($passenger_pre_start)
   }
   if ($log_by_lua != undef) {
     validate_string($log_by_lua)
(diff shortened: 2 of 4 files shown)


lib/facter/nginx_version.rb
```diff
@@ -2,7 +2,7 @@
   setcode do
     if Facter.value('kernel') != 'windows' && Facter::Util::Resolution.which('nginx')
       nginx_version = Facter::Util::Resolution.exec('nginx -v 2>&1')
-      %r{nginx version: nginx\/([\w\.]+)}.match(nginx_version)[1]
+      %r{nginx version: (nginx|openresty)\/([\w\.]+)}.match(nginx_version)[1]
```
alexjfisher: Needs to be `%r{nginx version: (nginx|openresty)\/([\w\.]+)}.match(nginx_version)[2]` ?

petems: Yep, adding an extra capture zone means you have to change this to the second selector as Alex says.

The tests are failing because of this:

```
 1) Facter::Util::Fact with current version output format should eq "1.8.1"
     Failure/Error: it { expect(Facter.fact(:nginx_version).value).to eq('1.8.1') }
     
       expected: "1.8.1"
            got: "nginx"
     
       (compared using ==)
     # ./spec/unit/facter/util/fact_nginx_version_spec.rb:11:in `block (3 levels) in <top (required)>'
```

petems:
@shoikan From my quick testing, this fact would not work, as the which check would not find nginx, so the fact would never be found.

You'd have to change it to something like this:

```ruby
Facter.add(:nginx_version) do
  setcode do
    if Facter.value('kernel') != 'windows'
      if Facter::Util::Resolution.which('nginx')
        nginx_version = Facter::Util::Resolution.exec('nginx -v 2>&1')
        %r{nginx version: nginx\/([\w\.]+)}.match(nginx_version)[1]
      end
      if Facter::Util::Resolution.which('openresty')
        nginx_version = Facter::Util::Resolution.exec('nginx -v 2>&1')
        %r{nginx version: openresty\/([\w\.]+)}.match(nginx_version)[1]
      end
    end
  end
end
```

shoikan:
@petems In openresty, the binary is still called nginx, so .which should still find it. It is only the output of the -v that is formatted differently. That is... Ruby isn't my strong suit, so I might be understanding that method incorrectly?

alexjfisher:
@shoikan Thanks, that looks better.  Do you think you could squash your commits?

shoikan:
@alexjfisher I gave it a try :)

petems:
@shoikan The which command looks to see if something with the given name is in the path:

```
# which nginx
which: no nginx in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
# which openresty
/usr/bin/openresty
```

This is on CentOS, is the binary called nginx on other OS's?

alexjfisher:
@shoikan Something's gone a little wrong with your rebase/squash.  You've gained an unrelated commit from somewhere.  It's always worth using a branch other than `master` in your fork when working on PRs.  Try not to create any commits on master other than the ones you sync from upstream.

But don't worry.  if the change is good, somebody at voxpupuli can rebase it for you if you're struggling.

shoikan:
@petems Odd! I compile openresty from source (we want some more LUA stuff added, plus a few plugins), but for me the binary is still called nginx, not openresty! I will do some more investigating....

wyardley:
Updated in #908

wyardley:
Per @3flex's comments, reverted my hastily merged PR, sorry for not noticing the additional comments in this one, beyond the requested changes.

I was just going to do a fix based on https://github.com/voxpupuli/puppet-nginx/pull/890#issuecomment-250581776, in addition to adding a test.
But I realized while doing a quick test on a vagrant box that both nginx and openresty (via the official RPM) can be installed:
```
# openresty -v
nginx version: openresty/1.11.2.1
# nginx -v
nginx version: nginx/1.10.1
```
Of course, there are other ways this could happen if multiple nginx binaries are installed (via SCL or similar systems, installed to an alternate path, etc.), but even @petems's fix leaves some ambiguity here. If the order were reversed, at least the nginx one would prevail if both exist, but that still doesn't seem ideal.

So what's the ideal behavior, and what should happen if both nginx and openresty are installed?

3flex:
Maybe one fact for openresty version, one for nginx?

wyardley:
_Maybe one fact for openresty version, one for nginx?_

That's what I was wondering, but would it be less useful then?

Here's one way that copies and just very slightly changes the behavior of @petems's fix, plus adds some tests (not sure if I need to add the stub for openresty version in the 3rd one since we don't expect to call it, or if that test is overkill):

https://github.com/voxpupuli/puppet-nginx/compare/master...wyardley:shoikan_pr_890_v3

petems:
Closing in favour of #890

PR #898, opened 2016-10-04
add auth_http_header

tjikkun:
Add the ability to specify auth_http_header in mailhost

### DIFF
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -57,6 +57,7 @@ define nginx::resource::mailhost (
   $starttls            = 'off',
   $protocol            = undef,
   $auth_http           = undef,
+  $auth_http_header    = undef,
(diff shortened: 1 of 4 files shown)


wyardley:
@tjikkun: can you add tests for this?
It would probably just be a few lines in `spec/defines/resource_mailhost_spec.rb`

tjikkun:
@wyardley Thanks for looking, I have added tests now

wyardley:
Thanks for the quick changes! Looks good to me

dol:
The documentation of the class parameter is missing in the manifests/resource/mailhost.pp file. If it's ok I'll add it in the #769 rebase.

wyardley:
@dol: adding docs is always a good thing. We're hoping to rework the docs in general, since the module obviously could be better documented, but definitely feel free to add it in your PR.

Issue #896, opened 2016-10-03
Defaults


ubellavance:
Hi,

is there a reason why the module doesn't use the binary's default values?  I think that a module should never set defaults other than the binary's unless it is necessary.  By necessary, I mean that nginx doesn't start at all if this or these parameters are not set correctly.

As an example, the module sets accept_mutex to on, while it is disabled by default in [nginx](http://nginx.org/en/docs/ngx_core_module.html#accept_mutex) and in the package's nginx.conf file.

There are many parameters like that...  I have checked on RHEL and Ubungu 16.04 and I don't see that many parameters in the conf file.

Thanks,

wyardley:
The particular example seems to have come from #747. That claimed that it used defaults, so either the nginx default has changed, or that was a mistake. @mlrobinson: any idea about that? I've created #903 to address that specific example.

As far as the more general problem, it is an issue when defaults get in the module vs. just leaving them out of the file, but in some cases it's necessary. From what I've seen, folks have tried to use default settings when they add parameters; this module has a lot of contributions from a lot of different people, so things do occasionally fall through the cracks. If you want to come up with a full list of params that don't match the defaults, we can try to get things synced up. Or, if you'd like to take a crack at correcting them in a pull-request, adjusting any tests as needed, we can take a look at that.

ubellavance:
Thanks for the PR for that, but I don't think that is the solution.  As you can see [here](http://nginx.org/en/docs/ngx_core_module.html#accept_mutex), "Prior to version 1.11.3, the default value was on." It means that nginx 1.10 default was on.  Most systems have nginx < 1.11.3, so almost all installs has "on" as default, so putting it to yes would break most installs. The more, more general problem is that, in my opinion, a module shouldn't re-state the defaults of the moment, it should just not set any parameter unless necessary.  As an example, you can look at [this module](https://github.com/edestecd/puppet-clamav/blob/master/manifests/params.pp). Clamd has a million parameters, but the module only defines the ones that are necessary and let the users define anything that he doesn't want to be at default value.

Defining defaults can be usefull to have a more consistent setup amongst a bunch of different kind of installs (different OSs, different nginx versions), but I think that it should be the responsibility of the puppet admin to check what are the defaults in the version that is to be installed and configure the module parameters accordingly.

It is also less of a management burden for the module maintainers because they don't need to track default values changes, and nginx seems not to be very static on the default values.

wyardley:
[...]

PR #747, opened 2016-01-26
Add events accept mutex delay

mlrobinson:
Adding the option to manage accept mutex and accept mutex delay. Defaults taken from here: http://nginx.org/en/docs/ngx_core_module.html#accept_mutex.

PR #903, opened 2016-10-07
accept_mutex default is off, not on

wyardley:
See also #896 and #747 
Not sure if the default has changed, but since that was fairly recent, I doubt it has... this sets what seems to be the correct default setting for accept_mutex.


wyardley:
D'oh, prior to 1.11.3, the default _was_ on, and many / most systems using this module will have 1.10.x or older. We could do this based on the nginx_version fact, but agree that just taking the param out might be a better fix.

Issue #895, opened 2016-10-03
log_dir works in vhost context, but not in main context


ubellavance:
Hi,

I defined nginx::config::conf_dir and it works perfectly for vhosts (it puts access log in $log_dir/vhostname.access.log), but if we also want to use a different path for the base nginx logs, we have to explicitely define them, even when nginx::config::conf_dir is defined.

nginx::config::log_dir: '/var/opt/rh/rh-nginx18/log/nginx'
nginx::http_access_log: '/var/opt/rh/rh-nginx18/log/nginx/access.log'
nginx::nginx_error_log: '/var/opt/rh/rh-nginx18/log/nginx/error.log'

wyardley:
You mean `log_dir`, not `conf_dir`, right?

I see what you're saying tho. it's getting pulled directly from params.pp, so config.pp override isn't being looked at. This is a good one, I'll work on a PR to fix this.

https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/params.pp#L98-L100
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/config.pp#L26-L28

wyardley:
I'm proposing one fix for this, but would like some input from others about whether there's a better way to accomplish this. It works in my tests if you'd like to test it out as well.

ubellavance:
Yes, you got it right.  It is log_dir.  Thanks a lot for your work on that one. With your change, will it still be possible to override it? I think I should be able to test your proposed code next week.


 Also, could you demystify something for me: I don't know if that's the case when setting the configs in the classes, but when using Hiera, some parameters are nginx::paramX and some others are nging::config::paramY.  Is there a reason why?

Example from my config:

Directly under nginx::
```
nginx::server_tokens: 'off'
````
Needs a "config" before:
````
nginx::config::client_body_temp_path: '/var/opt/rh/rh-nginx18/lib/nginx/tmp/client_body'
```

wyardley:
As far as the history on which configs are in nginx::config, I think the idea was to reduce the clutter in params, as well as to separate the parameters configuring the module itself (i.e., where nginx is installed from, say), vs. the parameters that relate to configuring nginx itself.

You don't have to use hiera, but you may need to use the weird pattern used in `doc/hiera.md` if you want to alter default configs.

There was also at one point an attempt to move to 'puppet-module-data' pattern; this text from an old version of `doc/hiera.md` explains it somewhat:
[...]

PR #904, opened 2016-10-08
fix for log_dir not being honored (#895)

wyardley:
Requesting review on this possible fix for #895.

Two things:
1) This seems to "work", and the other ways I could think of to do it didn't seem more appealing, but should any of this be moved out of the class parameters?
2) I'd like to add some additional spec tests for this, any thoughts on what / how to test?


spec/classes/config_spec.rb
```diff
@@ -691,5 +691,11 @@
 
       it { expect { is_expected.to contain_class('nginx::config') }.to raise_error(Puppet::Error, %r{\$nginx_error_log_severity must be debug, info, notice, warn, error, crit, alert or emerg}) }
     end
+
+    context 'when log_dir is non-default' do
+      let(:params) { { log_dir: '/foo/bar' } }
+
+      it { is_expected.to contain_file('/foo/bar').with(ensure: 'directory') }
```
bastelfreak: could you add tests for the $http_access_log and $nginx_error_log variables as well? not sure if they contain the correct path.

wyardley: How's that look?

Issue #894, opened 2016-10-03
No require for File: sites-enabled and sites-available folders


ubellavance:
Hi,

I think that this module lacks a set of require directives for the sites-enabled and sites-available folders.  On ther first puppet run on a freshly installed VM, the puppet run doesn't complete if puppet choses to create the files before the directory.

Error: Could not set 'file' on ensure: No such file or directory - /etc/opt/rh/rh-nginx18/nginx/sites-available/atqtestssl.traceability.ca.conf20161003-4240-zx4832
Error: Could not set 'file' on ensure: No such file or directory - /etc/opt/rh/rh-nginx18/nginx/sites-available/atqtestssl.traceability.ca.conf20161003-4240-zx4832
Wrapped exception:
No such file or directory - /etc/opt/rh/rh-nginx18/nginx/sites-available/atqtestssl.traceability.ca.conf20161003-4240-zx4832
Error: /Stage[main]/Nginx/Nginx::Resource::Vhost[atqtestssl.traceability.ca]/Concat[/etc/opt/rh/rh-nginx18/nginx/sites-available/atqtestssl.traceability.ca.conf]/File[/etc/opt/rh/rh-nginx18/nginx/sites-available/atqtestssl.traceability.ca.conf]/ensure: change from absent to file failed: Could not set 'file' on ensure: No such file or directory - /etc/opt/rh/rh-nginx18/nginx/sites-available/test.example.com.conf20161003-4240-zx4832

Please note that I use non-standard folders location because I use the rh-nginx18 package from Red Hat Software Collections, but I think that the module shouldn't assume that the sites-available and sites-enabled folders exist.

If I run puppet a second time, it works OK.

wyardley:
I think something like the attached might resolve it
[vhost_patch.txt](https://github.com/voxpupuli/puppet-nginx/files/517432/vhost_patch.txt)

Presumably #878 would resolve your issue (just in a different way), right?  Not 100% sure this would be totally compatible with that. Have you figured out how to test that one yet?

ubellavance:
I tested it once and it went OK, no error at all.  Without the patch it failed the every time the 5-6 times that I tried.  I'll try to test a few other times. I simply rollback to a snapshot that is just before the first puppet run.  It looks like you've done a very good job again.

wyardley:
Cool, #906 is there and is passing tests, so after I get some more feedback on it, should be able to get it merged.

ubellavance:
[...]

PR #906, opened 2016-10-08
'Require' vhost dir / enable dir in files

wyardley:
I think this might help with #894.
It passes spec and acceptance tests, but would like a review of this to make sure it's sane and makes sense. I couldn't think of an easy way to test this, but let me know if there's a test I should add.


bastelfreak:
Thanks @wyardley!

PR #908, opened 2016-10-08
Prevent custom fact from complaining when openresty is installed

wyardley:
Update of #890 that should apply to current master and preserve original authorship of commit.
@shoikan @alexjfisher 

(I tried making a pull request against the pull request, but because that master is now out of date again, it didn't work).


wyardley:
Merging since @alexjfisher already approved in #890

3flex:
This didn't address https://github.com/voxpupuli/puppet-nginx/pull/890#issuecomment-250581776?

wyardley:
@3flex: Unless I'm missing something, I think @shoikan already resolved it earlier, it's a [2], not a [1] in:
https://github.com/voxpupuli/puppet-nginx/blob/master/lib/facter/nginx_version.rb#L5

There was no test for it, though, which seems bad. I opened #911. If you think the regex needs to be fixed, let me know in there.

Issue #859, opened 2016-08-24
secure ssl configuration


octomike:
I was reading about [sweet32](https://www.openssl.org/blog/blog/2016/08/24/sweet32/) and wondered how this module keeps track of bad and good openssl cipher settings. Turns out the last update for this was in June 2015. What's the policy here? Can I just pull request a current list, e.g. according to [Mozilla's SSL Configuration Generator](https://mozilla.github.io/server-side-tls/ssl-config-generator/)?

ekohl:
:+1: on fixing this. The default config produces a B in ssllabs.com. I'd recommend submitting a PR with working tests.

wyardley:
Agree that this should be a priority. I was just looking through the module for something else and noticed that the cipher list looked out of date, and that TLSv1 was supported still.
(https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices has some examples).

the nginx default seems simple: "HIGH:!aNULL:!MD5;". Have not tested this one, what do folks think about that option?

https://cipherli.st/ seems probably a bit too strict.
https://support.cloudflare.com/hc/en-us/articles/200933580-What-cipher-suites-does-CloudFlare-use-for-SSL-
doesn't seem like a bad place to start.

I can start a PR for this, but probably needs some good testing beyond running acceptance tests on a few platforms.

wyardley:
Opened #909 for this, but would love for people to do some functional testing on this if at all possible. If someone can stand this up somewhere public and run the Qualys labs tests or similar against it, that would also be helpful.

This does seem to past existing acceptance tests on Ubuntu 12 and CentOS 6 (wanted to try against somewhat older stuff; couldn't get the CentOS 5 beaker test to work, though).

wyardley:
I have updated it to the defaults in:
https://mozilla.github.io/server-side-tls/ssl-config-generator/
specifying default (1.0.1e) as baseline OpenSSL version.

xaque208:
Thank you.

wyardley:
@xaque208 Just a heads up that the PR is still not merged. If anyone has a system they can test this config (or the PR) against that's publicly reachable and has a commercial cert, that would be great.

xaque208:
As long as FreeBSD compatibility has not broken during the transition to voxpupuli, I can likely test an evening this week.

ekohl:
Now https://github.com/voxpupuli/puppet-nginx/pull/909 has been merged, can this be closed?

[...]

PR #909, opened 2016-10-08
SSL cipher changes (issue 859)

wyardley:
This is my first stab at updating the SSL cipher list (re: #859)
I suppose we can't / shouldn't disable TLS 1.0 yet?


ekohl:
https://mozilla.github.io/server-side-tls/ssl-config-generator/ still enables TLSv1 on the intermediate profile which I think is a good default for the module.

Have you checked this against https://www.ssllabs.com/ as well? Did it give an A rating?

wyardley:
@ekohl: I didn't have an easy and quick way to expose a test box using the Puppet generated config to the network... Would love it if someone has time to test, but if not, I can come up with something early next week.

wyardley:
TODO: once cipher list is finalized, also need to update `vhost.pp` with new defaults in docs.

wyardley:
https://mozilla.github.io/server-side-tls/ssl-config-generator/ does look reasonable (though not sure what OpenSSL version we should specify as highest, I'm reasonably sure we don't have 1.0.1e everywhere). 

Someone on IRC suggested that site as well. I am taking that recommendation; updated docs and also had mailhost use nginx::config::ssl_ciphers as default rather than hard-coding it in the template.

templates/mailhost/mailhost_ssl.erb
```diff
@@ -28,14 +28,9 @@ server {
   ssl on;
   ssl_certificate      <%= @ssl_cert %>;
   ssl_certificate_key  <%= @ssl_key %>;
-
   ssl_session_timeout  5m;
 
   ssl_protocols TLSv1;
```
ekohl: I like the re-use of the ```$nginx::config::ssl_ciphers```. Should we do the same for ```ssl_protocols``` here?

jyaworski: Absolutely.

templates/mailhost/mailhost_ssl.erb
```diff
@@ -28,14 +28,9 @@ server {
...
-
-  # Suggested from https://wiki.mozilla.org/Security/Server_Side_TLS#Nginx
-  ssl_ciphers 'ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA:AES256-SHA:AES:CAMELLIA:DES-CBC3-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:!PSK:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-CBC3-SHA';
-
+  ssl_ciphers <%= @ssl_ciphers %>;
```
dol: The same change applies to the  puppet-nginx/templates/mailhost/mailhost.erb file. Because the SSL part of the two templates are equal I used to combine the two into one file that gets included. I'll fix the missing variable in my PR #769.

PR #911, opened 2016-10-09
add test for openresty nginx version fact

wyardley:
#890 and #908 didn't include a test, seems like that might be a good idea.

@3flex: if you think there's still something that needs to be fixed in the regex per @alexjfisher's comments, let me know, and I'll fix it here as well (and squash).

### DIFF
--- a/spec/unit/facter/util/fact_nginx_version_spec.rb
+++ b/spec/unit/facter/util/fact_nginx_version_spec.rb
@@ -17,4 +17,11 @@ describe Facter::Util::Fact do
     end
     it { expect(Facter.fact(:nginx_version).value).to eq('0.7.0') }
   end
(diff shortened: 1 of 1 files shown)


3flex:
No issue with the regex. But following #890 there was this comment from petems: https://github.com/voxpupuli/puppet-nginx/pull/890#issuecomment-250581776

With a normal openresty install, there is no nginx binary in the path (per petems comment and https://openresty.org/en/rpm-packages.html), meaning `Facter::Util::Resolution.which('nginx')` will be false and the fact would never be run.

petems proposed a solution in their comment.

wyardley:
Sorry, thought I clicked the comment link before, but didn't get taken to that comment. But I see what you're talking about now.

I will revert my merged PR, and re-open the original one.

PR #913, opened 2016-10-09
Adds ability to detect modified nginx for fact

petems:
Rebase and refactor of #890. 

Adds ability to detect modified nginx for fact
    * For instance openresty: http://openresty.org/en/) is installed.
    * Openresty can be installed from a package or compiled
    * This allows it to have two possible paths: nginx or openresty
    * Fact and spec updated to detect both paths
    * Regex updated to capture openresty output
    * Added spec for no nginx or openresty install
    * Moved spec to correct path (not in util folder)


lib/facter/nginx_version.rb
```diff
@@ -1,8 +1,9 @@
 Facter.add(:nginx_version) do
   setcode do
-    if Facter.value('kernel') != 'windows' && Facter::Util::Resolution.which('nginx')
-      nginx_version = Facter::Util::Resolution.exec('nginx -v 2>&1')
-      %r{nginx version: nginx\/([\w\.]+)}.match(nginx_version)[1]
+    if Facter.value('kernel') != 'windows' && (Facter::Util::Resolution.which('nginx') || Facter::Util::Resolution.which('openresty'))
+      nginx_version_command = Facter::Util::Resolution.which('nginx') ? 'nginx -v 2>&1' : 'openresty -v 2>&1'
+      nginx_version = Facter::Util::Resolution.exec(nginx_version_command)
+      %r{nginx version: (nginx|openresty)\/([\w\.]+)}.match(nginx_version)[2]
```
wyardley: Any reason not to just do
`%r{nginx version: (?:nginx|openresty)\/([\w\.]+)}.match(nginx_version)[1] ` here?

Or capture it but create a new fact nginx_flavor or similar?

PR #914, opened 2016-10-09
set manage_repo to "false" by default

### LINKED ISSUE #863: RFC: Upstream vs distro packages
This module is inconsistent in the way it chooses which package source to install nginx from. By default, the package from the distro will be installed, unless (with current `params.pp`) the distro is Ubuntu Lucid, Precise or Trusty; Debian 6, 7 or 8; or Red Hat/CentOS 5, 6 or 7, in which case upstream nginx.org stable package will be installed.

The current setup means the module in principle prefers to install upstream package from nginx.org, but because `params.pp` must be updated as nginx updates the OSs they support, and this hasn't been happening, it creates a situation where an Ubuntu Trusty user will get nginx.org package, but a Xenial user gets the distro package, even though nginx.org makes one available.

Also, while nginx.org builds packages for SLES 12, this module doesn't use them, though it could be updated to do so.

Options, as I see them, to dealing with this more consistently:

1. Keep status quo. Module will theoretically install upstream stable package by default, and rely on more eyes on the module to ensure `params.pp` is updated in line with nginx.org OS support documented at http://nginx.org/en/linux_packages.html. Perhaps this is not much of an issue as I don't recall any recent issues/PRs to update this file, but if this module is all of a sudden updated it will change the package source for Ubuntu Xenial users which might be unexpected.
2. Use distro package by default. Requires no ongoing maintenance in this module.
3. Install upstream stable package by default for Ubuntu, Debian, Red Hat and CentOS, but do not hard code that support in `params.pp`. Only issue with this is that on day 1 of a new OS release nginx.org may not have packages, meaning the module will fail to install anything.

My personal preference is 2. It's easiest to maintain and predictable for users. Those who want upstream packages can easily change their config to get them. 1 or 3 is theoretically better for security for those distros that don't provide security support for nginx, but Ubuntu does since Trusty and Debian always has. I don't know Red Hat or CentOS's policy.

A further wrinkle is the new support for dynamic modules. nginx.org is building some (http://nginx.org/en/linux_packages.html#dynmodules). Debian is testing them (see libnginx-mod-* packages) and presumably they'll be in Debian 9 and future Ubuntu releases (likely from 17.04).

3flex:
Closes #863 

From now on, this will have to be manually enabled by users according to the
needs of their environment.

cc @wyardley 

As I was updating the spec tests, it occurred to me that if this parameter defaults to "false", that it's probably redundant - if a user were to set a value for `package_source` then we already know that they must set `manage_repo => true` (so it's redundant). So maybe a better approach is to add a warning so that anyone who currently has `manage_repo` set to "true" is notified that in a future release they must explicitly set `package_source`, and that `manage_repo` is going away.

I think this is better than just changing the default without any notification, and eventually leads us to the same place.

Anyway I'll leave this open and await any feedback on the above, since otherwise this can be merged as is.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -107,7 +107,7 @@ class nginx (
   $package_name                   = $::nginx::params::package_name,
   $package_source                 = 'nginx',
   $package_flavor                 = undef,
-  $manage_repo                    = $::nginx::params::manage_repo,
+  $manage_repo                    = false,
   ### END Package Configuration ###
 
   ### START Service Configuation ###
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -18,7 +18,7 @@ class nginx::package(
   $package_source = 'nginx',
   $package_ensure = 'present',
   $package_flavor = undef,
-  $manage_repo    = $::nginx::params::manage_repo,
+  $manage_repo    = false,
 ) inherits ::nginx::params {
 
   anchor { 'nginx::package::begin': }
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -14,7 +14,7 @@
 #
 # This class file is not called directly
 class nginx::package::debian(
-    $manage_repo    = true,
+    $manage_repo    = false,
     $package_name   = 'nginx',
     $package_source = 'nginx',
     $package_ensure = 'present'
--- a/manifests/package/redhat.pp
+++ b/manifests/package/redhat.pp
@@ -14,7 +14,7 @@
 #
 # This class file is not called directly
 class nginx::package::redhat (
-  $manage_repo    = true,
+  $manage_repo    = false,
   $package_ensure = 'present',
   $package_name   = 'nginx',
   $package_source = 'nginx-stable',
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -9,7 +9,6 @@ class nginx::params {
     'log_dir'     => '/var/log/nginx',
     'run_dir'     => '/var/nginx',
     'package_name' => 'nginx',
-    'manage_repo'  => false,
   }
   case $::osfamily {
     'ArchLinux': {
@@ -19,16 +18,8 @@ class nginx::params {
       }
     }
     'Debian': {
-      if ($::operatingsystem == 'ubuntu' and $::lsbdistcodename in ['lucid', 'precise', 'trusty'])
-      or ($::operatingsystem == 'debian' and $::operatingsystemmajrelease in ['6', '7', '8']) {
-        $_module_os_overrides = {
-          'manage_repo' => true,
-          'daemon_user' => 'www-data',
-        }
-      } else {
-        $_module_os_overrides = {
-          'daemon_user' => 'www-data',
-        }
+      $_module_os_overrides = {
+        'daemon_user' => 'www-data',
       }
     }
     'FreeBSD': {
@@ -41,15 +32,6 @@ class nginx::params {
     'Gentoo': {
       $_module_os_overrides = {
         'package_name' => 'www-servers/nginx',
-      }
-    }
-    'RedHat': {
-      if ($::operatingsystem in ['RedHat', 'CentOS'] and $::operatingsystemmajrelease in ['5', '6', '7']) {
-        $_module_os_overrides = {
-          'manage_repo' => true,
-        }
-      } else {
-        $_module_os_overrides = {}
       }
     }
     'Solaris': {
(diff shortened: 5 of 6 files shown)


bastelfreak:
so this changes the default behavior, right? Can you add a section to the CHANGELOG.md that mentions it? So we don't forget to do a proper major verion bump for the next release.

wyardley:
@bastelfreak: looks like changelog creation is fully automated? I don't know enough about the release process to know if there's something / somewhere we can add notes about breaking changes to.

3flex:
As is, yes, this changes default behaviour.

Any feedback on my alternate approach, described above? That should have no effect on existing installs, other than throwing a new warning. I'm ambivalent about which way you want to go, just something to consider.

For the changelog - add something now, or will it be auto-generated later?

wyardley:
One other question... on CentOS / RHEL, there is no 'nginx' package unless EPEL is enabled, so if we're moving towards vendor packages, do we need some kind of docs change or soft dependency?

The acceptance tests already include EPEL, which I added fairly recently, more to do with something Passenger related:
```
      elsif fact('osfamily') == 'RedHat'
        # Soft dep on epel for Passenger
        install_package(host, 'epel-release')
      end
```
However, my guess is that with default params, these tests would fail if there was no EPEL.

I don't really love the approach of requiring an 'epel' module as a require or forced dependency, as users may have their own way of including EPEL locally if they use it that wouldn't create a class named 'epel', and whether or not to put it in metadata.json when it's only needed on RedHat type systems is also a question, but I also don't like the idea of enabling EPEL within this module either, since it could cause unpredictability for end-users.

wyardley:
I agree about getting rid of $manage_repo in favor of just using $package_source. It would be nice if there was a way for the end user to pass in a version that gets passed on to the provider, so that the package version can be ensured (to latest, or a specific version), rather than just "present", at the user's request.

bastelfreak:
@3flex, we fixed the failing ruby231 test on master, they will be gone if you rebase.

3flex:
I'm not sure what the best thing is for CentOS & Red Hat. I don't have experience with them, so hadn't considered the lack of vendor packages... though I see that it's apparently in the Red Hat Software Collections as of Red Hat 7? Could that be used? Is there an equivalent package source for CentOS? Even if there is, it doesn't help with Red Hat 5 or 6...

I'll make the change to push people towards using package_source to specify alternate sources, and can add a way to select the version.

Will need someone to provide a solution for Red Hat systems though.

wyardley:
EPEL is a semi-official repo; using it is very common among RedHat / CentOS users, and it does provide an Nginx package that's decent, and has a fairly recent version.

The issue, though, is that due to the way Puppet handles packages, we can't (easily) "just" install Nginx and its modules from EPEL, so the repo would need to be enabled for this to work. I still think some kind of requirement for EPEL with the default package source is reasonable, it's just whether it should be a "hard" dependency somehow, or a soft one. There are modules like https://forge.puppet.com/stahnma/epel, but some people prefer to handle configuring it differently.

Just listing it in the docs is how puppetlabs-rabbitmq handles it, FWIW (which is what I'm calling a 'soft' dep; you'll notice it's not listed in metadata.json):
https://github.com/puppetlabs/puppetlabs-rabbitmq#module-dependencies

wyardley:
@bastelfreak and I chatted about this, and think it makes sense to not try to change the code for a soft dep, but just mention the EPEL dependency in the docs, and also mention the workaround of using a different repo source instead of default.

@3flex: are you going to rework this to get rid of $manage_repo entirely?

SCL is an option, but I don't prefer it personally (as the default).

wyardley:
@3flex: is any more discussion needed?
Can you remove the $manage_repos option if we think that's what's best, and add a docs bit about the EPEL requirement?

Something like this, maybe, if we want it in README rather than quickstart?

"On RHEL and CentOS, when `$package_source` is not set, Puppet expects to have EPEL available to get the nginx package. This can be managed by the stahnma/epel module, if you like, or you can handle the dependency at the site level. Alternately, you can choose one of the other available package sources as described in quickstart.md`

quickstart.md already has a mention of EPEL, though @bastelfreak says that the term I used "soft dependency" is the wrong one here. In any event, the section in quickstart.md needs to be updated to reflect these changes for sure.

3flex:
Life's getting in the way of making these updates - if someone wants to open a new PR (or make changes to this branch directly) feel free. I know @wyardley has been making a tonne of great changes in short order so feel free to run with this.

I'll come back to it eventually but I don't want it to be a blocker for any new Forge release.

wyardley:
No worries, we can just hold off on this change and skip it in this release. If I have time during the contributor summit, might take a pass at reworking it, but going to leave it here for now.

wyardley:
Replacing with #938, leaving @3flex's commit attributed to him.
Still working on some of the structural changes in that one.

Issue #916, opened 2016-10-11
remove $configtest_enable parameter, look into nginx::service in general


wyardley:
When looking into #722, it looks like 
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/service.pp#L63-L66
and related parameters are totally broken (though strangely, I'm having a hard time actually . This code should be looked at and simplified, and the $configtest_enable param should probably be taken out (many systems already do this in their init scripts anyway).

I really don't see how the existing `nginx::service` is working at all, need to look at it more closely.

tux-o-matic:
@wyardley part of the problem is that the module today is depending on an init/systemd script which is not managed by the module.
Since Nginx only recently added some sort of dynamic module loading, compiling Nginx isn't that unusual and there for setting the service script is left up to you.

So in service.pp, it's optional because some script might not have reload (not a standard initd function like restart) and some use cases might want to use restart rather than reload, each using a specific kill signal. But the module design today makes some the assumptions about restart.

PR #921, opened 2016-10-11
Remove broken configtest_enable option

wyardley:
Per #916, the `$configtest_enable` option doesn't seem if it could possibly work. In general, it would be good to review if everything here does what it is supposed to, but I think this will do the right thing on most platforms without any additional work.

I'm not sure if anyone would be using the `$service_restart` command directly, I'd think in most cases, we should be relying on `$service_name` and letting Puppet do the right thing, though I'm open to suggestion if people think `$service_restart` still needs to be exposed directly. What I don't understand is that, though it defaults to `/etc/init.d/nginx reload`, which doesn't work, the systems I tested all were using systemctl to restart the service anyway.

We still may want to revisit some kind of reporting back if the config test fails in a more generic way. I'm not going to address that one way or another for now.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -111,10 +111,8 @@ class nginx (
   ### END Package Configuration ###
 
   ### START Service Configuation ###
-  $configtest_enable              = false,
   $service_ensure                 = running,
   $service_flags                  = undef,
-  $service_restart                = '/etc/init.d/nginx reload',
   $service_name                   = undef,
   $service_manage                 = true,
   ### END Service Configuration ###
@@ -302,12 +300,10 @@ class nginx (
   Class['::nginx::package'] -> Class['::nginx::config'] ~> Class['::nginx::service']
 
   class { '::nginx::service':
-    configtest_enable => $configtest_enable,
-    service_ensure    => $service_ensure,
-    service_restart   => $service_restart,
-    service_name      => $service_name,
-    service_flags     => $service_flags,
-    service_manage    => $service_manage,
+    service_ensure => $service_ensure,
+    service_name   => $service_name,
+    service_flags  => $service_flags,
+    service_manage => $service_manage,
   }
 
   create_resources('nginx::resource::upstream', $nginx_upstreams)
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -14,8 +14,6 @@
 #
 # This class file is not called directly
 class nginx::service(
-  $configtest_enable = false,
-  $service_restart   = '/etc/init.d/nginx reload',
   $service_ensure    = 'running',
   $service_name      = 'nginx',
   $service_flags     = undef,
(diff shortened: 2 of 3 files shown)


wyardley:
passed acceptance tests for
centos-66-x64
centos-72-x64
ubuntu-server-1404-x64
debian-78-x64

Getting a (presumably unrelated) error running on debian-82-x64
```/usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/host.rb:351:in `exec': Host 'debian-82-x64' exited with 1 running: (Beaker::Host::CommandFailure)
 puppet agent --configprint hiera_config
Last 10 lines of output were:
		from /usr/lib/ruby/vendor_ruby/puppet/util.rb:14:in `<top (required)>'
		from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
		from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
		from /usr/lib/ruby/vendor_ruby/puppet.rb:8:in `<top (required)>'
		from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
		from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
		from /usr/lib/ruby/vendor_ruby/puppet/util/command_line.rb:12:in `<top (required)>'
		from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
		from /usr/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
		from /usr/bin/puppet:7:in `<main>'
	from /usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/host.rb:39:in `[]'
	from /usr/local/lib/ruby/gems/2.3.0/gems/beaker-2.51.0/lib/beaker/dsl/install_utils/foss_utils.rb:300:in `block in install_puppet_on'
```

tux-o-matic:
@wyardley you removed a feature by deleting 'service_restart'. 
One example, Docker, when running Nginx within a container rather than a initd/systemd service, you can use 'service_restart' to have the Puppet module interact with the container.

wyardley:
@tux-o-matic: I can re-implement `$service_restart` separately (though I'd argue that it's generally a Good Thing to just use the `$service_name` parameter most of the time, and let puppet handle the service itself).

In my earlier test, even with `$configtest_enable` set, I couldn't get Puppet to call the specified custom init script, however, I am able to get that to work now. It is easy enough to restore this functionality without `$configtest_enable`; I'll open another PR for that (#927)

tux-o-matic:
Thanks @wyardley. For Nginx users wishing to mix Puppet and Docker, you can't just rely on the Puppet Service type (until we get a Docker provider for Service).

PR #922, opened 2016-10-11
Support for proxy_pass_header directive.

gallagherrchris:
It is better to explicitly define this than using raw_append or raw_prepend.

From nginx docs:
Permits passing otherwise disabled header fields from a proxied server to a client.
By default, nginx does not pass the header fields "Date", "Server", "X-Pad", and "X-Accel-..." from the response of a proxied server to a client.

### DIFF
--- a/spec/defines/resource_location_spec.rb
+++ b/spec/defines/resource_location_spec.rb
@@ -688,6 +688,15 @@ describe 'nginx::resource::location' do
           ]
         },
         {
+          title: 'should pass proxy headers',
+          attr: 'proxy_pass_header',
+          value: ['X-TestHeader1 value1', 'X-TestHeader2 value2'],
+          match: [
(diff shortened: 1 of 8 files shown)


wyardley:
I've done it with `$location_cfg_append` and it worked OK, but this looks great to me, especially since it's in vhost and location contexts.

spec/defines/resource_location_spec.rb
```diff
@@ -688,6 +688,15 @@
           ]
         },
         {
+          title: 'should pass proxy headers',
+          attr: 'proxy_pass_header',
+          value: ['X-testHeader1 value1', 'X-TestHeader2 value2'],
```
gallagherrchris: Ah nice catch.  I will update that

wyardley:
This looks reasonable to me... the conditionals already are inconsistent, so don't think the style thing is a dealbreaker.

Issue #741, opened 2016-01-20
Cannot deny access via location


Hufschmidt:
The following location (Deny access to .htaccess/.htpasswd files on nginx reverse proxy) does not work:

    location ~ /\.ht(access|passwd)$ {
      deny all;
    }

This location gets rejected by: **manifests/resource/location.pp:325**

    if (($www_root == undef) and ($proxy == undef) and ($location_alias == undef) and ($stub_status == undef) and ($fastcgi == undef) and ($uwsgi == undef) and ($location_custom_cfg == undef) and ($internal == false)) {
        fail('Cannot create a location reference without a www_root, proxy, location_alias, fastcgi, uwsgi, stub_status, internal, or location_custom_cfg defined')
      }

Website gets deployed and kept in sync on proxies and nginx loadbalancer via git, so removing htaccess on the loadbalancer is not a (good) solution.

zaroth:
I was under the impression that that's what `location_custom_cfg` is for? The following works for me:

    nginx::resource::location {'htaccess block':
      location => '~ /\.ht(access|passwd)$',
      location_custom_cfg => { 'deny' => 'all' },
      vhost => 'myhost',
    }

Hufschmidt:
well there a location_deny and location_allow setting on nginx::resource::location, so it feals kind of counter-intuitive having to resort back to using location_custom_cfg. (Which I'm currently using, since there is no other way)

This warning should either be extended to check location_deny, location_allow or removed all together and let nginx complain about invalid configurations...

wyardley:
You're right, doesn't seem to let you create a location resource without one of those things.

```
nginx::nginx_vhosts:
  'test_80':
    www_root: '/var/www'
    ensure: present
    listen_port: 80
    server_name: 
      - localhost
    listen_options: 'default_server'
    use_default_location: false

nginx::nginx_locations:
  '~ /\.ht(access|passwd)$':
    location_deny:
      - 'all'
    location_allow:
      - '127.0.0.1'
    vhost: 'test_80'
```
I'm not sure what the ramifications would be of totally dropping the check in line 352 of `manifests/resource/location.pp`, though does seem like the workaround of putting something in location_custom_cfg, even if it's not the allow / deny conditions.

wyardley:
Would adding `$location_deny` /  `$location_allow` to that long list of parameters be a reasonable solution? or are there too many other things that could also be in a location block?

[...]

PR #923, opened 2016-10-12
Allow location_allow / location_deny as well in location blocks

wyardley:
See #741 
I did not add any additional tests here, if anyone has a suggestion of a test that should be added, I can add one.
This also simplifies the syntax a little, will need to verify that the changes work as I expect with all Puppet versions.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -357,10 +357,10 @@ define nginx::resource::location (
   if ($vhost == undef) {
     fail('Cannot create a location reference without attaching to a virtual host')
   }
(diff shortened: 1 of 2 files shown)


wyardley:
Also add $try_files to allowed directives... maybe we should just consider dropping the sanity check here. h/t to @vladpanainte

Will try and figure out if there is any more testing I can do for this beyond https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_location_spec.rb#L823

jyaworski:
Please rebase.

wyardley:
@jyaworski: this was current but had conflicts due to some of my other merged commits. Rebased.

PR #924, opened 2016-10-12
Add expires directive to location

wyardley:
This is an update of #746, since @snowmean didn't have time to resolve conflict / add tests.


wyardley:
Leaving this in 2 commits to preserve original author's credit.

PR #926, opened 2016-10-13
uwsgi: allow custom uwsgi_param directives

darken99:
This patch allows custom uwsgi params to be added with uwsgi_param, same way fastcgi_param work.

@wyardley, It is a replacement for #767

### DIFF
--- a/spec/defines/resource_location_spec.rb
+++ b/spec/defines/resource_location_spec.rb
@@ -602,6 +602,31 @@ describe 'nginx::resource::location' do
           end
         end
       end
+
+      context "when uwsgi_param is {'CUSTOM_PARAM' => 'value'}" do
+        let(:params) { default_params.merge(uwsgi_param: { 'CUSTOM_PARAM' => 'value', 'CUSTOM_PARAM2' => 'value2' }) }
+        it 'sets uwsgi_param' do
+          should contain_concat__fragment('vhost1-500-' + Digest::MD5.hexdigest(params[:location].to_s)).
+            with_content(%r{uwsgi_param\s+CUSTOM_PARAM\s+value;}).
+            with_content(%r{uwsgi_param\s+CUSTOM_PARAM2\s+value2;})
+        end
+      end
+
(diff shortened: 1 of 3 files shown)


spec/defines/resource_location_spec.rb
```diff
@@ -596,6 +596,36 @@
...
+        end
+      end
+
+      context 'when uwsgi_param is not set' do
+        let(:params) { default_params }
+        it 'does not set uwsgi_param' do
+          should contain_concat__fragment('vhost1-500-' + Digest::MD5.hexdigest(params[:location].to_s)).
+            without_content(%r{uwsgi_param\s+CUSTOM_PARAM\s+.+?;}).
+            without_content(%r{uwsgi_param\s+CUSTOM_PARAM2\s+.+?;})
+        end
```
wyardley: This all looks pretty good to me. Thanks!
Assuming there shouldn't be *any* uwsgi_param lines in the file I'm wondering if maybe this should just be simplified to a single `without_content(%r{uwsgi_param})`?

darken99: I'm sorry, but i didn't got what file you mean and what exact section?
If you mean `when uwsgi_param is not set` should lead to empty params - then probably it's also needed to change fastcgi section where i copied it from?

wyardley: Either is technically a test that will pass, but, assuming there shouldn't be *any* uwsgi_param directives in the test case (which I _think_ is the case), the simpler test is also a better one.

Yeah, probably same thing with the example you copied from... this module has a ton of sub-optimal test. I'm Ok with merging this either way, but if you don't mind, can you test whether the simpler case passes tests?

darken99: Checks fail with this change. Or maybe i changed it in a wrong way? Can you take a look?

darken99:
@wyardley You attached wrong patch file :) But i got the idea from your comments and it seems now everything is fine.

wyardley:
@darken99: d'oh.... glad you got the idea, though. Merged.

PR #927, opened 2016-10-13
Restore $service_restart, now defaulting to undefined, but now withou…

wyardley:
Now defaulting to undef, but without checking whether now removed $configtest_enable param exists. This allows users to set a custom restart command if just changing the service name isn't enough. See comments in #921 

I'm still wondering if we should invoke nginx::service differently from how it is now:
          https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/init.pp#L305 
and just set `$nginx::service::foo` to `$nginx::foo`; seems to me this would be a bit easier to parse.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -114,6 +114,7 @@ class nginx (
   ### START Service Configuation ###
   $service_ensure                 = running,
   $service_flags                  = undef,
+  $service_restart                = undef,
   $service_name                   = undef,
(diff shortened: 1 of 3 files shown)


wyardley:
per @hunner, seems like it's a better practice to just pull the necessary params in the way I'm changing it to, since `nginx::service` is a private class.
I'll make a different one for `nginx::package` if that looks good.

PR #930, opened 2016-10-17
Use default ssl_protocols for ssl mailhosts

ekohl:
Based on https://github.com/voxpupuli/puppet-nginx/pull/909. It also adds a test for the $ssl_ciphers parameter.

I haven't had the time yet to verify this on a real system.


manifests/resource/mailhost.pp
```diff
@@ -101,6 +103,7 @@
   if ($ssl_cert != undef) {
     validate_string($ssl_cert)
   }
+  validate_string($ssl_protocols)
```
jyaworski: I feel blah about having this as a string rather than an array, but I guess there's a precedent.

templates/mailhost/mailhost_ssl.erb
```diff
@@ -30,7 +30,7 @@ server {
   ssl_certificate_key  <%= @ssl_key %>;
   ssl_session_timeout  5m;
 
-  ssl_protocols TLSv1;
+  ssl_protocols <%= @ssl_protocols %>;
```
dol: The same change applies to the  puppet-nginx/templates/mailhost/mailhost.erb file. Because the SSL part of the two templates are equal I used to combine the two into one file that gets included. I'll fix the missing variable in my PR #769.

PR #937, opened 2016-10-19
Update based on voxpupuli/modulesync_config 0.13.4

dhoppe:
Please merge.


wyardley:
Seems like the nginx_version fact is barfing again with this update.

PR #938, opened 2016-10-19
[WIP] remove manage_repo and rework repo management defaults

wyardley:
Update of #914 
@3flex

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -13,15 +13,14 @@
 # Sample Usage:
 #
 # This class file is not called directly
-class nginx::package(
-  $package_name   = $::nginx::params::package_name,
-  $package_source = 'nginx',
-  $package_ensure = 'present',
-  $package_flavor = undef,
-  $manage_repo    = $::nginx::params::manage_repo,
-) {
+class nginx::package { 
 
   assert_private()
+
+  $package_name   = $::nginx::package_name,
+  $package_source = $::nginx::package_source,
+  $package_ensure = $::nginx::package_ensure,
+  $package_flavor = $::nginx::package_flavor,
 
   anchor { 'nginx::package::begin': }
   anchor { 'nginx::package::end': }
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -13,14 +13,14 @@
 # Sample Usage:
 #
 # This class file is not called directly
-class nginx::package { 
+class nginx::package {
 
   assert_private()
 
-  $package_name   = $::nginx::package_name,
-  $package_source = $::nginx::package_source,
-  $package_ensure = $::nginx::package_ensure,
-  $package_flavor = $::nginx::package_flavor,
+  $package_name   = $::nginx::package_name
+  $package_source = $::nginx::package_source
+  $package_ensure = $::nginx::package_ensure
+  $package_flavor = $::nginx::package_flavor
 
   anchor { 'nginx::package::begin': }
   anchor { 'nginx::package::end': }
@@ -28,7 +28,6 @@ class nginx::package {
   case $::osfamily {
     'redhat': {
       class { '::nginx::package::redhat':
-        manage_repo    => $manage_repo,
         package_source => $package_source,
         package_ensure => $package_ensure,
         package_name   => $package_name,
@@ -41,7 +40,6 @@ class nginx::package {
         package_name   => $package_name,
         package_source => $package_source,
         package_ensure => $package_ensure,
-        manage_repo    => $manage_repo,
         require        => Anchor['nginx::package::begin'],
         before         => Anchor['nginx::package::end'],
       }
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -11,11 +11,10 @@ This will install the NGINX package from the software repository of your Linux d
 
 ```
 class{'nginx':
-    manage_repo => true,
     package_source => 'nginx-mainline'
 }
 ```
-The choices here are `nginx-stable` (the current 'production' level release), `nginx-mainline` (where active development is occuring), as well as `passenger` - you can read a full explanation of the differences [here][nginxpackages]. `passenger` will install Phusion Passenger, as well as their version of nginx built with Passenger support. Keep in mind that changing `package_source` may require some manual intervention if you change this setting after initial configuration. On CentOS / RHEL, there is a soft dependency on EPEL for this (i.e., the module doesn't configure EPEL for you, but will fail if you don't have it).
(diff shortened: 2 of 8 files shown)


wyardley:
Still having trouble figuring out how to resolve this warning with the new structure:
```
  1) nginx with defaults should contain Class[nginx::service] that subscribes to Class[nginx::package]
     Failure/Error: it { is_expected.to contain_class('nginx::service').that_subscribes_to('Class[nginx::package]') }
       expected that the catalogue would contain Class[nginx::service] with that is subscribed to Class[nginx::package]
     # ./spec/classes/nginx_spec.rb:28:in `block (3 levels) in <top (required)>'
```
Bryan mention that he usually uses the other pattern (`class { 'foo': bar => $bar }`) in a case like this; I can just switch that back if folks think that's better.

wyardley:
Ubuntu 14 is also failing on this: https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/package/debian.pp#L31
```
       	Info: Loading facts
       	Error: Could not find class ::apt for ubuntu-server-1404-x64.veggiechinese.net on node ubuntu-server-1404-x64.example.net
       	Error: Could not find class ::apt for ubuntu-server-1404-x64.veggiechinese.net on node ubuntu-server-1404-x64.example.net
```

wyardley:
Acceptance tests pass for Ubuntu 14 and CentOS 7

wyardley:
Updated to reflect some feedback from @danzilio 
Couldn't get it to work without an `inherits ::nginx` in service and package, though

This at least should be passing spec tests now. I did not yet make a feature for package notifying service, but we can remove this (and the test) if people think that would be an improvement, or make it configurable as @danzilio suggested. This is a little messy, but keeps the existing behavior, for better or for worse.

I think there are still more structural improvements that could be made (esp. in terms of whether we really need a separate `package/{apt,debian}.pp`, but trying to keep that out of scope.

wyardley:
@danzilio: Updated to include the `pre_condition` in the package and service class spec tests, however, still have some problems there, specifically, e.g.,
```
Duplicate declaration: Class[Nginx::Package] is already declared; cannot redeclare at line 2 at line 2:1 on node xyz
```
Is that what you were thinking, or does something need to be adjusted.

danzilio:
OK! This makes a lot more sense seeing the code. So, what's happening here is that in the `pre_condition` block we're including `nginx` but then rspec-puppet is automatically including `nginx::package` which is resulting in a duplicate definition error. This is because our `pre_condition` has already added the class to the catalog using `include` and we're trying to define it again with the rspec test.

Since we've made `nginx::package` a private class, the tests for this class could actually move to `spec/classes/nginx_spec.rb` since the `nginx` class is our interface. There's no reason to test our private classes separately, since we're more interested in validating the interface.

Now, that's just one way to handle it. These test failures are a sign of tight coupling between the `nginx` and `nginx::package` classes. Another way to handle this would be to create a `manage_package` parameter on the `nginx` class and use that parameter to conditionally include the `nginx::package` class. Then we could keep the tests in a separate file and change our `pre_condition` block to define the `nginx` class with `manage_package` set to `false`. This would avoid the duplicate resource definition error and allow us to test the `nginx::package` class separately.

Making this code change would do two things: it would create a looser coupling between the two classes and also make this module more reusable, especially for people who want to manage the package elsewhere in their code. This is not only designing for testability, it's enabling other use cases and making our code more modular. This would be my recommendation.

danzilio:
One other option is to just define all of the parameters necessary to test `nginx::package` in the `rspec` test and remove the `pre_condition` altogether. If you decide to go this route, I would strongly suggest adding some tests to `nginx_spec.rb` to validate the interface between `nginx` and `nginx::package` as well. You can see an example of this in https://github.com/danzilio/puppet-letsencrypt/blob/master/spec/classes/letsencrypt_install_spec.rb

wyardley:
@danzilio: 
I was hoping to get _rid_ of knobs and whistles here, not add more 😉.
Would just setting `package_ensure` to false or undef accomplish the same thing (I guess if they managed it elsewhere, they'd still get a duplicate resource warning).

I'll think about this and work on implementing your suggestions later today or tomorrow; thanks for the detailed feedback.

wyardley:
@danzilio:
So, even if I add a `$manage_package` parameter and call the class without it, the `assert_private()` seems to cause problems with testing the `nginx::package` class in its own spec test, I get
```
     Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, Class nginx::package is private at [...]/puppet-nginx/spec/fixtures/modules/nginx/manifests/package.pp:23:3 on node xyz
```
So maybe better to move all these tests into the main class spec, even though it seems less confusing to have them mapped directly? If I drop the `assert_private()`, the tests pass fine (tho leaving it in in the commit I'm going to push).

Did some work on service as well, for this one, I didn't want to make the whole class managed by a flag, so I just stubbed in all the expected values (some were there already, so this wasn't a huge change, and I think makes this at least better than it was).

Those tests pass except these two...

```
  1) nginx::service using default parameters should contain Service[nginx] with restart undefined
     Failure/Error: it { is_expected.to contain_service('nginx').without_restart }
       expected that the catalogue would contain Service[nginx] with restart undefined
     # ./spec/classes/service_spec.rb:24:in `block (3 levels) in <top (required)>'

  2) nginx::service when service_manage => false should not contain Service[nginx]
     Failure/Error: it { is_expected.not_to contain_service('nginx') }
     
     Puppet::Error:
       Could not find resource(s) Service[nginx] for overriding on node xyz
     # ./spec/classes/service_spec.rb:53:in `block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Puppet::ParseError:
     #   Could not find resource(s) Service[nginx] for overriding
     #   ./spec/classes/service_spec.rb:53:in `block (3 levels) in <top (required)>'
```

manifests/package.pp
```diff
@@ -14,20 +14,20 @@
 #
 # This class file is not called directly
 class nginx::package(
-  $package_name   = $::nginx::params::package_name,
-  $package_source = 'nginx',
-  $package_ensure = 'present',
-  $package_flavor = undef,
-  $manage_repo    = $::nginx::params::manage_repo,
-) inherits ::nginx::params {
+  $package_name   = $::nginx::package_name,
```
wyardley: Should this be directly from params, or as it is here?

wyardley:
So, I've reworked the way the classes are structured to match the convention of what we did with `nginx::config` a bit more, and moved back away from `include ::foo` again when invoking the classes.

I still need to re-do the tests, and change package_source to an enum.

While we will have an unmanaged package type, I think we're going to default to the nginx stable package source after all.

wyardley:
I'm going to kill this one since it's now way out of date.
I do think there's some more work to be done in this area, and I noticed recently that newer EPEL versions were causing some changes in the default behavior in the acceptance tests.
That said, I'm probably not going to have time to do any more reworking of the package behavior, and I'm guessing any major problems will get looked at / handled in a more incremental way.

I have done some work in our fork of this module to allow pinning passenger version; I can submit this separately if it would be helpful.

PR #939, opened 2016-10-20
Use SSL for nginx APT repository

saz:
The nginx APT repository is currently not using https.
This PR will fix this.

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -34,14 +34,14 @@ class nginx::package::debian(
     case $package_source {
       'nginx', 'nginx-stable': {
         apt::source { 'nginx':
-          location => "http://nginx.org/packages/${distro}",
+          location => "https://nginx.org/packages/${distro}",
           repos    => 'nginx',
           key      => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62',
         }
       }
       'nginx-mainline': {
         apt::source { 'nginx':
-          location => "http://nginx.org/packages/mainline/${distro}",
+          location => "https://nginx.org/packages/mainline/${distro}",
           repos    => 'nginx',
           key      => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62',
         }
--- a/spec/classes/nginx_spec.rb
+++ b/spec/classes/nginx_spec.rb
@@ -149,7 +149,7 @@ describe 'nginx' do
         it { is_expected.not_to contain_package('passenger') }
         it do
           is_expected.to contain_apt__source('nginx').with(
-            'location'   => "http://nginx.org/packages/#{operatingsystem.downcase}",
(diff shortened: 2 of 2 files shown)


wyardley:
@saz: at a minimum, the spec tests will probably need to be updated for us to merge this (to reflect the changed URLs). 
https://github.com/voxpupuli/puppet-nginx/blob/master/.github/CONTRIBUTING.md
has some info on how to run tests, etc. (any additional commits should be squashed down to 1).

wyardley:
Technically, it just needs the tests to be updated, but I'm adding the 'needs tests' flag until the tests are passing.

xaque208:
I believe this will require the apt-transport-tls package to function correctly, which is not installed by default.

wyardley:
@xaque208 Ah, didn't realize that, but that may be another reason not to do this.

wyardley:
Assuming what @xaque208 said is correct, I think it makes sense to either a) not implement this, or b) make it a configurable option, defaulting to off. If you want to make a pass at implementing b), I think this can be discussed, but there will need to be test cases for both.

Maybe a boolean option in `nginx` (not `nginx::config`), called something like `apt_use_ssl`, defaulting to `false`?

wyardley:
@xaque208: looking again, we're already using https for passenger repo, and there's already a:
`        ensure_packages([ 'apt-transport-https', 'ca-certificates' ])` in `package/debian.pp`, so this is already taken care of, I think, if the OP just changes the spec tests.

saz:
I'll change the pr asap. Sorry for missing this! 


Sent using Boxer

On 28 Oct 2016 8:29 PM, Will Yardley <notifications@github.com> wrote:

@xaque208: looking again, we're already using https for passenger repo, and there's already a:
ensure_packages([ 'apt-transport-https', 'ca-certificates' ]) in package/debian.pp, so this is already taken care of, I think, if the OP just changes the spec tests.

xaque208:
Does `ensure_packages()` just add the resources to the catalog if they are not already present? @wyardley

xaque208:
@wyardley Never mind, read the code and answered my own question.

wyardley:
@xaque208 Yeah, just noticed this offhand when I was testing some unrelated acceptance test stuff related to Ubuntu 16; the Passenger repo is already using https (and works in the acceptance tests), so I think this will work as-is. Still need to work out whether it's worth embedding the GPG key in the package *and* whether it's feasible / practical to deal with GPG key expiration directly in the module.

@saz: thanks! I mentioned it above, but make sure to squash the commits.

There are a couple PRs that are pending that have a bit of overlap, but I think (hope) they won't conflict.

wyardley:
@saz: have you had a chance to update the tests?

saz:
@wyardley should be working now.

wyardley:
👍

PR #941, opened 2016-10-25
[WIP] (do not merge) start adding docs via puppet-strings

wyardley:
This branch is where I'm going to try to work on converting docs (for the currently undocumented bits, to start with) to puppet-strings.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,30 +1,37 @@
-# Class: nginx
+# @param package_ensure [String] The 'ensure' value for the nginx package
+# @param package_name [String] The package name (defaults to nginx on most platforms)
+# @param package_source [String] The source for the package. Possible values
+#   are undef (default), meaning the package will be installed from the
+#   vendor's packaging system using configured repos. On supported platforms,
+#   this can also be set to: 'nginx-stable' (current production release from
+#   official repository), 'nginx-mainline' (current development release from
+#   official repository), or 'passenger' (Phusion Passenger repos)
+# @param package_flavor [String] Allows setting "flavor" in OpenBSD package provider
+# @param service_ensure [String] The 'ensure' value for the Service type (stopped / running).
+# @param service_flags [String] Allows setting additional flags on OpenBSD
+# @param service_restart
+# @param service_name
+# @param service_manage
+# @param geo_mappings
+# @param string_mappings
+# @param nginx_locations
+# @param nginx_mailhosts
+# @param nginx_streamhosts
+# @param nginx_upstreams
+# @param nginx_vhosts
+# @param nginx_vhosts_defaults
 #
-# This module manages NGINX.
+# @example Install and bootstrap and Nginx instance
+#     include nginx
 #
-# Parameters:
+# @example Nginx with pre-compiled Passenger (supported platforms only)
+#    class { 'nginx':
+#      package_source  => 'passenger',
+#      http_cfg_append => {
+#        'passenger_root' => '/usr/lib/ruby/vendor_ruby/phusion_passenger/locations.ini',
+#      }
+#    }
 #
-# Actions:
-#
-# Requires:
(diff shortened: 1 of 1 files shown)


wyardley:
It seems like my idea of being able to generate README from the code may not be feasible. So not sure where we should keep the examples.

The current auto-generation of docs generates HTML that's fairly useful, but has some limitations in terms of using it for a module like this, so want to get some feedback and figure out a good workflow to make sure this is practical before doing all the work.

If you wrap lines, you need to indent to continue.

Feel free to submit PRs to this branch on my fork if anyone has some time to help out.

wyardley:
ps - I'm not totally clear on whether items that are booleans or have limited choices should be 'params' or 'options'.

Also, not sure of what you'd use to represent a parameter that can be multiple types, or a parameter that's a deeper structure (say a hash of arrays of strings)
Puppet-strings has things like: `[Array<String>]` in its example, which I'd take to be a parameter that's an array of strings; this module has a lot of examples where a parameter may be a string, hash, _or_ array.

whopper:
Woot! As per our discussion on Slack, there doesn't seem to be a current standard or best practice in YARD itself for variable types. I'm going to talk to some internal Puppet module devs to see if we can come up with something that looks good. The README idea is now a feature request on Jira!

3flex:
> So not sure where we should keep the examples

There's an `examples` directory... that's where the old smoke tests were moved to comply with newer puppetlabs_spec_helper version which says that's the new convention.

So the examples from the README could be dumped in there? Since the examples are mostly undocumented in the README I don't see an issue with moving them as-is into that directory.

HAIL9000:
Disclaimer that I haven't worked on strings in a while, but I believe the way you do multiple variable types is like this:

```
@param name [Array, Hash] description of parameter
```

This is not currently in our documentation but probably should be.

And strings also supports the YARD `@example` tag if that's helpful. We should definitely provide some examples of how to use that for puppet code specifically, but you can find info about in the [YARD docs](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#example).

wyardley:
@HAIL9000 Thanks! That's very helpful.

re examples, I have one (very simple) `@example` in the current PR, and the idea would be to convert the existing examples as well.  I'm not sure *where* they should go, but I think the project of moving variables out of the ::config namespace and back into the main module namespace probably will help with this. My concern is that the examples remain somewhere useful once they're rendered, which may be hard if they're all inline near the section of the code that deals with them.

wyardley:
https://puppetcommunity.slack.com/archives/puppet/p1477419377006057 from @whopperpdx "looks like you’d use it when your parameter is a hash. Each @option would describe a key...."... since the class itself is defined as a hash, not sure if that means each named option would be an `@option` or a `@param` when writing stuff in the DSL

wyardley:
btw, I've left this open mostly for reference, but fine to close it out if folks want.
I do think that, especially now that we've started moving stuff to Puppet 4 and done all the other work to populate data types, it would be really nice if we could take a stab at adding in puppet-strings. If someone wants to rough in how the format could be, I can try to make the time to do some copy / pasta work, or we could ask on Slack / IRC and see if there are still any volunteers to take this task on....

Do we know if there is a reference implementation for modules specifically now post-1.0 puppet-strings?

wyardley:
Seems like, as usual, https://github.com/puppetlabs/puppetlabs-ntp is probably the reference?

HAIL9000:
Hey @wyardley sorry for the lack of activity on this! I think this is valuable so I'm gonna put it on my todo list to take a look at where this is at and hopefully I can get back to you this week!

wyardley:
We got some more clarification in 
https://github.com/voxpupuli/puppet-rabbitmq/pull/562
which is now merged. I got some other suggestions as well for existing examples in modules, including puppet-autofs.

I'm going to close this one now, since we'll probably want to start over from scratch if we implement strings for this module. There are a ton of missing docs if we can find anyone who would like to help with this project.

PR #943, opened 2016-10-25
version bump and changelog for 0.5.0

wyardley:
It's been way too long, time to cut a 0.5.0, though we should follow with another release soonish.


CHANGELOG.md
```diff
@@ -1,4 +1,124 @@
-# Change Log
+# Changelog
+
+## 2016-10-25 Release [0.5.0](https://github.com/voxpupuli/puppet-nginx/tree/0.5.0)
```
bastelfreak: we're parsing this line with a regex. I'm not sure if it matches when the number is surrounded by braces.

wyardley: Based on the other examples, looks like a 'v' is expected before the version number. rebased and force-pushed.

PR #944, opened 2016-10-26
add before => Package['nginx'] on repo absent ensures

wyardley:
This fixes a minor failure in the acceptance tests, though may still have to rework slightly for #938


wyardley:
Added some additional tests for this as well (though these will also probably have to change in my other PR).

PR #945, opened 2016-10-26
Allow vhost ssl cert andn key inheritance from http section

jeffmccune:
Without this patch the SSL cert and key are always configured within vhost
server blocks.  This is a problem when using a single wildcard certificate
operating with multiple virtual host entries because the configuration results
in increased memory consumption according to the NGINX documentation:

> It is better to place a certificate file with several names and its private key
> file at the http level of configuration to inherit their single memory copy in
> all servers.

This patch addresses the problem by making it possible to pass the boolean
value `false` to the ssl_key and ssl_cert parameters.  When doing so, the SSL
key and certificate will be inherited from the http section of the
configuration.

References:

 * http://nginx.org/en/docs/http/configuring_https_servers.html#name_based_https_servers


wyardley:
This seems similar to #794 (though that one is not in working state right now).
Looks pretty sensible to me, and thanks for adding tests. I'm not sure what the deal is with the coveralls failure, but I've been getting those too, I'll ask around.

bastelfreak:
thanks @jeffmccune !

PR #948, opened 2016-10-27
default proxy_redirect to undef in locations (resolves #395)

### LINKED ISSUE #395: proxy_redirect default value
i'm not sure what is the rationale behind making default value for `proxy_redirect` `off` but it is different from default value `default` used by nginx itself when the option is not present in config file. http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect

also it looks like default value in params.pp was `undef` and option was omitted unless a non-nil value was passed - was there a reason to change that?

wyardley:
I think this is pretty simple and sane, though does change the default 'location' config when proxy is enabled.

cc:  @keymone

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -94,7 +94,7 @@ class nginx::config(
   $proxy_headers_hash_bucket_size = '64',
   $proxy_http_version             = undef,
   $proxy_read_timeout             = '90',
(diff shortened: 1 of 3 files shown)


keymone:
looks good, thanks 👍  🚢

PR #950, opened 2016-10-27
Major change: Rework namespace (get rid of ::config namespace again)

wyardley:
Not sure how this should be accomplished, but would welcome some input, here's my first pass at restructuring things.

A few questions:
1. Should params be pulled in in init, then passed to config, or should config take things that come from params directly? The problem is, we want users to be able to override most params, so I think most of them should be set in init, then passed to config.
2. Should config reference everything at top scope ($::nginx::foo)? The params used directly in config.pp are not many, but all the configs generated by templates are also in the config namespace. Having `inherits ::nginx` doesn't seem to make a difference here.

I thought templates would always take `@foo` from the top level namespace of the module, but with these changes, I'm getting some errors (see below).

I think most of these changes will break or cause unpredictable behavior when people set nginx::config::foo in hiera... thoughts on how to warn on this or keep backwards compatibility


wyardley:
Pinging @danzilio and @bastelfreak again

wyardley:
Spoke with @hunner, @bmjen and others about this in the modules meeting, and got some recommendations about how this should proceed.

wyardley:
After community modules review, went in the direction that modules team is recommending as current best practice. This is pretty big and nasty, so would appreciate some eyes to make sure I didn't massively screw something up, though I am sure there will need to be some more improvements.

This will definitely be a breaking change and a big 180; the 

I'm not quite sure how the config class was getting Debian / Ubuntu type defaults before in the 'with defaults' context; I changed tests in that scope to the actual `daemon_owner` that are the default in `params.pp`, but could especially use a look at LL 341-347 in `spec/classes/nginx_spec.rb`.

service and package still are private classes, but are not yet quite following the same convention, but want to tackle that separately if / after this gets merged.

Followed @danzilio's first suggestion about the specs (which @hunner agreed with), namely, including all the private classes in the main class spec. For now, I still left the private classes in their own contexts.

wyardley:
Also pulled out the redundant validations in `config.pp`

manifests/config.pp
```diff
@@ -247,12 +139,12 @@
       File["${conf_dir}/conf.d"] {
         purge   => true,
         recurse => true,
-        notify  => Class['::nginx::service'],
+        notify  => Class['service'],
```
bastelfreak: just curious, why did you remove the ::nginx?

wyardley: That is a  _good_ question. To be honest, I think that's the result of an overzealous sed command or search / replace!

Updated and pushed, thanks for catching that.

wyardley:
ps - I'll do a rebase too

xaque208:
This seems like a good change to make here, but why?  I'm on a version of the module (in one of my environments) where setting values on the `nginx` class throws deprecation warnings and says set things on `nginx::config`.  Are we thinking now that `nginx::config` was the wrong move?

Fwiw, this also should make it simpler to migrate that data to puppet4 native hiera if we wanted.

bastelfreak:
@xaque208 we figured out that throwing the warning doesn't meet the current best practices so we are removing it.

xaque208:
Makes sense, thanks @bastelfreak.

PR #954, opened 2016-10-28
Bypass puppet_install_helper, revert to install_puppet_(agent)_on

wyardley:
@3flex: do you remember why you switched to puppet_install_helper?
I feel like we probably want Puppet 4 by default at this point, and seems easiest to just specify the `install_puppet_agent_on` method?

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,13 +1,12 @@
 require 'beaker-rspec'
-require 'beaker/puppet_install_helper'
-
-run_puppet_install_helper
 
 RSpec.configure do |c|
   # Project root
   proj_root = File.expand_path(File.join(File.dirname(__FILE__), '..'))
 
   c.formatter = :documentation
+
+  install_puppet_agent_on(hosts, options)
 
   # This is where we 'setup' the nodes before running our tests
   c.before :suite do


wyardley:
btw, tested and CentOS 6.6 and Ubuntu 14 still seem to work with it this way.

3flex:
I think there was something that didn't seem to work without that change but I can't remember which unfortunately.

It would be great if we had these running in Travis!

wyardley:
Huh, let's keep an eye out. It worked on the couple I tested. I put in a ticket with Beaker as well as an issue with `puppet_install_helper`; we can switch back if we can figure out how to use it to default to Puppet 4 without having to set env variables everywhere.

I think some folks didn't want to have the acceptance tests running in Travis.  They are (sort of) setup to run here, though:
https://ci.virtapi.org/view/voxpupuli/job/puppet-nginx/14/console
I'm not sure why it's exiting 1 even though apparently successful.

wyardley:
Discussed this with @dhoppe at some length, and I do see now why the current solution is the better one... that is, doing `install_puppet_agent_on()` would only support agent type install.

That said, I don't love that puppet_install_helper only supports env variables for configuring the install type.

_if_ there's a way to set default env variables for rake and / or beaker in those configs, may be able to set a default that way?

I'm going to close this, and maybe file an issue so this doesn't fall through the cracks (though setting `PUPPET_INSTALL_TYPE` and / or `PUPPET_VERSION` will fix the problem with acceptance tests in the meantime).

Should probably also update the text in:
https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.github/CONTRIBUTING.md 
to reflect the new rake tasks, and the fact that there's no default nodeset on many modules (or that the default is per-module).

PR #956, opened 2016-11-01
Add fastcgi_param parameter to vhost resource

xaque208:
Without this change, specifying fastcgi parameters are not possible on
the vhost.  This change modifies vhost resource to allow passing the
fastcgi_param to the location resource directly, as is the case with a
few of the other fastcgi related parameters.

### DIFF
--- a/manifests/resource/vhost.pp
+++ b/manifests/resource/vhost.pp
@@ -46,6 +46,7 @@
 #   [*resolver*]            - Array: Configures name servers used to resolve
 #     names of upstream servers into addresses.
 #   [*fastcgi*]             - location of fastcgi (host:port)
+#   [*fastcgi_param*]       - Set additional custom fastcgi_params
 #   [*fastcgi_params*]      - optional alternative fastcgi_params file to use
 #   [*fastcgi_script*]      - optional SCRIPT_FILE parameter
 #   [*uwsgi_read_timeout*]  - optional value for uwsgi_read_timeout
@@ -248,6 +249,7 @@ define nginx::resource::vhost (
   $proxy_buffering              = undef,
   $resolver                     = [],
   $fastcgi                      = undef,
+  $fastcgi_param                = undef,
   $fastcgi_params               = "${::nginx::config::conf_dir}/fastcgi_params",
   $fastcgi_script               = undef,
   $uwsgi                        = undef,
@@ -645,6 +647,7 @@ define nginx::resource::vhost (
       proxy_set_body              => $proxy_set_body,
       proxy_buffering             => $proxy_buffering,
       fastcgi                     => $fastcgi,
+      fastcgi_param               => $fastcgi_param,
       fastcgi_params              => $fastcgi_params,
       fastcgi_script              => $fastcgi_script,
       uwsgi                       => $uwsgi,
--- a/spec/defines/resource_vhost_spec.rb
+++ b/spec/defines/resource_vhost_spec.rb
@@ -948,6 +948,14 @@ describe 'nginx::resource::vhost' do
         it { is_expected.to contain_file('/etc/nginx/fastcgi_params').with_mode('0770') }
       end
 
+      context 'when fastcgi_param => {key => value}' do
+        let :params do
+          default_params.merge(fastcgi_param: { 'key' => 'value' })
+        end
+
+        it { is_expected.to contain_nginx__resource__location("#{title}-default").with_fastcgi_param('key' => 'value') }
+      end
+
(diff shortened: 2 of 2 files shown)


wyardley:
This seems sane to me, but you'll need to add tests (and then squash commits).

xaque208:
Tests added.  Let me know if thats not enough.

xaque208:
Hmm, I'm not used to rubocop.  Any idea what I've done wrong.  The log doesn't load here.

wyardley:
@xaque208 Yeah, it can be kind of a hassle. Basically, before pushing, try to run:
`bundle install` (if hasn't already been run)
`bundle exec rake spec_clean`
`bundle exec rake spec_prep`
`bundle exec rake rubocop`
`bundle exec rake release_checks`

Currently, the complaints I see are:
```
Running RuboCop...
/private/tmp/puppet-nginx/.rubocop.yml: Metric/BlockLength has the wrong namespace - should be Metrics
Inspecting 20 files
................C...

Offenses:

spec/defines/resource_vhost_spec.rb:953:47: C: Style/SpaceInsideHashLiteralBraces: Space inside { missing.  (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
          default_params.merge(fastcgi_param: {'key' => 'value'})
                                              ^
spec/defines/resource_vhost_spec.rb:953:64: C: Style/SpaceInsideHashLiteralBraces: Space inside } missing.  (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
          default_params.merge(fastcgi_param: {'key' => 'value'})
                                                               ^
spec/defines/resource_vhost_spec.rb:957:1: C: Style/EmptyLinesAroundBlockBody: Extra empty line detected at block body end. 
```

You don't need to test _all_ the things from the example in location, but I'd test for absence when unset, and then test for presence of multiple variables, similar to what's done in the location example.

Also, you'll need to validate the parameter, right before:
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/vhost.pp#L443-L445
```
  if ($fastcgi_param != undef) {
    validate_hash($fastcgi_param)
  }
```

xaque208:
Alright, thank you for the information.  I just run `bundle exec rake spec` before pushing, not the rest of the rake tasks as you have shown here.  I'll come back to this.

wyardley:
You're also not actually putting the parameters into any templates, and you're testing the wrong thing (nginx__resource__location). 

Location does it a bit differently than we'd want to do it for vhost, see:
https://github.com/voxpupuli/puppet-nginx/blob/f29890840b3ff5ff235f5d4b4b59d3feacbec621/manifests/resource/location.pp#L374-L391
https://github.com/voxpupuli/puppet-nginx/blob/f29890840b3ff5ff235f5d4b4b59d3feacbec621/templates/vhost/locations/fastcgi.erb#L17-L21

The location manifest needs to change, but currently, it will use one of several templates, which causes some other problems. For vhost I _think_, you'll probably want to put in both of:

https://github.com/voxpupuli/puppet-nginx/blob/master/templates/vhost/vhost_footer.erb#L5
https://github.com/voxpupuli/puppet-nginx/blob/master/templates/vhost/vhost_ssl_footer.erb#L5
(which means you will probably also need to test both ssl and non-ssl contexts)

xaque208:
I'm building 2.3.1 locally for testing rubocop, so I'll sort that out in a moment.

For the testing you mention, I don't believe I'm testing the wrong thing.  Specifically what I'm testing, is that when you specify the $fastcgi_param parameter on the `nginx::resource::vhost` resource, that it gets passed to the `nginx::resource::location` resource as is, unmodified.  The tests for `nginx::resource::location` should handle everything that is necessary to validate this input, including the template content, but this PR doesn't touch any of that as its not a change captured in this PR.

That said, Is there a different change I should be making?  I'm under the assumption here that the location created by default for every vhost when `use_default_location` is `true` is where the params for the fastcgi should go, as looks to be the case with most of the other params when using `use_default_location`.   In which case, we just punt this param down the line to the location resource and do nothing if users disable lthe `use_default_location` param.

xaque208:
I've rubocop'd like a pro.  Hooray!

xaque208:
I still want to make sure this is the right change to make, so if my thinking above is off, please correct me.

wyardley:
I see what it's doing now and how (with the default location). I did find it a little confusing that it's in a different scope, but in any event, this is merged, and does behave the same way some similar / related params do. Thanks for fixing the Rubocop warnings.

xaque208:
Great, thank you @wyardley

Issue #818, opened 2016-06-15
HTTP->HTTPS


dowlingw:
Feature request: Add a `https_redirect` option to the `vhost` type that will:

- Implies `ssl_only` for the default location
- Installs an additional location with `ssl => false` that performs a temporary or permanent HTTP redirect to the same URL at HTTPS.

This will significantly reduce the amount of boilerplate resources required to do HTTPS-only sites with a friendly redirection for HTTP visitors.

I should be able to generate a PR, but will need some assistance with rspec because n00b.

pckls:
Happy to assist with the testing :)

wyardley:
I think $rewrite_to_https does this, though it would be easier to find if it had 'ssl' in the parameter name and / or was grouped with the SSL related parameters. Can you see if that works for you?

wyardley:
Feature already exists, though do want to support setting the redirect port, and possibly rename the config option to be more obvious.

Issue #872, opened 2016-09-05
Add more complex examples


c33s:
one thing is missing for this awesome module, a better documentation of complex examples. 

the mini examples are quite nice to get a quick overview but for easy an fast usage of this module it would be good to also have complex examples. they should be available as puppetcode and hiera examples.

having a simple `profile::nginx` i am try to get a hiera config for [symfony2][1] up and running.
```
class profile::nginx {
  class { 'nginx': }
  $vhosts = lookup('profile::nginx::vhosts', Hash)|$vhosts| { {} }
  $locations = lookup('profile::nginx::locations', Hash)|$locations| { {} }
  create_resources('nginx::resource::vhost', $vhosts)
  create_resources('nginx::resource::location', $locations)
```

```
server {
    server_name domain.tld www.domain.tld;
    root /var/www/project/web;

    location / {
        # try to serve file directly, fallback to app.php
        try_files $uri /app.php$is_args$args;
    }
    location ~ ^/(app_dev|config)\.php(/|$) {
        fastcgi_pass unix:/var/run/php5-fpm.sock;
        fastcgi_split_path_info ^(.+\.php)(/.*)$;
        include fastcgi_params;
        fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
        fastcgi_param DOCUMENT_ROOT $realpath_root;
    }
    location ~ ^/app\.php(/|$) {
        fastcgi_pass unix:/var/run/php5-fpm.sock;
        fastcgi_split_path_info ^(.+\.php)(/.*)$;
        include fastcgi_params;
        fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
        fastcgi_param DOCUMENT_ROOT $realpath_root;
        internal;
    }
    location ~ \.php$ {
      return 404;
    }

    error_log /var/log/nginx/project_error.log;
    access_log /var/log/nginx/project_access.log;
}
```
the following hiera config is currently not working, i get a `Duplicate declaration: Concat::Fragment[example.com-...` error.
i also was forced to use the `internal: true` setting on `loc1`  because of the `Cannot create a location reference without a www_root, proxy, location_alias, fastcgi, uwsgi, stub_status, internal, or location_custom_cfg defined` error.  i will post the working hiera file here in the comments as soon as i figured out (looks like i have to use `location_custom_cfg`).

```
profile::nginx::vhosts:
  'example.com':
    server_name: ['example.com','www.example.com']
[...]

PR #957, opened 2016-11-01
Rename rewrite_to_https => ssl_redirect (backwards-incompatible change)

wyardley:
This PR has some backwards-incompatible changes:
1. It renames `rewrite_to_https` to `ssl_redirect`. I know this will be a pain for some, but the parameter is hard to find since 'ssl' and 'redirect' are not in the current parameter name. Also, we aren't using the 'rewrite' directive, so the current name is a bit misleading (see #818, for some discussion of this). I've also grouped it along with the other SSL / TLS related parameters. 
1. It forces `ssl_port` and `listen_port` to be integers (there has already been a deprecation warning in place for a long time if they're set to strings)

Additionally, it:
1. Adds an option to override the redirect port with the new `ssl_redirect_port` parameter (open to suggestion on the exact naming of the parameters). 
1. Suppresses the location block on the non-SSL vhost
1. Suppresses other default items that aren't needed (~~logs~~ and index files) when ssl_only is set

This also gets rid of the `if ($ssl_protocol = "") {` (which apparently is un-needed here (per comments by @law), and is also bad style), updated the wording slightly, sets the default to 'false' rather than undef, and adds a test case for when $ssl_port is alternate.

Does anyone know of any reason this if statement is needed here? I think in normal contexts, if SSL or TLS is enabled, the other template will be used anyway.

This is also reworking #820 (try to suppress some additional locations in the non-SSL vhost when just redirecting, accomplished by setting `ssl_only`), and updating the conditional logic in this section.


wyardley:
Don't merge this one yet (even when it's passing tests).
I'm going to rework #789 as well, but probably need to take the conditional logic out of the template.

wyardley:
I think this also handles #789 as mentioned above, though I coded it slightly differently.

I've added additional tests to cover various scenarios that weren't really covered before, though please review them carefully and make sure I'm doing them in a sane way.

It also now fails instead of giving a deprecation warning when `listen_port` and `ssl_port` are set as strings, rather than integers; there has been a warning on the books for quite a while, and making sure they're ints avoids having to do the `$x + 0` trick for the comparison.

It should be possible to accomplish this transition in a softer / nicer way if we decide we want to... probably a lot of people are using this parameter.

wyardley:
@law noticed some redundant code in the template, and also mentioned that the logs can / should stay in by default; this all simplifies the code somewhat, and I re-pushed with those edits.

manifests/resource/vhost.pp
```diff
@@ -214,6 +216,8 @@
   $ssl_client_cert              = undef,
   $ssl_verify_client            = 'on',
   $ssl_dhparam                  = undef,
+  $ssl_force_redirect           = false,
```
hunner: or `ssl_redirect` to match `ssl_redirect_port`

manifests/resource/vhost.pp
```diff
@@ -616,7 +616,21 @@
     require => File[$vhost_dir],
   }
 
-  $ssl_only = ($ssl == true) and (($ssl_port + 0) == ($listen_port + 0))
+  # This deals with a situation where the listen directive for SSL doesn't match
+  # the port we want to force the SSL redirect to.
+  if ($ssl_redirect_port) {
+    $ssl_redirect_port_real = $ssl_redirect_port
```
hunner: Current practice is to use `$_ssl_redirect_port` instead of `$ssl_redirect_port_real`

wyardley: Updated.

manifests/resource/vhost.pp
wyardley: renamed as suggested.

manifests/resource/vhost.pp
```diff
@@ -214,6 +216,8 @@
   $ssl_client_cert              = undef,
   $ssl_verify_client            = 'on',
   $ssl_dhparam                  = undef,
+  $ssl_redirect           = false,
```
invidian: That would be nice to fix that whitelist.

xaque208:
Seems like maybe a documentation update somewhere is in order.

manifests/resource/vhost.pp
wyardley: Thanks! I missed that and one other when I renamed the parameter. Fixed now.

wyardley:
@xaque208 the parameter change itself is documented the way other params of the vhost resource are:
https://github.com/wyardley/puppet-nginx/blob/cc26a390a7aa5e1a1c3d83f5a881ca10be84f7db/manifests/resource/vhost.pp#L65-L68

As far as docs in general, yes, the current situation is pretty untenable, with examples needing to be updated, and the main classes not documented at all. And, an example using this param in README or examples would also be great, though one didn't exist before this change either.

There was an old PR that had some docs updates but never got merged; #941 is my attempt to get started on moving towards a 1.0 and adding a lot of the missing docs, hopefully with puppet-strings. Hopefully, I'll get a better idea of the exact format and conventions soon. Help with docs would definitely be a great way for people to help contribute!

xaque208:
I'm happy to write docs.  #872 is one very similar to an item I'm struggling to get working using this module.  A fastcgi application proxied to php-fpm.  I have the raw nginx config that works, but have yet to be able to replicate that in the module.  For docs @wyardley, there are lots of things in flight right now, so I'd only be wanting to write docs and give examples for items that were relatively stable.  Also, not sure where the majority of the conversation about upcoming work is happening, but some idea about where folks want to take this module would help flesh out examples, etc.  Happy to chat in IRC too.

wyardley:
@xaque208 Yeah, IRC (#voxpupuli) or the Puppet Community Slack in #voxpupuli is where most of the discussion happens; there are also community module triage online meetings (on Blue Jeans) every week on Thursdays.

I think with these community supported modules, it can be kind of hard as there isn't necessarily one consistent vision of how things should be. But we are trying to move towards a 1.0, which means mostly improving the docs, and trying to get the structure and interface of the module a little more in line with current best practices.

Even though the module itself may get some more structural changes, I don't know of any plans to make more big changes to the interface to the module beyond the one we just merged.

Even though the module has been around for a while, and has been somewhat widely used, because it's still pre 1.0, and because it's community supported, we're able to be a bit looser in terms of making changes like this without leaving in deprecation warnings etc. This can be frustrating as a user at times, but it does help make some of these improvements happen at a quicker pace and leaving less cruft.

xaque208:
This all sounds good.  Users can stick to older versions of the module if they require stability.  Looking forward to 1.0!

wyardley:
I think the requested changes have been made; can someone give me a final review / merge on this, assuming it looks good?

Issue #942, opened 2016-10-25
Upstreams do not depend on package


jarro2783:
The nginx package resource should come before the upstreams. I see the following error in my logs:

```
Could not set 'file' on ensure: No such file or directory @ dir_s_rmdir - /etc/nginx/conf.d/restapi_pool-upstream.conf20161025-2146-tn3v2u.lock
```

It looks like the config files are trying to be written before the nginx package has installed `/etc/nginx/conf.d`. I can fix it in my own code by writing:

```
Class['nginx::package'] -> Nginx::Resource::Upstream <||>
```

joshuaspence:
This seems to apply to other resources too, such as `Nginx::Resource::Map`.

wyardley:
@jarro2783 @joshuaspence: can you try the most recent release?
I think between #906 and #933 (and possibly one other?) this may now be fixed. If you're still seeing this with 0.5.0, comment here, and I'll reopen.

joshuaspence:
Nope, this is still an issue with 0.5.0. It looks like the problem was fixed for `Nginx::Resource::Vhost`, but not for `Nginx::Resource::Upstream`.

wyardley:
@jarro2783 @joshuaspence: If you're willing / able, can you test out #958 (you can append .diff or .patch to the PR URL to get a patch of it, if that helps).

jarro2783:
As far as I can tell that solves it.

wyardley:
Thanks @jarro2783. PR is now merged, so I'm going to close this issue for now.

PR #958, opened 2016-11-02
Add 'require' for parent dir of upstream, map, and geo configs as wel…

wyardley:
I think this should come a lot closer to resolving #942.

Few questions:

1. Should these be class specs, not defines? And should I have one test for the file existing, and another for it existing and having the require, or is this fine?
1. Should I avoid using `$conf_dir` within the resource manifests (since we've got `$::nginx::config::conf_dir`? Or is this fine?
1. I went back and forth on making some of these variables within the scope of `$::nginx::config`, but decided to leave it this way, but open to discussion. Upstream and vhost specifically differs depending on settings like `$confd_only`, and some of the other resources have their own directories, like `conf.mail.d`, so I'm not sure if making it more generic is worth the effort.

### DIFF
--- a/manifests/resource/geo.pp
+++ b/manifests/resource/geo.pp
@@ -16,7 +16,7 @@
 #   [*proxy_recursive*] - Changes the behavior of address acquisition when
 #                         specifying trusted proxies via 'proxies' directive
 #   [*proxies*]         - Hash of network->value mappings.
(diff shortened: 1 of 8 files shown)


dhoppe:
I think this PR is just fine and has already been tested via #942.

Issue #664, opened 2015-08-03
Log directory ownership and permissions do not respect OS


Maliuta:
The setting up of the $log_dir does not respect the variance in operating systems, and what the owner and group of that directory should be. Currently it is set to 
   owner => $global_user,
   group =>$global_group,

This should vary with different operating systems - for example in Debian/Ubuntu it should be:
   owner => 'www-data',
   group => 'adm',

wyardley:
One other option would be to not force creation of it? I think in most cases, the package handles creation of the log directories.

wyardley:
It seems like on most (all?) platforms, `$daemon_user` should also be the owner of the log directories

Default: nginx
ArchLinux: http
Debian / Ubuntu: www-data
FreeBSD: www

There isn't a direct equivalent that I can see for the group. Would it make sense to use daemon_user as the owner, and maybe make a new parameter for the log group as well?

wyardley:
I've got a branch that will set this explicitly (and, potentially, also force the mode). However, the existing code doesn't force the owner _or_ mode, and that may be a good thing. 

The default permissions, based on spec tests, are mode 0644), so I think we'd be seeing a lot of problems if the ownership and permissions were not already coming from the package, rather than from our module. I'm not totally convinced that the module should handle this if it doesn't have to.

Also, it seems like different package sources even within the same platform expect different ownership of the log directory and logfiles, for example, on CentOS the official nginx package seems to create the log dir and log files as root:root, but the epel (and possibly the Phusion / passenger) package do nginx:nginx.

It does seem like a good idea to force 0751 or 0700 on the directory permissions, though doing so might require adding another configurable knob for end-users who want to set more restrictive permissions.

Which platform are you actually seeing this problem come up on, and which version of the nginx package (i.e., which `package_source` and `manage_repos` settings are you using)? I'm a bit concerned that a fix might cause more problems than it solves.

That said, my first pass at this is at:
https://github.com/wyardley/puppet-nginx/tree/issues_664_log_ownership
I'm going to try to get a bit more discussion of how it should work before formally putting this in as a PR

wyardley:
Oh right, the File resource has it set to global_owner / global_group in
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/config.pp#L223
I can verify that it is forcing permissions as described above, so I guess we do have to figure out how to do this in a "smart" way.

wyardley:
[...]

PR #959, opened 2016-11-02
Set log directory ownership / permissions explicitly

wyardley:
Does this look sane as a fix for #664? This is one we should definitely try and get right, as it could cause problems if the ownership isn't correct on various platforms.

1. Should this also have an interface in `::nginx` as well as `::nginx::config`? I think with new parameters, it's better not to do that until we work out the exact structural changes.
1. @bastelfreak: can you confirm permissions on ArchLinux?

I added `log_mode` in nginx::config, as I think it's important to have a sane default that's restrictive (0700), but people often do also need to grant group / other permissions on the nginx logdir that are custom. We could make the default 0750 or 0751 if that makes more sense.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -24,6 +24,8 @@ class nginx::config(
   $global_group                   = $::nginx::params::global_group,
   $global_mode                    = $::nginx::params::global_mode,
   $log_dir                        = $::nginx::params::log_dir,
+  $log_group                      = $::nginx::params::log_group,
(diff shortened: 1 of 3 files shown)


dhoppe:
@wyardley Since you use a different set of user/ group you should use 0750 as mode. Otherwise logging tools like Logstash could have problems accessing the logs.

wyardley:
@dhoppe: you're right, updated config and spec tests to have it default to 0750.

PR #965, opened 2016-11-09
Fix Bug: ensure => absent was not working on nginx::resource::location

artberri:
- Fixes: ensure => absent was not working on nginx::resource::location
- Added tests to avoid happening again

I hope this helps. We are having this issue since we updated the module version. Let me know if I need to do some changes before merging.


wyardley:
I'll wait for someone else to weigh in, but at first glance, this looks sane to me.

PR #966, opened 2016-11-09
allow try_files and index in location resource

wyardley:
This is _not_ the fix I had been hoping to submit, reworking the location templates logic a bit more dramatically, but this resolves a problem I'm seeing, and I think should not introduce any new problems.


wyardley:
Crap, index_files isn't undef in locations by default :/ probably we don't want to generate these by default in alias blocks....

Issue #929, opened 2016-10-13
Apt can't authenticate package


seocam:
- Puppet: 4.3.2
- Ruby: 2.1.5p273
- Distribution: Debian 8.3
- Module version: 0.4.0

### How to reproduce (e.g Puppet code you use)

`include nginx`

### What are you seeing

I'm receiving daily emails from my cron with the following content:

```
Cron <root@74f4a22b-web> test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.daily )

/etc/cron.daily/apt: WARNING: The following packages cannot be authenticated!

nginx
```

If I run `apt-get update` I get:
```
W: GPG error: http://nginx.org jessie Release: The following signatures were invalid: KEYEXPIRED 1471427554
```

### What behaviour did you expect instead

Nginx packages should be correctly authenticated using GPG key provided by nginx.org

### Any additional information you'd like to impart

If I download the key and add to the servers manually I stop to get the errors.

```
wget https://nginx.org/keys/nginx_signing.key
sudo apt-key add nginx_signing.key
```

wyardley:
Seems similar to this issue (though of course different key):
https://puppet.com/blog/updated-puppet-gpg-signing-key
https://tickets.puppetlabs.com/browse/MODULES-3307
We also have a ticket for discussion of moving the signing key into the module itself, rather than pulling it directly from the nginx site (#651); this in and of itself might not fix this problem, but maybe we could add some code around it?

Not sure if anyone can think of a clean way to update the key without resorting to the hackery in the examples in the links above.

emrtldrgn:
Is there a reason that you are just not using the ppa?
`
apt::ppa { 'ppa:nginx/stable':
    ensure => present,
  }
`

wyardley:
Not sure what that is; you're welcome to contribute a pull request if you'd like, and someone can review it. Keep in mind, though, that we're moving towards making the non-vendor repository non-default very shortly

I don't think that will address the issue when signing keys expire though, will it?

kwisatz:
I'm seeing the same as OP, but the key doesn't seem to be outdated. Even when importing the nginx_signing.key manually, apt still refuses to authenticate them:

```
root@front1:~# apt-key list
/etc/apt/trusted.gpg
--------------------
pub   2048R/7BD9BF62 2011-08-19 [expires: 2024-06-14]
uid                  nginx signing key <signing-key@nginx.com>
```

[...]

PR #967, opened 2016-11-10
fix apt::source to use global ppa

micah-uber:
Fix source to use ppa that is available for nginx which will also solve #929

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -33,17 +33,13 @@ class nginx::package::debian(
 
     case $package_source {
       'nginx', 'nginx-stable': {
-        apt::source { 'nginx':
-          location => "http://nginx.org/packages/${distro}",
-          repos    => 'nginx',
-          key      => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62',
+        apt::ppa { 'ppa:nginx/stable':
+          ensure => present,
         }
       }
       'nginx-mainline': {
-        apt::source { 'nginx':
-          location => "http://nginx.org/packages/mainline/${distro}",
-          repos    => 'nginx',
-          key      => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62',
+        apt::ppa { 'ppa:nginx/stable':
+          ensure => present,
         }
       }
       'passenger': {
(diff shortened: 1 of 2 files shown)


manifests/package/debian.pp
```diff
@@ -33,17 +33,13 @@
...
+        apt::ppa { 'ppa:nginx/stable':
+          ensure => present,
         }
       }
       'nginx-mainline': {
-        apt::source { 'nginx':
-          location => "http://nginx.org/packages/mainline/${distro}",
-          repos    => 'nginx',
-          key      => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62',
+        apt::ppa { 'ppa:nginx/stable':
```
hunner: mainline is `ppa:nginx/development` according to https://launchpad.net/~nginx/+archive/ubuntu/development

wyardley:
@micah-uber: thanks! I think it's still been working in the acceptance tests and in my quick functional tests, but spoke to folks in the community meeting, and it seems like moving to ppa is a fine way to go.
Would you mind resolving the test errors, resolving @hunner's feedback above, and switching to https (this will then preempt #939)

Generally, with voxpupuli modules, they want the commits commits to all be squashed to one, so if you can squash as well (interactive rebase / force-push), that would be great.

wyardley:
A bunch of the failures have this error: `apt::ppa is not currently supported on Debian. at /home/travis/build/voxpupuli/puppet-nginx/spec/fixtures/modules/apt/manifests/ppa.pp:14 on node testing-worker-linux-docker-481429b0-3451-linux-10.prod.travis-ci.org`

wyardley:
It sounds like ppa is Ubuntu specific.

3flex:
As @wyardley said PPAs are Ubuntu specific. I don't see a good reason to make this change, unless the PPA were only used on Ubuntu and the current method were still used for other Debian-derived distros, but even then I don't think it's a good idea to have the development or mainline versions come from any source other than nginx.org for consistency's sake.

wyardley:
I'm going to close this one, though I agree we still do need to get a handle on things if there's a problem with signing key somewhere.

If you're seeing problems with the signing keys on new installs, can you open a new issue and provide `apt-cache policy nginx` output as well as the package related settings from your config, or else propose a new PR that reworks this in a way that preserves Debian support? I personally think it's probably not a good idea to use a different package install strategy for Ubuntu here.

PR #932, opened 2016-10-17
issue #348: vhost to server

renich:
Based on the discussion we had, I've renamed all instances of 'vhost' to
'server'.

The main reason for this is that NginX does not uphold the concept of
vHost: https://www.nginx.com/resources/wiki/start/topics/examples/server_blocks/

Stating:

```
"Note: “VirtualHost” is an Apache term. NGINX does not have Virtual
hosts, it has “Server Blocks” that use the server_name and listen
directives to bind to tcp sockets."
```

Besides, the extended use of the concept causes confusion and
misunderstanding when it comes to NginX.

I still need to work on two more requests.

PR #968, opened 2016-11-10
Rename all vhost ocurrences to server

### LINKED ISSUE #348: Rename vhost to server.d
I'd rename vhost to server.d for the following reasons:
- vhost is an apache term: Virtual Hosts. NginX's term is Server.
- The <term>.d convention is used widely accross services and distributions.

DavidS:
This closes #348 and closes #932. It was created by running

    git ls-files | xargs -i{} sed -i -e 's/vhost/server/g' {}

and manually fixing lint issues and renaming all files called '*vhost*'.


DavidS:
cc @wyardley

wyardley:
@DavidS Can't figure out how to submit a PR against your fork, but these are my formatting changes: https://github.com/wyardley/puppet-nginx/tree/rename_vhost_server_davids

Also got a couple instances of vHost.

FWIW, spec tests are passing.

3flex:
Food for thought: the "server" block can be used in several contexts in nginx...

- ngx_http_core_module
- ngx_http_upstream_module
- ngx_mail_core_module
- ngx_stream_core_module
- ngx_stream_upstream_module

This module supports some (all?) of those. Of course most people will assume a "server" will configure an HTTP server since that's most common, but perhaps resources should be namespaced somehow e.g. nginx::http::server and nginx::mail::server or something like that so that it's clear what type of server is being created.

I only mention because `nginx::resource::mailhost` also exists and that creates a server block too, but in the `mail` context. If you're renaming to align the resource name to the nginx config name then `mailhost`, `streamhost` and the upstream member(s) should be looked at too.

wyardley:
@3flex Yeah, I'm not sure if there's an easy way to handle that. I agree that it's a little confusing, though maybe we should just do the great renaming (which is going to have to be re-done now, by the way, since stuff got merged), and make any other changes iteratively?

If it were namespaced `nginx::http::server`, how would we deal with situations like
`nginx::nginx_vhosts`? Would they become `nginx::nginx_http_servers`?

hunner:
This needs a rebase (or a rerun of the sed)

3flex:
I'd remove the "nginx" prefix from the parameter and have nginx::http_servers.

For the future, I THINK this makes sense...

* Resources
  * nginx::resource::geo => nginx::http::geo
  * nginx::resource::location => nginx::http::location (this one's a bit funny, it doesn't quite match the pattern of the others)
  * nginx::resource::mailhost => nginx::mail::server
  * nginx::resource::map => nginx::http::map
  * nginx::resource::streamhost => nginx::stream::server
  * nginx::resource::upstream => nginx::http::upstream
  * nginx::resource::vhost => nginx::http::server
  * nginx::resource::upstream::member => nginx::http::upstream::server
* Parameters on `nginx` class
  * geo_mappings => http_geos
  * string_mappings => http_maps
  * nginx_locations => http_locations (again, a bit funny...)
  * nginx_mailhosts => mail_servers
  * nginx_streamhosts => stream_servers
  * nginx_upstreams => http_upstreams
  * nginx_vhosts => http_servers
  * nginx_vhosts_defaults => http_servers_defaults

This is just one option, I'm not overly attached, but it's now very explicit as to what each resource will configure. One inconsistency with the above is locations, perhaps could go with `nginx::http::server::location`?

sacres:
@wyardley Submitted PR #980 to address this for the rebase/sed and [vV]host

wyardley:
Replaced by #980

Issue #971, opened 2016-11-14
nginx::locations puts locations in wrong order


kwisatz:
- Puppet: 3.7.5
- Ruby: 2.1.5
- Distribution: debian
- Module version: git-master

I realize this might not necessarily belong here, but it is specifically creating problems with nginx and I have the feeling that it actually changed with my upgrade from 0.5.0 to the git-master branch.
Also, I haven't found any information in the documentation on [create_resources](https://docs.puppet.com/puppet/latest/reference/function.html#createresources) that would point to it being responsible for this behavior.

Our problem is, that nginx locations are put in the wrong order. With regex locations, order matters and in our case it puts a return 404 in front of the location it should actually match first, which makes my site 404 on all requests that should go to the main controller.

It does not seem to order by order of definition, nor alphabetically by the hash key/name. But I can't find the place where the locations are sorted.

wyardley:
When did you clone master? There was a recent PR that just reworked the logic of locations.
CCing @xaque208 

Can you provide an example of the expected vs. actual ordering?

FWIW, this is where most of the location logic happens:
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp
https://github.com/voxpupuli/puppet-nginx/blob/master/templates/vhost/location.erb
https://github.com/voxpupuli/puppet-nginx/tree/master/templates/vhost/locations

called from

kwisatz:
Cloned it today, a few hours back, approx. 5.

The expected ordering would be as I have it in hiera:

```yaml
  'dev':
    ensure: present
    vhost: "%{hiera('vhost')}"
    www_root: "/var/www/%{hiera('vhost')}/current/web"
    fastcgi: phpfcgi
    location: '~ ^/(app_dev|config)\.php(/|$)'
    location_cfg_append:
      fastcgi_split_path_info: '^(.+\.php)(.*)$'
      fastcgi_param SCRIPT_FILENAME: $realpath_root$fastcgi_script_name
      fastcgi_param DOCUMENT_ROOT: $realpath_root
  'prod':
    ensure: present
    vhost: "%{hiera('vhost')}"
    www_root: "/var/www/%{hiera('vhost')}/current/web"
    fastcgi: phpfcgi
    location: '~ ^/app\.php(/|$)'
    location_cfg_append:
      fastcgi_split_path_info: '^(.+\.php)(.*)$'
[...]

PR #972, opened 2016-11-14
fix validation range for location priority

wyardley:
See #971 
Fix the validation range for location's "priority" attribute to match docs. Using a higher value seems to cause unpredictable behavior.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -362,8 +362,8 @@ define nginx::resource::location (
     fail('$priority must be an integer.')
   }
   validate_array($rewrite_rules)
-  if (($priority + 0) < 401) or (($priority + 0) > 899) {
-    fail('$priority must be in the range 401-899.')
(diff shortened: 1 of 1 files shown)


wyardley:
It looks like ssl priority adds 300, which might be why it's 899, not 599?

 https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp#L436
But since it's adding 300 already in the code, I think the validation step should still have 599 as the max?

nibalizer:
Seems to jive with the docs. I think we should land this. Just be aware a few users might show up with 'it used to work ARGH'

I think we're fixing a bug here, not making a backwards incompat change

wyardley:
@nibalizer Yeah, sorry, two unrelated questions. This PR is not backwards-incompatible, just was asking for general feedback about if it looked sane. I _think_ this is the right behavior.

Issue #964, opened 2016-11-08
Nginx location as exported resource not working anymore


ITler:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: current pc1 with puppet 4.8.0
- Ruby: from pc1(irrelevant for this issue)
- Distribution: Ubuntu 14
- Module version: 0.5.0 (forge)

### How to reproduce (e.g. Puppet code you use)

I need to define a location on an application server as an *exported resource*, that will get realized on another host with nginx installed. So on my appserver I only use:

    create_resources('@@nginx::resource::location', $locations, $location_params)

### What are you seeing

In an older version of `jfryman-nginx` (some version not far before the last 9999999) this worked fine. With the last version of jfryman and the current 0.5.0 from voxpupli this doesn't work anymore.

### What behaviour did you expect instead

I would have expected to have things working, like they did before.

### Output log

    Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Unknown variable: '::nginx::config::proxy_redirect'. at /etc/puppetlabs/code/environments/dev/modules/nginx/manifests/resource/location.pp:165:27 at [...]

### Any additional information you'd like to impart
Of cause, the appserver doesn't have an nginx installation, so I can't call `include ::nginx`

However when explicitly defining *proxy_redirect* in $location_params this seem to work. Then issue cascading to next parameter.. and so on... This means, that every parameter definition of `nginx/manifests/resource/location.pp` assigning a default value from `$::nginx` will cause this issue. I think the obvious approach (usual pattern) should be to assign default values from `$::nginx::parms`, which could be included beforehand.

Kind regards
ITler

wyardley:
I kind of get what you're doing here, but I'm not totally sure what the expected behavior is, or what changes exactly caused this.

Is it possible to test with current git master branch and see if the behavior works as you expect? The next release will get rid of the `nginx::config` namespace that was introduced a while back (i.e., all parameters will just be defined at top scope again). However, I'm not sure if that will change this behavior. fwiw, `proxy_redirect` defaults to `undef`, not sure if that relates to your problem, though I think you could set an explicit default when you override the class (when you instantiate the class, or in hiera).

Usually, as I understand it, it's parameters that vary depending on, say, OS or other factors, are in params, other defaults are just assigned directly in the class params (if they're overridable). Does that kind of make sense?

[...]

PR #973, opened 2016-11-16
Support exported resources  without installing nginx on a node.

ITler:
This should provide a solution for the use case in #964. The implementation introduces a major conditional for deciding whether to implement nginx (default) or not. If not, all OS dependent and independent variables are defined and can be used when defining exported resources i.e. for resources in `manifests/resource/*`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -257,7 +257,7 @@ class nginx (
   validate_bool($super_user)
   ### END VALIDATIONS ###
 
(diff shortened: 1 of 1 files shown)


ITler:
2nd commit added as tests failed. This is ugly, but as long as `$package_ensure` is allowed to hold *semver* strings (e.g. 3.4.1), I currently see no other chance to

PR #974, opened 2016-11-16
Support exported resources without installing nginx on a node.

ITler:
This should provide a solution for the use case in #964. The implementation introduces a major conditional for deciding whether to implement nginx (default) or not. Therefore I reuse and check `$package_ensure` for a flag called `omit`. If `omit` given, all OS dependent and independent variables are defined and can be used when defining exported resources i.e. for resources in `manifests/resource/*` without installing nginx.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -257,36 +257,38 @@ class nginx (
   validate_bool($super_user)
   ### END VALIDATIONS ###
 
-  class { '::nginx::package':
-    package_name   => $package_name,
-    package_source => $package_source,
-    package_ensure => $package_ensure,
-    package_flavor => $package_flavor,
-    notify         => Class['::nginx::service'],
-    manage_repo    => $manage_repo,
-  }
+  if $package_ensure != 'omit' {
+    class { '::nginx::package':
+      package_name   => $package_name,
+      package_source => $package_source,
+      package_ensure => $package_ensure,
+      package_flavor => $package_flavor,
+      notify         => Class['::nginx::service'],
+      manage_repo    => $manage_repo,
+    }
 
-  include '::nginx::config'
-  include '::nginx::service'
+    include '::nginx::config'
+    include '::nginx::service'
 
-  Class['::nginx::package'] -> Class['::nginx::config'] ~> Class['::nginx::service']
+    Class['::nginx::package'] -> Class['::nginx::config'] ~> Class['::nginx::service']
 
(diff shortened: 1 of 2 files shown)


wyardley:
@ITler: This will be implemented in #938 (still needs a bunch more work)... the design we discussed at most recent community modules triage meeting is slightly different than what's there now (https://etherpad.openstack.org/p/puppet-community-triage), basically, `manage_repo` will go away, `package_source` will become an enum with 'unmanaged' as one of the options (to manage the repo source), and there will be a new boolean `manage_package`, which will control whether the package is installed (though it will be controlled in a different way than it's currently implemented in the PR).

There is already a parameter to manage the service or not (it's currently `service_manage`, though perhaps it should be renamed `manage_service` for consistency. I think it's overreach to not manage the service just because the package itself isn't being ensured.

In general, I'm not sure if this is a supported use case for the module; we are working on making the general changes outlined above, which I think may help somewhat.

I think you're right that `nginx::conf_dir` should be used instead of `nginx::config::conf_dir`, though it should work as-is (since it's internal), I'm not sure how a few of those survived the purge. However, I can address those separately in a separate PR.

ITler:
> In general, I'm not sure if this is a supported use case for the module
Do you refer my use case? Well, it already did support my use case in past, otherwise my hosting environment wouldn't run. It would be great to get this support back. However using exported resources to transfer information from one node to another (which can't gather these information in any other way - no etcd or consul or whatever) is a basic concept of puppet.

What else do you expect me to do here?

wyardley:
So, if the question is, can it be supported to not manage the package / service, yes, that should be possible, however, I think it needs to be implemented slightly differently.

As far as your specific timeline, you may want to use a local fork if you have specific requirements in the short term.

xaque208:
I'm -1 on this change.  I see the value of using an exported resource, but I think you can get functionally the same thing without exported resources.  For example, if `profile::myapp` exports an item to the nginx server, the exported code could be relocated into `profile::myapp::web` that is included directly on the nginx server.  Then the code is still namespaced in a clean way, but the nodes are including only the pieces of code they are responsible for.  I'm making a bunch of assumptions on your use case.  I'm happy to chat more about how you might avoid exported resources for this work.

nibalizer:
I'm also dubious of this change.

@ITler can you post some of the code you use now, to help me grasp your use?

wyardley:
I'm going to close this one as a wontfix type thing for now, #964 is still open, so feel free to post some more example code for @nibalizer there.

PR #980, opened 2016-11-30
Rename v[hH]ost to server everywhere

sacres:
Replaces PR: #968 This addresses rebase/sed and [vV]host as well as linting.


wyardley:
This looks good to me, can you just tidy up the rest of the hashrocket and = alignment, and then I'll merge? Let's also try to not merge anything else until this gets merged so we don't have to redo it again.

@3flex: I totally agree that your way is a bit more elegant, unfortunately, I don't have the time / energy to implement it properly, however, if someone wants to pick up that torch, I'd be happy to wait on the next release until we can rework this to use your model.

manifests/resource/location.pp
```diff
@@ -393,41 +388,64 @@
...
+  } elsif ($uwsgi != undef) {
+    $content_real = template('nginx/server/locations/uwsgi.erb')
+  } elsif ($www_root != undef) {
+    $content_real = template('nginx/server/locations/directory.erb')
+  } elsif ($try_files != undef) {
+    $content_real = template('nginx/server/locations/try_files.erb')
+  } else {
+    $content_real = template('nginx/server/locations/empty.erb')
+  }
+
```
3flex: This block shouldn't have been added back. I think you'll have to do the search/replace on the latest code from master.

manifests/resource/location.pp
```diff
@@ -261,9 +259,6 @@
   if ($fastcgi_split_path != undef) {
     validate_string($fastcgi_split_path)
   }
-  if ($fastcgi_index != undef) {
-    validate_string($fastcgi_index)
-  }
```
3flex: This shouldn't be removed

sacres:
@wyardley @3flex Ok, changes done. Thanks.

3flex:
@wyardley likewise I'm not in a position to rename things according to my suggestion in #968. Once this is merged though consider opening a new issue to track that greater renaming if you wish.

wyardley:
@3flex good catch on that other stuff.
There are still some misaligned items, but I can fix that separately in another PR.
I agree that we should create an issue about the structural stuff, though not sure how many changes to the interface we want to subject users to....

Fabian1976:
Common people. Breaking changes are major bumps. Not minor. Read semver.org.

wyardley:
Aside from the fact that this was merged and released years ago:

https://semver.org/#spec-item-4

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

PR #986, opened 2016-12-16
Added auth_request configuration capability

mvisonneau:


wyardley:
Can you add spec tests?

mvisonneau:
@wyardley definitely, sorry about that ! :)

wyardley:
This looks good to me. It would probably have been good to check for absence of it as well in the default case (undef), but I think this is good enough. Thanks for squashing commits.

PR #987, opened 2016-12-20
add passenger_package_ensure parameter to allow pinning passenger version

wyardley:
I'd have to integrate this into #938, but needed this Right Now at work, and thought I'd add it in.

Basically, the passenger and nginx versions for the passenger repo need to be in lockstep, or problems come up, and setting `package_ensure` for nginx without being able to lock passenger version too causes some problems.


wyardley:
Tests now passing

PR #989, opened 2016-12-21
Bump puppet minimum version_requirement to 3.8.7

juniorsysadmin:
We currently only run automated tests against Puppet 3 latest and
therefore cannot guarantee that this module works as is expected with
earlier Puppet 3 versions

Also remove deprecated pe version_requirement

Also remove explicit references to Ruby and Facter versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,7 @@ This module manages NGINX configuration.
 
 ### Requirements
 
-* Puppet 3.0.0 or later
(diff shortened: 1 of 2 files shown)


wyardley:
This looks pretty good to me, I've been thinking the same thing.
I could go both ways in terms of the Ruby version requirement in docs, but I think it'll be OK without.

Issue #991, opened 2016-12-22
add_header quoting issue for Content-Security-Policy


bryangwilliam:
I am unable to use add_header for adding a Content-Security-Policy header to a server. Because the  CSP syntax uses single-quotes explicitly (https://content-security-policy.com/) for certain terms, the single-quotes that are being used to wrap the header value (added in #804) cause the nginx syntax to break.

Passing `{ "Content-Security-Policy" : "default-src 'self';" }` into the $add_header parameter of nginx::resource::server will result in the configuration `add_header 'Content-Security-Policy' 'default-src 'self';'`, which is invalid. I tried using double-quotes in the CSP header as a workaround, but it breaks on certain keywords in the browser. You can use the raw_prepend option as a workaround, but a more elegant solution would be nice.

This could be fixed by either
- changing to double-quotes in the template when applying the add_header values
- allowing users to supply their own quotes
- something else ?

wyardley:
Not sure how 
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_server_spec.rb#L1197-L1227
are passing as of now, since they're not matching the single quotes.

Does #992 resolve this for you? (note that if you're not using git, you may need to change the filenames, due to the vhost => server rename)

bryangwilliam:
That seems to do the trick.

Your comment about possibly having double quotes in the header is valid. I just did some research and it appears to me that DQUOTE is the acceptable delimiter for quoted-strings within the HTTP syntax (https://tools.ietf.org/html/rfc7230 "3.2.6.  Field Value Components"). As best I can tell, this means there is no case where you can include an unescaped double-quote within a header value (DQUOTE is explicitly disallowed as a value in a token and must be escaped within a quoted-string since double-quotes are the delimiters). Granted this is just my interpretation, but it makes sense in my head.

benh57:
I have the same problem but wiht the 'always' parameter to Access-Control-Allow-Origin. This used to work with this module, but now that the value is being single quoted, it fails. 

      add_header           =>  { 'Access-Control-Allow-Origin'      => '"*" always' } ,

results in:

  add_header              'Access-Control-Allow-Origin' '"*" always';

Which fails in the browser with:

The 'Access-Control-Allow-Origin' header contains multiple values '"*" always', but only one is allowed.

bryangwilliam:
I have been doing some more reading on this, and I think my initial interpretation was incorrect. 

[...]

PR #992, opened 2016-12-23
Use double, vs single quotes around add_header values (#991)

wyardley:
Not sure how 

https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_server_spec.rb#L1197-L1227
passes either before / after, but this seems to pass tests. I'm going to mark this as in-progress, though.

### DIFF
--- a/spec/defines/resource_server_spec.rb
+++ b/spec/defines/resource_server_spec.rb
@@ -1200,12 +1200,7 @@ describe 'nginx::resource::server' do
         end
 
         it 'has correctly ordered entries in the config' do
-          is_expected.to contain_concat__fragment("#{title}-header").with_content(%r{
-            %r|
-            \s+add_header\s+header1 test value 1;\n
-            \s+add_header\s+header2 test value 2;\n
-            \s+add_header\s+header3 test value 3;\n
-            |})
+          is_expected.to contain_concat__fragment("#{title}-header").with_content(%r{\s+add_header\s+"header1" "test value 1";\n\s+add_header\s+"header2" "test value 2";\n\s+add_header\s+"header3" "test value 3";\n})
         end
       end
 
@@ -1218,12 +1213,7 @@ describe 'nginx::resource::server' do
         end
 
         it 'has correctly ordered entries in the config' do
-          is_expected.to contain_concat__fragment("#{title}-ssl-header").with_content(%r{
-            %r|
-            \s+add_header\s+header1 test value 1;\n
-            \s+add_header\s+header2 test value 2;\n
-            \s+add_header\s+header3 test value 3;\n
-            |})
+          is_expected.to contain_concat__fragment("#{title}-ssl-header").with_content(%r{\s+add_header\s+"header1" "test value 1";\n\s+add_header\s+"header2" "test value 2";\n\s+add_header\s+"header3" "test value 3";\n})
         end
       end
     end
--- a/templates/server/server_header.erb
+++ b/templates/server/server_header.erb
@@ -131,7 +131,7 @@ server {
 <% end -%>
 <% if @add_header -%>
   <%- @add_header.keys.sort.each do |key| -%>
-  add_header              '<%= key %>' '<%= @add_header[key] %>';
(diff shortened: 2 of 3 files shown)


wyardley:
TODO: also need to add a test case where a value has single quotes in it, e.g.,
` "default-src 'self';" `
Of course, there could be the reverse case where there are double-quotes, so not sure how "smart" we want to get about this in code.

wyardley:
@xaque208 @bastelfreak does this change make sense to you all?
And any ideas why the current test passes in either case when it seems like it shouldn't pass either way (because it's just matching \s+ in between with no literal single or double quotes)?

wyardley:
@xaque208 IIRC, we use `%r{}` because the Voxpup Rubocop conventions dictate it. I don't think it's that different from a normal regex, except that you don't have to escape certain things (slashes, for example). I don't think it should be causing a problem in this case, though.

For example (I'm not a Ruby person, so sorry if this is not the best or most idiomatic):
```
#!/usr/bin/env ruby

string = 'test1 "test test" test3'
regexp = Regexp.new(%r{test1\s+test test\s+test3})
puts regexp.match(string)
regexp2 = Regexp.new(%r{test1\s+"test test"\s+test3})
puts regexp2.match(string)
```
returns
```
% ./test2.rb

test1 "test test" test3
```

wyardley:
@rnelson0 @bastelfreak Any thoughts on the tests issue? Also, more generally, does anyone have an opinion about whether there are any cases where the change to double quotes would cause a problem here?

rnelson0:
I don't see how it would matter here, since this is inside the plain text portion of the ERB, not the interpreted portion. Because we don't know how it is passing, I would change either [L1199](https://github.com/voxpupuli/puppet-nginx/blob/1d903c197d5b49d867f3b2022bb305da09a6da6a/spec/defines/resource_server_spec.rb#L1199) or [L1205-1207](https://github.com/voxpupuli/puppet-nginx/blob/1d903c197d5b49d867f3b2022bb305da09a6da6a/spec/defines/resource_server_spec.rb#L1205-L1207) so they do not match and see how that behaves.

wyardley:
Seems to still pass for me even if I change https://github.com/voxpupuli/puppet-nginx/blob/1d903c197d5b49d867f3b2022bb305da09a6da6a/spec/defines/resource_server_spec.rb#L1199 to
```
      context 'when add_header is set' do
        let :params do
          default_params.merge(add_header: { 'header3' => 'test value 3', 'header2' => 'test value 2', 'header1' => 'asdfsadfdafsafdsfads value 1' })
```

rnelson0:
Wow, that's amazing, clearly a bad test.

mterzo:
Rubocop was updated and replaced a confit option

From their changelog. 

#3737: Add new Style/MethodCallWithArgsParentheses cop. (@dominh)
Renamed MethodCallParentheses to MethodCallWithoutArgsParentheses. (@dominh)

https://github.com/bbatsov/rubocop/pull/3797/files

wyardley:
@xaque208 looking at it more closely, this does seem kind of odd to me; not totally getting why there are the nested %r expressions:
```
with_content(%r{
            %r|
            [...]
            |})
```
Having it all on one line does seem to work, though it's messy and make a long line (but does pass rubocop)
```
          is_expected.to contain_concat__fragment("#{title}-header").with_content(%r{\s+add_header\s+"header1" "test value 1";\n\s+add_header\s+"header2" "test value 2";\n\s+add_header\s+"header3" "test value 3";\n})
```

wyardley:
```
        it 'has correctly ordered entries in the config' do
          is_expected.to contain_concat__fragment("#{title}-header").with_content(%r{
            \s+add_header\s+"header1"\s"test\svalue\s1";\n
            \s+add_header\s+"header2"\s"test\svalue\s2";\n
            \s+add_header\s+"header3"\s"test\svalue\s3";\n
          }x)
```
This will also work, though not sure if it's an improvement in terms of readability.

wyardley:
I'm going to update it to use the test that should actually work, even if it could be formatted more nicely... if folks have suggestions on how to do it in a prettier way, let me know.

PR #1001, opened 2017-01-18
modulesync 0.19.3

dhollinger:


wyardley:
Seems like a pretty small rubocop error?

rnelson0:
Delete any examples that are repeats (we've found one or two but it's rare) and surround any remaining examples with `#rubocop:disable RSpec/RepeatedExample` and `#rubocop:enable RSpec/RepeatedExample` and that should resolve.

rnelson0:
Replaced by #1004, because I forgot I could push to this branch :D

PR #1003, opened 2017-01-20
Permissions on fastcgi_params and uwsgi_params files (#1002)

### LINKED ISSUE #1002: fastcgi_params file set to permission 770 by default
For all versions of the Puppet module, the `fastcgi_params` file is set to permissions 770. This seems really weird to me. Since it is a normal file, it should be 644 like everything else.

wyardley:
This attempts to resolve #1002 (but curious if anyone knows why the permissions on `fastcgi_params` and `uwsgi_params` were 0770 and 0660 respectively, rather than 0644 or 0444).


rnelson0:
@wyardley you can rebase this now that the modulesync is complete.

hunner:
lgtm 👍

hunner:
Closes #1002

PR #1005, opened 2017-01-23
Added support for proxy_next_upstream directive.

HeWhoWas:
This adds support for the proxy_next_upstream directive.

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream

### DIFF
--- a/manifests/resource/vhost.pp
+++ b/manifests/resource/vhost.pp
@@ -104,6 +104,7 @@
 #   [*proxy_method*]            - If defined, overrides the HTTP method of the
 #     request to be passed to the backend.
 #   [*proxy_set_body*]          - If defined, sets the body passed to the backend.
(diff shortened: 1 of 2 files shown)


wyardley:
This needs tests (and it looks like there are some conflicts, so you may need to rebase against upstream master). It looks like yours may have been forked before the vhost => server switch.

It's not required, but I also think it would be best if the parameter could be added to all the scopes where it applies (main http config, location) as well as the 'server' resource.

Also there are some proxy related directives in the streamhost section; not sure if this setting is relevant there.

HeWhoWas:
No worries, this was cherry picked from our internal fork. I'll never be given the appropriate amount of time to complete it as it should be, so I'll close the request off.  Thanks!

PR #1009, opened 2017-01-30
Fixed invalid path in upstream member resource and added upstream weight configurable parameter to the upstream member.

triforce:


wyardley:
The new parameter should have tests.
See my other notes in the review
Good catch with the typo; if possible, can you break it out into a separate PR, though?

triforce:
I have cherry picked the typo fix commit on pull request 1010.

PR #1011, opened 2017-01-31
Add upstream_weight parameter into the upstream and upstream member…

triforce:
… resource definitions. Added new parameter as optional value in the upstream_member/upstream_members templates.


triforce:
This is the PR that requires tests correct?

wyardley:
@triforce Yeah, should have tests for this one. It looks like maybe we're not even testing for that fragment being there too, if my quick read is correct.

triforce:
You mean the upstream_member fragment?

manifests/resource/upstream.pp
```diff
@@ -10,6 +10,7 @@
 #   [*upstream_cfg_prepend*]  - It expects a hash with custom directives to put before anything else inside upstream
 #   [*upstream_fail_timeout*] - Set the fail_timeout for the upstream. Default is 10 seconds - As that is what Nginx does normally.
 #   [*upstream_max_fails*]    - Set the max_fails for the upstream. Default is to use nginx default value which is 1.
+#   [*upstream_weight*]       - Set the weight for the upstream. Default is to use nginx default value which is 1.
```
vinzent: reading the erb template it seems to me the default is not to set the param and use whatever default nginx has. 

Reading this line I'd expect that it writes the param with value 1 in config if I don't specify anything.

manifests/resource/upstream.pp
```diff
@@ -49,6 +50,7 @@
   $upstream_fail_timeout                    = '10s',
   $upstream_max_fails                       = undef,
   Enum['http', 'stream'] $upstream_context  = 'http',
+  $upstream_weight                          = undef,
```
vinzent: Could you add the data type? if its integer:  `Optional[Integer] $upstream_weight = undef`

manifests/resource/upstream/member.pp
```diff
@@ -14,6 +14,7 @@
 #   [*server*]                  - Hostname or IP of the upstream member server
 #   [*port*]                    - Port of the listening service on the upstream member
 #   [*upstream_fail_timeout*]   - Set the fail_timeout for the upstream. Default is 10 seconds
+#   [*upstream_weight*]         - Set the weight for the upstream. Default is to use nginx default value which is 1.
```
vinzent: same as above

manifests/resource/upstream/member.pp
```diff
@@ -40,6 +41,7 @@
   Enum['present', 'absent'] $ensure = 'present',
   Integer $port                     = 80,
   $upstream_fail_timeout  = '10s',
+  $upstream_weight        = undef,
```
vinzent: same as above - use puppet4 data types

manifests/resource/upstream.pp
triforce: Fair enough I will change that.

triforce:
All changes requested have been completed.

manifests/resource/upstream.pp
vinzent: @triforce maybe forgot to push?

vinzent:
@triforce  and travis has some issue: https://travis-ci.org/voxpupuli/puppet-nginx/builds/221250489

manifests/resource/upstream.pp
triforce: The default is now set in 4456d4c.

vinzent: hmm... i would have preferred to have just reformulated the doc line. Now you have a change that affects every user. 

And how is one supposed to pass set the param undef? I don't think this is possible because then the default of 1 will slip in again.

triforce: Apologies that is how I read your comment. I will change that back.

triforce:
@vinzent could you advise on what tests this needs?

wyardley:
@triforce 
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_upstream_spec.rb#L80-L89
is where the existing tests for that fragment are, though I think for this, you'd want to make a new context in that same file.

You can get some help on the Puppet community Slack (#testing or #voxpupuli channels) or on IRC (#voxpupuli) if you want additional help on what to test and how.

Also, voxpupuli usually asks that the commits be squashed down to 1.

bastelfreak:
Hi @triforce. Can you please add tests for this, squash it down and rebase against latest master?

bastelfreak:
Also please take a look at the used email address in your commits. It isn't associated with your github account.

SaschaDoering:
The `weight` parameter was part of https://github.com/voxpupuli/puppet-nginx/commit/f0bf83a1abac6c2ee7fe7257a37514319a96f0fa which is merged. So I think this PR is obsolete by now.

ekohl:
You're right @SaschaDoering. Thanks!

PR #1012, opened 2017-01-31
Changed upstream_member.erb template directory path to match new loca…

triforce:
… resource definitions. Added new parameter as optional value in the upstream_member/upstream_members templates.


wyardley:
Looks good to me

PR #1016, opened 2017-02-02
Switched apt::source key from string to hash.

### LINKED ISSUE #995: Fix deprecated apt::source usage
Use the new syntax for apt 2.x

darkstego:
Updated to resolve #995 for new apt usage.

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -37,21 +37,21 @@ class nginx::package::debian (
         apt::source { 'nginx':
           location => "https://nginx.org/packages/${distro}",
           repos    => 'nginx',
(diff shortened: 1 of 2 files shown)


dhollinger:
@darkstego could you update the unit tests to account for the change and test?

https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb#L162
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb#L186

darkstego:
Updated unit tests. All Green.

vinzent:
thanks @darkstego !

Issue #1019, opened 2017-02-14
Add "udp" for "listen_port" parameter, add stream resource example into README


jurim76:
Please add "udp" option for "listen_port" parameter (need for nginx UDP load-balancer for example).
Also stream resource usage is not documented in README

```
stream {
  server {
    listen 514 udp;
  }
}
```

wyardley:
I think this should probably be accomplished with a new parameter, like `listen_protocol` vs. trying to extend `listen_port`.

It shouldn't be too hard, if I get a chance, I'll try to create a PR for this. As far as docs fixes, if you have time to write up a PR (or even draft the text or post a patch), it will definitely be reviewed.

jurim76:
@wyardley 

Thank you! I've added an example patch for README.md file

[README.md_patch.txt](https://github.com/voxpupuli/puppet-nginx/files/777218/README.md_patch.txt)

wyardley:
Actually, I started looking at implementing this, and I think `listen_options` will do what you want with the module as-is.
In fact, your readme seems to suggest the same thing. Can you explain why that doesn't work?

wyardley:
I was thinking of a more general purpose example, but that works... I submitted #1021
I'm pretty sure `listen_options` is sufficient, so I'm going to close this for now -- reoopen or comment if you think there's a reason that won't work.

PR #1021, opened 2017-02-16
docs fix from @jurim76

wyardley:
Add streamhost example to README from @jurim76 (#1019)


wyardley:
I think it would be good to have a more generic example too, but more docs seem like a good thing to me.

roidelapluie:
Thanks!

Issue #1015, opened 2017-02-02
Using ssl_port without ssl => true makes module fail silently


mzhaase:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet:              4.8.1
- Ruby:                  ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]
- Distribution:       Ubuntu 16.04
- Module version: 0.6.0

## How to reproduce (e.g Puppet code you use)
```
class { 'nginx': }
nginx::resource::server {'<name>':
      www_root    => '/var/www/test',
      listen_port => 443,
      ssl_port    => 443,
}
```

## What are you seeing
This creates <name>.conf in /etc/nginx/sites-enabled which only contains:

```
location / {
    root      /var/www/test;
    index     index.html index.htm index.php;
  }
```

## What behaviour did you expect instead
Any kind of error message.

wyardley:
I'm not sure if this is the right fix, but I'm proposing one in #1022

wyardley:
#1022 has been merged; hopefully this is sufficient.

PR #1022, opened 2017-02-16
warn if $ssl=false but $ssl_port == $listen_port (#1015)

wyardley:
See #1015 
I would appreciate review on this... I'm not totally sure why these tests exist, or why there appear to be 2 sets of identical tests there:
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_server_spec.rb#L1019-L1057
vs, say, https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_server_spec.rb#L1071-L1086

or what the use case is for having `ssl_port` == `listen_port` when `ssl` is _not_true, but I think this PR might make some sense.

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -354,6 +354,12 @@ define nginx::resource::server (
     }
   }
 
+  # Try to error in the case where the user sets ssl_port == listen_port but
+  # doesn't set ssl = true
+  if (!($ssl == true) and ($ssl_port == $listen_port)) {
+    warning('nginx: ssl must be true if listen_port is the same as ssl_port')
+  }
+
   concat { $config_file:
     owner   => $owner,
     group   => $group,
--- a/spec/defines/resource_server_spec.rb
+++ b/spec/defines/resource_server_spec.rb
@@ -990,14 +990,16 @@ describe 'nginx::resource::server' do
         it { is_expected.to contain_file('/etc/nginx/uwsgi_params').with_mode('0644') }
       end
 
-      context 'when listen_port == ssl_port' do
+      context 'when listen_port == ssl_port but ssl = false' do
         let :params do
           default_params.merge(listen_port: 80,
-                               ssl_port: 80)
+                               ssl_port: 80,
+                               ssl: false)
         end
 
-        it { is_expected.not_to contain_concat__fragment("#{title}-header") }
-        it { is_expected.not_to contain_concat__fragment("#{title}-footer") }
+        # TODO: implement test after this can be tested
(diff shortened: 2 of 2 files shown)


manifests/resource/server.pp
```diff
@@ -615,6 +615,12 @@
     }
   }
 
+  # Try to error in the case where the user sets ssl_port == listen_port but
+  # doesn't set ssl = true
+  if (!($ssl == true) and ($ssl_port == $listen_port)) {
+    fail('nginx: ssl must be true if listen_port is the same as ssl_port')
```
xaque208: I'd support a warning here, but a failure seems over the top.

oranenj:
Is this the correct fix? Setting $ssl_port to $listen_port and $ssl = true sounds like valid configuration, so perhaps the module behaviour should be fixed instead?

oranenj:
I think this might be at least tangentially related to #1029 as well?

wyardley:
@oranenj I am not sure if this should fail either, but, just to be clear, it's !($ssl = true), not $ssl = true.

wyardley:
I updated it to just warn, and rebased. Still need to look into what the rspec syntax is to test for a warning, and fix the test

vinzent:
maybe @oranenj  can review this again?

manifests/resource/server.pp
wyardley: Yeah, I agree, it's now a warning.

wyardley:
Do folks think this is a reasonable way to handle the inability to test warnings in this context for now? Basically, I left the test there as a pending test, and the message string is still in there for reference.

oranenj:
@wyardley: I have no idea how to test for warnings, unfortunately. All I know of rspec is what I've seen in other specs.

Anyway, the code I see still has a fail() call. It might also be possible to enforce $ssl = true (maybe with a warning) if $ssl_port is set. Would that be surprising behaviour?

manifests/resource/server.pp
oranenj: The failure is still here :/

wyardley:
@oranenj D'oh, sorry about that. Updated. Take a look.

_It might also be possible to enforce $ssl = true (maybe with a warning) if $ssl_port is set. Would that be surprising behaviour?_

I think there are valid use cases for this; it's only when `ssl_port` and `listen_port` are the same that it's an issue, as there are valid configs that have both SSL and non-SSL in the same server resource, if that makes sense.

`ssl_port` defaults to 443, so it doesn't have to be set unless you're using a non-default port. But I think you could have a single resource with listen_port=80 and ssl_port=4443 and that would be valid.

oranenj:
Yeah, these changes look fine to me.

PR #1023, opened 2017-02-21
Validate $proxy_headers_hash_bucket_size to be integer, not string

sgaduuw:
This is same behaviour as the validation of $names_hash_bucket_size. This way the bucket behavior stays the same with data from hiera.


xaque208:
The current master version only supports Puppet4.  I think it'd be better to use the native type validation to achieve your desires here.

wyardley:
Master now has:
```
  Integer $proxy_headers_hash_bucket_size                                                       = 64,
```
So I'm thinking this is probably sufficient and achieves the same goal?

vinzent:
obsoleted by change to puppet4 data types. Just reopen of you feel this isn't the case.

PR #1025, opened 2017-02-21
Part of the README regarding Hiera is not up to date

ryayon:


wyardley:
I think this works as intended; it is usage that's not as well documented as perhaps it should be, but I think the docs are as intended.
See the code in
 https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/init.pp#L285-L291

ryayon:
Indeed.

PR #1030, opened 2017-03-06
allow service and package to be called outside of main class

cam-itv:
Hi, this allows the service and package classes to be called outside the main class. This reverts part of the change made in https://github.com/voxpupuli/puppet-nginx/pull/950 which disallowed calling any other sub classes. This does not break existing functionality.

Its use case is to allow installing the nginx package via the module, without doing any of the other configuration. This is pretty useful in a wrapper module, which installs the rest of the configuration via templates instead.
```
    include nginx::params

    # install nginx package only
    class { 'nginx::package' : }
```

I think in the future having a feature flag `install_only` would be pretty handy too.

Thanks

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -22,8 +22,6 @@ class nginx::package(
   $manage_repo              = $::nginx::params::manage_repo,
 ) {
 
-  assert_private()
-
   anchor { 'nginx::package::begin': }
   anchor { 'nginx::package::end': }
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @cam-itv, are you able to provide a PR that adds the install_only param instead? This would match the current coding best practices.

juniorsysadmin:
Perhaps a better idea is parameter ```manage_config``` defaulting to ```true```

wyardley:
I like the idea of this, however, I think someone will probably just need to rework this from scratch. OP: if you want to update it, or want this reopened, please comment here or reopen.

PR #1031, opened 2017-03-07
replace validate_* with datatypes

bastelfreak:

### DIFF
--- a/manifests/resource/map.pp
+++ b/manifests/resource/map.pp
@@ -63,22 +63,14 @@
 
 
 define nginx::resource::map (
-  $string,
-  $mappings,
-  $default    = undef,
-  $ensure     = 'present',
-  $hostnames  = false
+  String $string,
+  Variant[Array, Hash] $mappings,
+  Optional[String] $default         = undef,
+  Enum['absent', 'present'] $ensure = 'present',
+  Boolean $hostnames                = false
 ) {
-  validate_string($string)
   validate_re($string, '^.{2,}$',
     "Invalid string value [${string}]. Expected a minimum of 2 characters.")
-  if ! ( is_array($mappings) or is_hash($mappings) ) {
-    fail("\$mappings must be a hash of the form { 'foo' => 'pool_b' } or array of hashes of form [{ 'key' => 'foo', 'value' => 'pool_b' }, ...]")
-  }
-  validate_bool($hostnames)
-  validate_re($ensure, '^(present|absent)$',
-    "Invalid ensure value '${ensure}'. Expected 'present' or 'absent'")
-  if ($default != undef) { validate_string($default) }
 
   $root_group = $::nginx::root_group
   $conf_dir   = "${::nginx::conf_dir}/conf.d"
--- a/manifests/resource/streamhost.pp
+++ b/manifests/resource/streamhost.pp
@@ -47,59 +47,24 @@
 #    ensure   => present,
 #  }
 define nginx::resource::streamhost (
-  $ensure                       = 'present',
-  $listen_ip                    = '*',
-  $listen_port                  = 80,
-  $listen_options               = undef,
-  $ipv6_enable                  = false,
-  $ipv6_listen_ip               = '::',
-  $ipv6_listen_port             = 80,
-  $ipv6_listen_options          = 'default ipv6only=on',
-  $proxy                        = undef,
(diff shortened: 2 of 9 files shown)


manifests/resource/upstream/member.pp
```diff
@@ -37,26 +37,11 @@
...
-    "${ensure} is not supported for ensure. Allowed values are 'present' and 'absent'.")
-
-  if is_string($port) {
-    warning('DEPRECATION: String $port must be converted to an integer. Integer string support will be removed in a future release.')
-  }
-  elsif !is_integer($port) {
-    fail('$port must be an integer.')
-  }
-
-  $ensure_real = $ensure ? {
```
alexjfisher: Not sure $ensure is used at all at the moment.  Use of `$ensure_real` was removed in https://github.com/voxpupuli/puppet-nginx/commit/e40ce7cea63f03c1a4c0076a773739a773f866cb ??

manifests/resource/upstream/member.pp
```diff
@@ -37,26 +37,11 @@
 define nginx::resource::upstream::member (
   $upstream,
   $server,
-  $ensure                 = 'present',
-  $port                   = 80,
+  Enum['present', 'absent'] $ensure = 'present',
```
alexjfisher: Since this parameter isn't used anymore, maybe default it to `undef` instead?

Then
```puppet
if $ensure { warning('DEPRECATION: some useful message here') }
```

bastelfreak:
reminder for @bastelfreak: drop the parameter in a separate PR to get it into the changelog.

manifests/resource/upstream/member.pp
bastelfreak: I would prefer it to directly drop it in the next release, since this has to be a major one we can get rid of a lot of stuff.

bastelfreak:
This should be enough for the first run. I don't want to create a bigger diff. Let me know if it is already too big, I could cherry-pick the changes in each file to own PRs.

alexjfisher:
Just picking out the breaking change(s) into their own PR(s) works for me.

bastelfreak:
we should wait with merging until I opened PRs for all breaking changes

manifests/resource/map.pp
```diff
@@ -63,22 +63,14 @@
...
+  String $string,
+  Variant[Array, Hash] $mappings,
+  Optional[String] $default         = undef,
+  Enum['absent', 'present'] $ensure = 'present',
+  Boolean $hostnames                = false
 ) {
-  validate_string($string)
   validate_re($string, '^.{2,}$',
     "Invalid string value [${string}]. Expected a minimum of 2 characters.")
-  if ! ( is_array($mappings) or is_hash($mappings) ) {
```
binford2k: this one might be useful to keep around? The error message is a little more explanatory than the default.

manifests/resource/streamhost.pp
```diff
@@ -47,60 +47,25 @@
...
-  $mode                         = $::nginx::global_mode,
+  Enum['absent', 'present'] $ensure       = 'present',
+  Variant[Array, String] $listen_ip       = '*',
+  Integer $listen_port                    = 80,
+  Optional[String] $listen_options        = undef,
+  Boolean $ipv6_enable                    = false,
+  Variant[Array, String] $ipv6_listen_ip  = '::',
+  Integer $ipv6_listen_port               = 80,
+  String $ipv6_listen_options             = 'default ipv6only=on',
+  $proxy                                  = undef,
```
binford2k: do we have data types for these handful of un-decorated parameters?

bastelfreak: yes. But I want to add them in another PR. I want to keep the focus here on replacing validate functions. and in the next step adding types for all parameters + maybe improve existing ones.

manifests/resource/map.pp
bastelfreak: as discussed on IRC: we can kep the fail() call but this will never be reached. I wanted to move it as comments to the header documentation, but I jut noticed that it is already present.

manifests/resource/upstream/member.pp
bastelfreak: as mentioned in the other comment. We will drop the param. We can safely do this since the next release has to be a major one.

PR #1050, opened 2017-03-10
replace validate_* calls with datatypes in location.pp

bastelfreak:


manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
-  $location                    = $name,
-  $server                      = undef,
-  $www_root                    = undef,
-  $autoindex                   = undef,
-  $index_files                 = [
+  String $location                    = $name,
+  String $server                      = undef,
+  Optional[String] $www_root                    = undef,
+  Optional[String] $autoindex                   = undef,
+  Optional[Array ]$index_files                 = [
```
vinzent: unnecessary whitespace

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
 #  }
 
 define nginx::resource::location (
-  $ensure                      = present,
+  Enum['present', 'absent'] $ensure                      = present,
```
vinzent: unnecessary whitespace

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
-  $location_allow              = undef,
-  $location_deny               = undef,
-  $option                      = undef,
-  $stub_status                 = undef,
-  $raw_prepend                 = undef,
-  $raw_append                  = undef,
-  $location_custom_cfg         = undef,
-  $location_cfg_prepend        = undef,
-  $location_cfg_append         = undef,
+  Optional[String] $proxy                       = undef,
```
vinzent: unnecessary whitespace (for all param lines))

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
+  Optional[String] $proxy_http_version          = undef,
+  Optional[String] $proxy_set_body              = undef,
+  Optional[Enum['on', 'off']] $proxy_buffering             = undef,
+  Optional[String] $auth_basic                  = undef,
+  Optional[String] $auth_basic_user_file        = undef,
+  Optional[String] $auth_request                = undef,
+  Array $rewrite_rules               = [],
+  Integer $priority                    = 500,
   $mp4                         = false,
   $flv                         = false,
```
vinzent: missing type

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
 #  }
 
 define nginx::resource::location (
-  $ensure                      = present,
+  Enum['present', 'absent'] $ensure                      = present,
   $internal                    = false,
```
vinzent: missing type

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
+  Optional[Enum['any', 'all']] $location_satisfy            = undef,
+  Optional[Array] $location_allow              = undef,
+  Optional[Array] $location_deny               = undef,
+  Optional[Boolean ] $stub_status                 = undef,
+  Optional[Variant[String, Array]] $raw_prepend                 = undef,
+  Optional[Variant[String, Array]] $raw_append                  = undef,
+  Optional[Hash] $location_custom_cfg         = undef,
+  Optional[Hash] $location_cfg_prepend        = undef,
+  Optional[Hash] $location_cfg_append         = undef,
   $location_custom_cfg_prepend = undef,
```
vinzent: missing type

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
   $location_custom_cfg_append  = undef,
```
vinzent: missing type

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
+  Optional[String] $proxy_method                = undef,
+  Optional[String] $proxy_http_version          = undef,
+  Optional[String] $proxy_set_body              = undef,
+  Optional[Enum['on', 'off']] $proxy_buffering             = undef,
+  Optional[String] $auth_basic                  = undef,
+  Optional[String] $auth_basic_user_file        = undef,
+  Optional[String] $auth_request                = undef,
+  Array $rewrite_rules               = [],
+  Integer $priority                    = 500,
   $mp4                         = false,
```
vinzent: missing type

manifests/resource/location.pp
```diff
@@ -224,167 +223,15 @@
...
-    validate_string($auth_basic_user_file)
-  }
-  if ($auth_request != undef) {
-    validate_string($auth_request)
-  }
-  if !is_integer($priority) {
-    fail('$priority must be an integer.')
-  }
-  validate_array($rewrite_rules)
   if (($priority + 0) < 401) or (($priority + 0) > 599) {
```
vinzent: the integer data types accepts min max params: https://docs.puppet.com/puppet/4.9/lang_data_number.html#the-integer-data-type

this conditional can be directly specified with the param type.

manifests/resource/location.pp
bastelfreak: woah shit. somehow forogt to reformat. sorry!

spec/defines/resource_server_spec.rb
```diff
@@ -904,18 +904,6 @@
         it { is_expected.to contain_nginx__resource__location("#{title}-default").with_ssl_only(true) }
       end
 
-      context 'SSL cert missing' do
```
vinzent: what is legacy about the tests?

vinzent:
did you run the beaker acceptance tests? are the ok?

spec/defines/resource_server_spec.rb
bastelfreak: the datatypes enforce the correct param, so we don't need to test wrong/missing parameter. We dropped these kinds of tests on all other puppet4 only modules as well.

bastelfreak:
mhm beaker... https://gist.github.com/bastelfreak/947477bff061607dc797119189a008ac
it is somehow broken. I will take a look at this in the next days, but maybe needs some time. if anybody else wants to test this... feel free :)

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
+  Optional[String] $proxy_cache_use_stale             = undef,
+  Optional[Variant[Array, String]] $proxy_cache_valid = undef,
+  Optional[String] $proxy_method                      = undef,
+  Optional[String] $proxy_http_version                = undef,
+  Optional[String] $proxy_set_body                    = undef,
+  Optional[Enum['on', 'off']] $proxy_buffering        = undef,
+  Optional[String] $auth_basic                        = undef,
+  Optional[String] $auth_basic_user_file              = undef,
+  Optional[String] $auth_request                      = undef,
+  Array $rewrite_rules                                = [],
```
igalic: how come this isn't `Optional`?

spec/defines/resource_server_spec.rb
alexjfisher: Dropping the tests is easier than fixing them, but what's testing that you've correctly chosen the right data types when refactoring?

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
-  $internal                    = false,
-  $location                    = $name,
-  $server                      = undef,
-  $www_root                    = undef,
-  $autoindex                   = undef,
-  $index_files                 = [
+  Enum['present', 'absent'] $ensure                   = present,
+  Boolean $internal                                   = false,
+  String $location                                    = $name,
+  String $server                                      = undef,
```
alexjfisher: Does defaulting to `undef` make sense if you're not using `Optional[String]` here?

manifests/resource/location.pp
bastelfreak: it was already an existing array. I didn't want to change the datatype to undef and make it optional.

spec/defines/resource_server_spec.rb
bastelfreak: that should be handled by the reviewer + the existing tests. They will fail when wrong datatypes are supplied.

manifests/resource/location.pp
bastelfreak: yes, that makes it a required parameter. It currently already is a required param, just not handled nicely:
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp#L385-L387
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp#L230-L232

It would have been more clear if it didn't default to undef but to nothing.

alexjfisher: Yeah, old code does look a bit funky.  Do you want to remove the default then since now it is much more obvious it provides no value.

bastelfreak: I'm not sure what the best practice is. in the other PRs we tried to always default to undef on required params without default value

spec/defines/resource_server_spec.rb
alexjfisher: *if* there are other tests.  I don't have massively strong feelings though.  It was more devil's advocate.  They probably are more trouble than they're worth.

bastelfreak: yes. In general we need more tests. but this is more of a global issue on all of our modules.

manifests/resource/location.pp
```diff
@@ -153,66 +153,65 @@
...
+  Optional[Array] $location_allow                     = undef,
+  Optional[Array] $location_deny                      = undef,
+  Optional[Boolean ] $stub_status                     = undef,
+  Optional[Variant[String, Array]] $raw_prepend       = undef,
+  Optional[Variant[String, Array]] $raw_append        = undef,
+  Optional[Hash] $location_custom_cfg                 = undef,
+  Optional[Hash] $location_cfg_prepend                = undef,
+  Optional[Hash] $location_cfg_append                 = undef,
+  Optional[Hash] $location_custom_cfg_prepend         = undef,
+  Optional[Hash] $location_custom_cfg_append          = undef,
```
dominics: Should probably have been `Optional[Variant[Hash, Array]]` to match existing behaviour (#1100)

alexjfisher: Possibly, but maybe not. See https://github.com/voxpupuli/puppet-nginx/issues/1100#issuecomment-319899019

PR #1057, opened 2017-03-27
replace validate_* calles with datatypes in server.pp

bastelfreak:

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -197,381 +197,120 @@
 #    ssl_key  => '/tmp/server.pem',
 #  }
 define nginx::resource::server (
-  $ensure                       = 'present',
-  $listen_ip                    = '*',
-  $listen_port                  = 80,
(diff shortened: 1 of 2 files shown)


manifests/resource/server.pp
```diff
@@ -197,382 +197,121 @@
...
+  Optional[String] $log_by_lua                                                   = undef,
+  Optional[String] $log_by_lua_file                                              = undef,
+  $use_default_location                                                          = true,
+  $rewrite_rules                                                                 = [],
+  $string_mappings                                                               = {},
+  $geo_mappings                                                                  = {},
+  Optional[String] $gzip_types                                                   = undef,
+  String $owner                                                                  = $::nginx::global_owner,
+  String $group                                                                  = $::nginx::global_group,
+  String $mode                                                                   = $::nginx::global_mode,
```
alexjfisher: Shouldn't this be a Regexp?

PR #1058, opened 2017-03-28
Use some more puppet 4 features to reduce code

igalic:
and increase readability!

* Use contain instead of include + anchor
* Use type alias in place over overly long inline Type Declaration


manifests/init.pp
```diff
@@ -172,8 +172,8 @@
     manage_repo              => $manage_repo,
   }
```
alexjfisher: `contain '::nginx::package'` needed too?

igalic: well, no, as you can see, it's instantiated in resource format for reasons not entirely clear to me…

alexjfisher: @igalic You can, (and I think should), still `contain` it.

> To contain classes that are declared with the resource-like declaration syntax, use the contain function after declaring the class
https://docs.puppet.com/puppet/4.9/lang_containment.html

bastelfreak:
@igalic can you check the failed tests? We can probably just drop them.

igalic:
funnily enough, the failed tests were because of the removed anchors.
i'll look into incorporating @alexjfisher's idea to use `contain` for `nginx::package`, too.

wyardley:
+1 for this (even if we have to adjust / kills some tests). This should make things a lot more readable.

Agree on the 'fixme' bits, that pattern is really redundant and should probably have a function that handles it in a generic way. If someone comes up with the base pattern, I could probably help converting some of the templates.

bastelfreak:
@igalic thanks!

PR #1061, opened 2017-03-31
Choose the right path for stream members

cova-fe:
It seems that the configuration files are not written in the right place
when streams are used. So use the same logic of upstream.pp to determine
the right path.

### DIFF
--- a/manifests/resource/upstream/member.pp
+++ b/manifests/resource/upstream/member.pp
@@ -14,6 +14,7 @@
 #   [*server*]                  - Hostname or IP of the upstream member server
 #   [*port*]                    - Port of the listening service on the upstream member
 #   [*upstream_fail_timeout*]   - Set the fail_timeout for the upstream. Default is 10 seconds
+#   [*upstream_context*]        - Sets the context for upstream: 'http' (default) or 'stream'
 #
 #
 # Examples:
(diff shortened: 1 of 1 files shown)


manifests/resource/upstream/member.pp
```diff
@@ -40,11 +42,17 @@
...
 
+  $conf_dir_real = $upstream_context ? {
+    'stream' => 'conf.stream.d',
+    default  => 'conf.d',
+  }
+
   # Uses: $server, $port, $upstream_fail_timeout
   concat::fragment { "${upstream}_upstream_member_${name}":
-    target  => "${::nginx::conf_dir}/conf.d/${upstream}-upstream.conf",
+    target  => "${::nginx::conf_dir}/${conf_dir_real}/${upstream}-upstream.conf",
```
igalic: i think it's worth adding an rspec test for this

cova-fe: Fair enough, will do asap

wyardley: @cova-fe is this something you're still able to work on?

cova-fe: Yes, I should be able to do it :)

bastelfreak:
Ping @cova-fe. Can you please add tests for this?

bastelfreak:
Hi. I'm going to close this PR due to inactivity. Please reopen if you're still interested.

PR #1066, opened 2017-04-06
Avoid spurious location block when redirecting to SSL in another server block

### LINKED ISSUE #1029: nginx 0.6.0: bad location block causes nginx restart to fail

- Puppet: 4.5.3  (Puppet Enterprise 2016.2)
- Ruby: ruby 2.1.9p490
- Distribution: CentOS 7
- Module version: puppet-nginx 0.6.0

Trying to create a default entry to redirect HTTP to HTTPS
```yaml
nginx::nginx_servers:
  '_':
    ensure: 'present'
    listen_options: 'default_server'
    listen_port: 80
    ssl: false
    ssl_redirect: true
```

## What are you seeing
Nginx service fails to restart due to `location` block outside the `server` block in the generated file:

```nginx
# MANAGED BY PUPPET
server {
  listen *:80 default_server;
  server_name           _;

  access_log            /var/log/nginx/_.access.log combined;
  error_log             /var/log/nginx/_.error.log;
}

  location / {
    index     index.html index.htm index.php;
  }
```

A `location` block inside the `server` block, or a better solution would be no location block for this redirect.

```nginx
# MANAGED BY PUPPET
server {
  listen *:80 default_server;
  server_name           _;

  access_log            /var/log/nginx/_.access.log combined;
  error_log             /var/log/nginx/_.error.log;
}
```

## Output log

```puppet
Info: /Stage[main]/Nginx/Nginx::Resource::Server[_]/Concat[/etc/nginx/conf.d/_.conf]/File[/etc/nginx/conf.d/_.conf]: Scheduling refresh of Class[Nginx::Service]
Info: Concat[/etc/nginx/conf.d/_.conf]: Scheduling refresh of Class[Nginx::Service]
Info: Nginx::Resource::Location[_-default]: Scheduling refresh of Class[Nginx::Service]
Info: Class[Nginx::Service]: Scheduling refresh of Service[nginx]
Error: /Stage[main]/Nginx::Service/Service[nginx]: Failed to call refresh: Could not restart Service[nginx]: Execution of '/bin/systemctl restart nginx' returned 1: Job for nginx.service failed because the control process exited with error code. See "systemctl status nginx.service" and "journalctl -xe" for details.
Error: /Stage[main]/Nginx::Service/Service[nginx]: Could not restart Service[nginx]: Execution of '/bin/systemctl restart nginx' returned 1: Job for nginx.service failed because the control process exited with error code. See "systemctl status nginx.service" and "journalctl -xe" for details.
```

Adding `use_default_location: false` with `ssl_redirect: true` will remove the location block. 
This should not be required with `ssl_redirect: true`

oranenj:
Fixes #1029

I'm pretty sure this is sane behaviour. Any thoughts?

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -370,7 +370,7 @@ define nginx::resource::server (
     $_ssl_redirect_port = $ssl_port
   }
 
-  # Suppress unneded stuff in non-SSL location block when certain conditions are
+  # Suppress unneeded stuff in non-SSL location block when certain conditions are
   # met.
   if (($ssl == true) and ($ssl_port == $listen_port)) or ($ssl_redirect) {
     $ssl_only = true
@@ -378,7 +378,11 @@ define nginx::resource::server (
     $ssl_only = false
   }
 
-  if $use_default_location == true {
+  # If we're redirecting to SSL, the default location block is useless, *unless*
+  # SSL is enabled for this server
+  # either       and    ssl -> true
+  # ssl redirect and no ssl -> false
+  if ($ssl_redirect != true or $ssl == true) and $use_default_location == true {
     # Create the default location reference for the server
     nginx::resource::location {"${name_sanitized}-default":
       ensure                      => $ensure,
--- a/spec/defines/resource_server_spec.rb
+++ b/spec/defines/resource_server_spec.rb
@@ -898,10 +898,28 @@ describe 'nginx::resource::server' do
         it { is_expected.to contain_concat__fragment("#{title}-header").with_content(%r{  return 301 https://\$host:9787\$request_uri;}) }
       end
 
-      context 'ssl_redirect should set ssl_only' do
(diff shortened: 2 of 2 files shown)


wyardley:
@oranenj 
Even though I updated some of this code not that long ago, this is still making my head hurt to re-read.
That said, did you look through
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp#L273-L288 too?

I think this looks sane from a quick read, though I wonder if there's a way to rework the logic to be a bit easier to read

oranenj:
@wyardley: The current logic is a bit hard to keep track of, yes, but I think the location resource works as it should. At least, I didn't spot any obvious issues with it.

Refactoring the whole thing would take more effort than I have time to spend on this at the moment. :/

wyardley:
@oranenj Yeah, no worries! From my end this looks good, but would like to see another reviewer as well.

PR #1068, opened 2017-04-10
Adds hiera nginx_mailhosts_defaults like nginx_servers_defaults

dol:
Set mailhosts default value. This is useful when setting the SSL settings on multiple ports and protocols or disable xclient.
Similar to this commit for vhost (aka servers):
https://github.com/voxpupuli/puppet-nginx/commit/ad391f67bf5297f472bf7ccdd163dc7899dbe1fd


vinzent:
thanks @dol  !

dol:
Thanks for the merge. 👍 🎉

PR #1069, opened 2017-04-11
Add access_by_lua_file server directive.

oc243:
This lets you control access to an nginx server using lua

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -184,6 +184,7 @@
 #   [*error_pages*]             - Hash: setup errors pages, hash key is the http
 #                                 code and hash value the page
 #   [*locations*]               - Hash of servers ressources used by this server
+#   [*access_by_lua_file*]      - Path of a lua file that controls server access
 # Actions:
 #
 # Requires:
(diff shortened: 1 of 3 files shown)


dol:
The [lua-nginx-module](https://github.com/openresty/lua-nginx-module#directives) provides approx. 70 directives. `access_by_lua_file` is only one of them. I suggest adding a few more or separate them into it's own resource that inherits from the server resource. This makes it easier to update the lua-nginx-module directives more independent.
The [OpenResty](https://openresty.org/) project bundles quite a few [components](https://openresty.org/en/components.html). All the non Lua based components have their own nginx directives.

At the current state non default directives can be loaded via the different cfg_prepend/append possibilities.

vinzent:
@dol defined types can't inherit in puppet. Or what you think of by inheriting?

dol:
@vinzent By inherit i meant class inheritance. But since you pointed out, thats not possible with defined type. My bad.

oc243:
@dol what do you suggest doing with this PR?

dol:
@oc243 A proper integration needs some planing. For example `access_by_lua_file` could be defined in the following context: `http, server, location, location if`. This would imply to adding access_by_lua_block to different resources. In addition there is the related 'access_by_lua` and `access_by_lua_block` directive, that needs to be offered as well.

Regarding this PR I would close this PR without merging and would transfer the findings and comment into it's own issue/feature request.

PR #1070, opened 2017-04-11
Fail defined types if nginx class was not declared before

### LINKED ISSUE #983: Please add a 'warn' when someone is using 'nginx::resource::vhost' without previously including the nginx class

- Puppet: 4.x
- Ruby: 2.3.x
- Distribution: Ubuntu
- Module version: 0.5.0

Given Hiera data like: 
```
role::web::myhost_nginx_vhosts:
  'somehost.mycorp.lan':
    www_root: '/var/www'
    listen_options: 'default_server'
    server_name: ["%{networking.fqdn}"]
    access_log: '/var/log/nginx/myhost-access.log'
    error_log: '/var/log/nginx/myhost-error.log'
```

and instantiating it in a far-flung role with
```
class role::web ($myhost_nginx_vhosts) {
  create_resources('nginx::resource::vhost', $myhost_nginx_vhosts)
} 

```
there will be failures along the lines of: 
`Evaluation Error: Error while evaluating a Function Call, nil is not an Array.  It looks to be a NilClass at /etc/puppetlabs/code/environments/vagrant/modules/nginx/manifests/resource/vhost.pp:407:3 at /etc/puppetlabs/code/environments/vagrant/modules/role/manifests/web.pp:1 on node blasrgh
`
  This provides few hints as to the root cause of the error, which in this case is our hapless hero forgot to put: 
`class { 'nginx': }`

in the role, kicking off three days of hair-pulling debugging, swearing, and kicking things.  Total PEBKAC, yes, but having
`unless defined('class nginx { warn('You might want to consider including the nginx class if you are getting strange evaluation errors') }` 

would've been MOST helpful.  
## What are you seeing
Red.  Mainly from my own foibles.  

I expected the module to compile without complaint, but I would've settled for a warning informing me the 'nginx' parent class was not included when I tried to use the vhost defined-type.

vinzent:
Because the `ngnix::resource:*` types access ::ngnix class
parameters the nginx class needs to be declared before calling
the defined type.

`include ::nginx` inside the defined type is not enough because
params are evaluated before include is parsed.

Two of the defined types only access params inside. The include
way would work there - but for consistency I also added the fail.

Closes #983

### DIFF
--- a/manifests/resource/geo.pp
+++ b/manifests/resource/geo.pp
@@ -63,6 +63,10 @@ define nginx::resource::geo (
   Optional[Boolean] $proxy_recursive  = undef
 ) {
 
+  if ! defined(Class['nginx']) {
+    fail('You must include the nginx base class before using any defined resources')
+  }
+
   $root_group = $::nginx::root_group
   $conf_dir   = "${::nginx::conf_dir}/conf.d"
 
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -214,6 +214,10 @@ define nginx::resource::location (
   Optional[String] $expires                           = undef,
 ) {
 
+  if ! defined(Class['nginx']) {
+    fail('You must include the nginx base class before using any defined resources')
+  }
+
   $root_group = $::nginx::root_group
 
   File {
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -118,6 +118,10 @@ define nginx::resource::mailhost (
   Array $server_name                             = [$name]
 ) {
 
+  if ! defined(Class['nginx']) {
+    fail('You must include the nginx base class before using any defined resources')
+  }
+
   $root_group = $::nginx::root_group
 
   File {
--- a/manifests/resource/map.pp
+++ b/manifests/resource/map.pp
@@ -69,6 +69,10 @@ define nginx::resource::map (
   Enum['absent', 'present'] $ensure = 'present',
   Boolean $hostnames                = false
 ) {
+  if ! defined(Class['nginx']) {
+    fail('You must include the nginx base class before using any defined resources')
+  }
+
   validate_re($string, '^.{2,}$',
     "Invalid string value [${string}]. Expected a minimum of 2 characters.")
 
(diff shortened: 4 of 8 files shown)


vinzent:
@law @timatooth @triforce you might be interested in reviewing because of your interest in #983

vinzent:
probably needs proper tests? :-)

manifests/resource/geo.pp
```diff
@@ -63,6 +63,10 @@
   Optional[Boolean] $proxy_recursive  = undef
 ) {
 
+  if ! defined(Class['nginx']) {
```
wyardley: I think this is fine and probably goes with the style of the module as a whole... that said, I think some people would say that 'unless' is more idiomatic, and probably lines up better with https://docs.puppet.com/puppet/4.9/lang_conditional.html ? I like 'if not' better than 'if !' but IIRC, it may not work in the DSL?

vinzent:
@wyardley  this isn't related at all with puppet4. The very same parse order problem exists with puppet since forever. This fail if not defined behaviour here is the very same as puppetlabs/apache uses for apache::vhost

manifests/resource/geo.pp
vinzent: If or unless is IMHO very opinionated. I for myself prefer if over unless in most situations. But I'll change to unless if requested.

vinzent:
@wyardley I havent checked if any of the defined types are just private-style types. If there sre some it might make sense to make to mark them private in addition. In addition to even ensure with internal use the class gets included before.

manifests/resource/geo.pp
wyardley: I mostly just don't like `if !` vs `if not` but I don't have a strong opinion, and it's used all throughout the code, so I don't see any reason to change it.

wyardley:
_this isn't related at all with puppet4_

Sorry, I just meant that there might be a different way to test for this with some of the features that are being added now that we're dropping Puppet 3 support.

_I havent checked if any of the defined types are just private-style types._

I think most of these are intended to be called directly via
```
nginx::resource::foo { 'something':
```
as described in the docs, so something along these lines is probably the right approach.

wyardley:
@vinzent gentle bump

wyardley:
Ran this by @bastelfreak and decided we should merge this in even without tests

PR #1071, opened 2017-04-11
Add location defaults to init and server resource

dol:
By adding location defaults it's possible to overwrite location
value like 'index_files'.

Related to #1068

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -154,6 +154,7 @@ class nginx (
   $geo_mappings                                     = {},
   $string_mappings                                  = {},
   $nginx_locations                                  = {},
+  $nginx_locations_defaults                         = {},
   $nginx_mailhosts                                  = {},
   $nginx_mailhosts_defaults                         = {},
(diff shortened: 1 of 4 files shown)


dol:
My goal is to avoid [config_pitfalls](https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/). I'd like to clear the `www_root` and `index_files` when using locations in the context of `nginx::nginx_servers`(hiera).

elmobp:
+1!

wyardley:
This seems sane to me (though I'd love to see someone take a stab at reworking the configs in a way that avoids so much duplication in a cleaner way).

@dol can you rebase?

wyardley:
@dol I think the extra commit ends up pulling in other changes, so if possible, can you reset back, then rebase, then force push?

dol:
@wyardley It took some time to get what you were asking for. But the merge commit is cleaned up.

oranenj:
Looks good to me. Thanks :)

PR #1062, opened 2017-04-04
Add fastcgi index

elmobp:
Add the fastcgi index to the server resource

PR #1073, opened 2017-04-12
Adding FastCGI index

elmobp:
This should do the trick!


wyardley:
(continued from #1062)

Issue #1074, opened 2017-04-12
'Cannot create a location reference without' rahter anoying and blocks some possibilities


dol:
The following example is not possible due to the current check of the locations paramters in https://github.com/voxpupuli/puppet-nginx/blob/0e56fa0f7bc09e7c034813556ad82fcfecd9b339/manifests/resource/location.pp#L233

```
nginx::nginx_servers:
    myserver:
        server_name:
            - 'foobar.example.com'
        listen_port: 80
        www_root: /var/www/web
        use_default_location: false
        locations:
            foobar_deny_root:
                www_root: ~
                index_files: []
                location: /
                location_custom_cfg:
                    return: 301 $scheme://$server_name/testfile
            testfile:
                www_root: ~
                index_files: []
                location: '= /testfile'
```

This would generate a config file like this.
```
# MANAGED BY PUPPET
server {
  listen *:80;
  server_name           foobar.example.com;

  root /var/www/web;
  access_log            /var/log/nginx/myserver.access.log main;
  error_log             /var/log/nginx/myserver.error.log;

  location / {
    return 301 $scheme://$server_name/testfile;
  }

  location = /testfile {
  }
}
```

But this is not possible, because the following error occurs:
```
Cannot create a location reference without a www_root, proxy, location_alias, stub_status, fastcgi, uwsgi, location_custom_cfg, internal, try_files, location_allow, or location_deny
```

I understand that this check was needed in combination with `use_default_location` to prevent the misconfiguration of the default location.

My suggestion to this problem:
Remove the restriction in the location.pp file and move the check to the [server.pp file](https://github.com/voxpupuli/puppet-nginx/blob/f36527afcfae9cb53bff1d61c424c50062ab4204/manifests/resource/server.pp#L321).
Apply the check where the context switch is made.

dol:
This check was introduced 6 years ago and was since then altered from this commit: https://github.com/voxpupuli/puppet-nginx/commit/ebd481d57004e0146e13c40e60b51798224f7b1f

I don't see any usefulness having this check in the location part. Does @jfryman or someone else know the origin of this check.
Looking into the history I see the following pattern:
The check was 6 years ago introduced. Since then a random number of additional parameters where added to the list by various people. Most of them where like 'Add parameter xyz to check' like https://github.com/voxpupuli/puppet-nginx/commit/d83822810ead08db899dbaa3a0c0a7e1f8fd40ef 

[...]

PR #1075, opened 2017-04-12
Remove location check of some random values to be set in the context of location

dol:
Relates to https://github.com/voxpupuli/puppet-nginx/issues/1074


wyardley:
This has been discussed some in the past. Personally I'm basically for this -- give people enough rope to hang themselves (though it would be awesome if we could stage the configs, and verify config syntax and have a configurable option to fail if there's an error).

Issue #862, opened 2016-08-30
RFC: Upstream vs module-shipped configurations


3flex:
Related PRs: #777, #855, #499, #500, #682.

This module currently ships two configuration files that override upstream defaults, fastcgi_params and uwsgi_params. There are a couple of differences:

Module fastcgi_params differs from upstream fastcgi_params as follows:

```
- fastcgi_param  REQUEST_SCHEME     $scheme;
- fastcgi_param  HTTPS              $https if_not_empty;
+ fastcgi_param  HTTPS              $https;
+ fastcgi_param  SCRIPT_FILENAME    $request_filename;
+ # Mitigate httpoxy, see https://httpoxy.org/#fix-now
+ fastcgi_param  HTTP_PROXY          "";
```

Module uwsgi_params differs from upstream:

```
- uwsgi_param  REQUEST_SCHEME     $scheme;
- uwsgi_param  HTTPS              $https if_not_empty;
+ uwsgi_param  SERVER_ADDR        $server_addr;
```

The "modern" fastcgi.conf changes `SCRIPT_FILENAME` declaration to `fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;` line as well which this module doesn't ship with.

As a matter of policy, should this module take upstream's config as is, or should it provide its own? The problem is upstream's config includes new-ish parameters like `if_not_empty` (added in 1.1.11) which if added to the module may cause configuration errors on machines where older versions are being run.

This may (will?) be a breaking change for some users.

bastelfreak:
My idea: A new release that announces the breaking change in the next major release. Then a new major release that ships the upstream configuration. We should be able to implement a warning somehow if the person uses an older nginx version? There is already a fact that returns  the nginx version which we could use here.

3flex:
I'm not sure what the behaviour is for brand new installs - is the fact available on the initial Puppet run? My understanding is Facter runs prior to Puppet, so on a first run the fact wouldn't be available and the warning wouldn't trigger. Could be wrong.

But yes, a release announcing the upcoming breaking change would be welcome. There are other breaking changes that could be made (e.g. #863) at the same time.

igalic:
facter runs before puppet, but if your new module rolls out a new fact that you rely on, you have a bootstrapping issue

wyardley:
[...]

Issue #1063, opened 2017-04-05
fastcgi_params file when set to non-default path if File resource not declared


wyardley:
I can try to fix this myself, but would appreciate some thoughts / suggestion (or feel free to submit a PR if someone else has a good fix).

Basically, if you set the fastcgi_params path to something non-default, Puppet still tries to manage it if the File resource doesn't exist. 
The current logic is:
```
if $ensure == present and $fastcgi != undef and !defined(File[$fastcgi_params])
```
which mostly deals with the case where the user creates the fastcgi_params file except in the case where
- the file is created by something outside of Puppet or
- when concat is used to build the file (this is the problem I bumped up against)

```
  concat { $fastcgi_params_xyz:
    ensure => present,
    owner  => 'root',
    group  => 'root',
    mode   => '0644',
    notify => Service['nginx'],
  }
```
I think there's a param that was supposed to create a file resource, but this also didn't seem to work.

I can't think of a scenario where we'd want to use a non-default filename but manage the file with the template in the module; it would be useful to be able to use a custom template, possibly, or just to let the user define a filename that's managed elsewhere in Puppet.

https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp#L254-L259

PR #1076, opened 2017-04-13
Use nginx defaults for fastcgi_params / uwsgi_params

wyardley:
Per various discussions (#862), use nginx shipped defaults for fastcgi_params / uwsgi_params. 

Obsoletes #855, etc.

I think long-term, we still may need a way to make this more customizable, but I think we have signoff in principle on at least pulling in the current vendor defaults from 1.10 / 1.11.

### DIFF
--- a/templates/server/fastcgi.conf.erb
+++ b/templates/server/fastcgi.conf.erb
@@ -1,16 +1,18 @@
 # This file managed by puppet on host <%= @fqdn %>
 
+fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;
 fastcgi_param  QUERY_STRING       $query_string;
 fastcgi_param  REQUEST_METHOD     $request_method;
 fastcgi_param  CONTENT_TYPE       $content_type;
 fastcgi_param  CONTENT_LENGTH     $content_length;
 
-fastcgi_param  SCRIPT_FILENAME    $request_filename;
 fastcgi_param  SCRIPT_NAME        $fastcgi_script_name;
 fastcgi_param  REQUEST_URI        $request_uri;
 fastcgi_param  DOCUMENT_URI       $document_uri;
 fastcgi_param  DOCUMENT_ROOT      $document_root;
 fastcgi_param  SERVER_PROTOCOL    $server_protocol;
+fastcgi_param  REQUEST_SCHEME     $scheme;
+fastcgi_param  HTTPS              $https if_not_empty;
 
 fastcgi_param  GATEWAY_INTERFACE  CGI/1.1;
 fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
@@ -21,10 +23,5 @@ fastcgi_param  SERVER_ADDR        $server_addr;
 fastcgi_param  SERVER_PORT        $server_port;
 fastcgi_param  SERVER_NAME        $server_name;
 
-fastcgi_param  HTTPS              $https;
-
 # PHP only, required if PHP was built with --enable-force-cgi-redirect
 fastcgi_param  REDIRECT_STATUS    200;
-
-# Mitigate httpoxy, see https://httpoxy.org/#fix-now
-fastcgi_param  HTTP_PROXY          "";
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -173,12 +173,12 @@ define nginx::resource::location (
   Optional[String] $fastcgi                           = undef,
   Optional[String] $fastcgi_index                     = undef,
   Optional[Hash] $fastcgi_param                       = undef,
-  String $fastcgi_params                              = "${::nginx::conf_dir}/fastcgi_params",
+  Optional[String] $fastcgi_params                    = "${::nginx::conf_dir}/fastcgi.conf",
   Optional[String] $fastcgi_script                    = undef,
   Optional[String] $fastcgi_split_path                = undef,
   Optional[String] $uwsgi                             = undef,
   Optional[Hash] $uwsgi_param                         = undef,
-  String $uwsgi_params                                = "${nginx::config::conf_dir}/uwsgi_params",
+  Optional[String] $uwsgi_params                      = "${nginx::config::conf_dir}/uwsgi_params",
   Optional[String] $uwsgi_read_timeout                = undef,
   Boolean $ssl                                        = false,
   Boolean $ssl_only                                   = false,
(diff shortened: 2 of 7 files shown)


wyardley:
@jostmart: how does this look to you, and would you be Ok with this superceding #855?

templates/server/fastcgi_params.erb
```diff
@@ -5,12 +5,13 @@ fastcgi_param  REQUEST_METHOD     $request_method;
 fastcgi_param  CONTENT_TYPE       $content_type;
 fastcgi_param  CONTENT_LENGTH     $content_length;
 
-fastcgi_param  SCRIPT_FILENAME    $request_filename;
```
wyardley: It does seem like this line doesn't exist in current upstream, though there is `SCRIPT_NAME`.

wyardley:
As noted in comment on the relevant text, this doesn't seem to be in current upstream:
`fastcgi_param  SCRIPT_FILENAME    $request_filename; `
so I guess that would make it breaking (though hopefully in a minor way).

templates/server/fastcgi_params.erb
3flex: Upstream has two files, fastcgi.conf and fastcgi_params. As written the PR syncs with fastcgi_params which doesn't set `SCRIPT_FILENAME`. fastcgi.conf is more modern and does set it (that's the only difference between them).

I suggest syncing to fastcgi.conf, and renaming this template to `fastcgi.conf.erb`... that's likely to have the least impact on users while also being explicit about which upstream fastcgi config file it's trying to replicate.

Some background here if you like: https://blog.martinfjordvald.com/2013/04/nginx-config-history-fastcgi_params-versus-fastcgi-conf/

wyardley: Thanks, that's very helpful context, so I'm glad I bugged you about this one. And here I had hoped this was going to be simple and uncontroversial.

I agree with you about this, do we just leave an existing `/etc/nginx/fastcgi_params` alone if it exists, though? I guess if the module doesn't include it, it's not an issue, and maybe safer than assuming we managed it in the past? Should we rename the parameter itself from `fastcgi_param`?

wyardley:
@3flex: pushed a new version with those changes and updated tests.
Also added a fix for #1063, and a test for that case.

templates/server/fastcgi_params.erb
3flex: Yeah I think it should be left alone.

I'd like the parameter to be renamed, but I don't think it's worth it...

3flex:
With this change in, I think it makes sense to default to fastcgi.conf shipped in this module only and leave fastcgi_params alone (if it exists). And allowing users to remove the config would also be useful.

wyardley:
_And allowing users to remove the config would also be useful._

Ok, this version makes it optional, and adds a few more tests that I _think_ make sense.

wyardley:
_Edit: Probably should have checked the test results and spotted that merge conflict before clicking this :man_shrugging: but besides that, looks great!_

@3flex I think the conflict came from another recent change that got merged. I don't know how I missed the rubocop failure. Fixed all that and another minor issue with a test.

oranenj:
Seems fine to me.

templates/server/fastcgi.conf.erb
```diff
@@ -21,10 +23,5 @@ fastcgi_param  SERVER_ADDR        $server_addr;
 fastcgi_param  SERVER_PORT        $server_port;
 fastcgi_param  SERVER_NAME        $server_name;
 
-fastcgi_param  HTTPS              $https;
-
 # PHP only, required if PHP was built with --enable-force-cgi-redirect
 fastcgi_param  REDIRECT_STATUS    200;
-
```
vchepkov: Curious, why this block was removed?
Seems like a good idea to keep

wyardley: @vchepkov The idea is to synchronize it with the upstream config from nginx.
But it's still possible to configure additional fastcgi params with the module, or refer to a different file.

vchepkov: It seems to me this change makes configuration to deviate from upstream package:

```
# rpm -qf /etc/nginx/fastcgi_params
nginx-1.12.1-1.el7.ngx.x86_64

# rpm -qi nginx
Name        : nginx
Epoch       : 1
Version     : 1.12.1
Release     : 1.el7.ngx
Architecture: x86_64
Install Date: Mon 18 Sep 2017 07:13:02 AM EDT
Group       : System Environment/Daemons
Size        : 2640392
License     : 2-clause BSD-like license
Signature   : RSA/SHA1, Tue 11 Jul 2017 10:20:52 AM EDT, Key ID abf5bd827bd9bf62
Source RPM  : nginx-1.12.1-1.el7.ngx.src.rpm
Build Date  : Tue 11 Jul 2017 09:50:21 AM EDT
Build Host  : centos7-amd64-builder-builder.gnt.nginx.com
Relocations : (not relocatable)
Vendor      : Nginx, Inc.
URL         : http://nginx.org/
Summary     : High performance web server
Description :
nginx [engine x] is an HTTP and reverse proxy server, as well as
a mail proxy server.

# diff -u /etc/nginx/fastcgi_params /etc/nginx/fastcgi.conf 
--- /etc/nginx/fastcgi_params	2017-07-11 09:50:19.000000000 -0400
+++ /etc/nginx/fastcgi.conf	2017-09-16 12:09:44.121363628 -0400
@@ -1,4 +1,6 @@
+# This file managed by puppet on host host.example.com
 
+fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;
 fastcgi_param  QUERY_STRING       $query_string;
 fastcgi_param  REQUEST_METHOD     $request_method;
 fastcgi_param  CONTENT_TYPE       $content_type;

```

This change also rolling back https://github.com/voxpupuli/puppet-nginx/commit/3404c4c651c0401d341eb7af2b77f5365b436434

which improves security of the server and is recommended by upstream. I don't think a typical user expects to have their configuration less secure by upgrading puppet module. It would be nice if there was at least an option to easily re-enable previous configuration.

wyardley:
I think `fastcgi.conf` is in sync with upstream still: https://github.com/nginx/nginx/blob/f8a9d528df92c7634088e575e5c3d63a1d4ab8ea/conf/fastcgi.conf
I don't think the module is making it _less_ secure than the default distribution; if you think other items should be added, bring that up with nginx. 

As best I could tell at the time we made these changes, `fastcgi.conf` is newer and slightly preferred, so we made the decision to use it instead of `fastcgi_params` These links exist above but just for reference:
https://www.nginx.com/resources/wiki/start/topics/examples/fastcgiexample/
https://blog.martinfjordvald.com/2013/04/nginx-config-history-fastcgi_params-versus-fastcgi-conf/

PR #1078, opened 2017-04-15
Test modulesync: allow running acceptance tests on travis #230

vinzent:
testing if it works.

allow running acceptance tests on travis https://github.com/voxpupuli/modulesync_config/pull/230


vinzent:
as expected the beaker tests were run.

PR #1080, opened 2017-04-19
Add hiera defaults configuration options for all resources

mvisonneau:
- Added hiera defaults configuration options for all resources
- Sorted variables
- Casted them to an hash

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -167,19 +167,23 @@ class nginx (
   ### END Service Configuration ###
 
   ### START Hiera Lookups ###
-  $geo_mappings                                              = {},
-  $string_mappings                                           = {},
-  $nginx_locations                                           = {},
-  $nginx_locations_defaults                                  = {},
-  $nginx_mailhosts                                           = {},
-  $nginx_mailhosts_defaults                                  = {},
-  $nginx_streamhosts                                         = {},
-  $nginx_upstreams                                           = {},
-  Nginx::UpstreamMemberDefaults $nginx_upstream_defaults     = {},
-  $nginx_servers                                             = {},
-  $nginx_servers_defaults                                    = {},
-  Boolean $purge_passenger_repo                              = true,
-  Boolean $add_listen_directive                              = $nginx::params::add_listen_directive,
+  Hash $geo_mappings                                      = {},
+  Hash $geo_mappings_defaults                             = {},
+  Hash $string_mappings                                   = {},
+  Hash $string_mappings_defaults                          = {},
+  Hash $nginx_locations                                   = {},
+  Hash $nginx_locations_defaults                          = {},
+  Hash $nginx_mailhosts                                   = {},
+  Hash $nginx_mailhosts_defaults                          = {},
+  Hash $nginx_servers                                     = {},
+  Hash $nginx_servers_defaults                            = {},
+  Hash $nginx_streamhosts                                 = {},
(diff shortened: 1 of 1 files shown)


oranenj:
Should there perhaps be new tests related to this change?

Either way, looks good to me.

wyardley:
I think this is partially covered by #1071 and similar PRs; aside from lack of tests, I think some of these resources may already have defaults that I think would get overwritten by these, e.g., https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/server.pp#L447-L453 (which is handled in #1071).

oranenj:
I merged the other PR since it was smaller and had tests, so this will need a rebase for the remaining functionality.

mvisonneau:
👋 apologies for the late update. Just rebased and added some additional checks.

wyardley:
@mvisonneau thanks for rebasing and adding tests. It seems as if some tests are failing in CI

bastelfreak:
Hi @mvisonneau, are you still interested in this PR? If so, please rebase against latest master. Afterwards we can check if travis is still failing.

manifests/init.pp
```diff
@@ -151,30 +151,34 @@
...
   contain '::nginx::service'
 
-  create_resources('nginx::resource::upstream', $nginx_upstreams)
-  create_resources('nginx::resource::server', $nginx_servers, $nginx_servers_defaults)
-  create_resources('nginx::resource::location', $nginx_locations, $nginx_locations_defaults)
-  create_resources('nginx::resource::mailhost', $nginx_mailhosts, $nginx_mailhosts_defaults)
-  create_resources('nginx::resource::streamhost', $nginx_streamhosts)
-  create_resources('nginx::resource::map', $string_mappings)
-  create_resources('nginx::resource::geo', $geo_mappings)
+  create_resources( 'nginx::resource::geo', $geo_mappings, $geo_mappings_defaults )
```
bastelfreak: Isn't this now considered bad practice? Instead we should loop over the hash. As an example: https://github.com/bastelfreak/puppet-ferm/blob/master/manifests/init.pp#L48

CC: @hunner @eputnam. What is your opinion here? We should update our review guidelines I guess.

mvisonneau: wouldn't this change imply another request due to its nature ?

bastelfreak: yes, let's keep it as is.

manifests/init.pp
```diff
@@ -167,33 +167,37 @@
...
+  create_resources( 'nginx::resource::location', $nginx_locations, $nginx_locations_defaults )
+  create_resources( 'nginx::resource::mailhost', $nginx_mailhosts, $nginx_mailhosts_defaults )
+  create_resources( 'nginx::resource::map', $string_mappings, $string_mappings_defaults )
+  create_resources( 'nginx::resource::server', $nginx_servers, $nginx_servers_defaults )
+  create_resources( 'nginx::resource::streamhost', $nginx_streamhosts, $nginx_streamhosts_defaults )
+  create_resources( 'nginx::resource::upstream', $nginx_upstreams, $nginx_upstreams_defaults )
```
alexjfisher: As per https://github.com/voxpupuli/puppet-nginx/pull/1309, good luck creating `nginx::resource::upstream` resources with defaults of type `Nginx::UpstreamMemberDefaults`...

PR #1081, opened 2017-04-25
Bump puppetlabs-concat, puppetlabs-stdlib and Puppet minimum versions

tdevelioglu:
This bumps the dependency on puppetlabs-concat to v3.0.0 and raises the minimum versions of the following dependencies to match:
  * puppet:            4.6.1 -> 4.7.0
  * puppetlabs-stdlib: 4.6.0 -> 4.13.1

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.6.0 <5.0.0"
+      "version_requirement": ">= 4.13.1 <5.0.0"
     },
     {
       "name": "puppetlabs/apt",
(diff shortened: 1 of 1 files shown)


yastupin:
stdlib dependencty would need to be bumped to 4.13.1 too

Do you need the updated concat module for something specific?

tdevelioglu:
@yastupin yup, I just updated it.

The reason I need this bump are changes in a [PR](https://github.com/puppetlabs/puppetlabs-concat/pull/428) that I have outstanding.

yastupin:
So maybe we should wait that the puppetlabs-concat PR is merged and released, and then bump the dependency to the version including the changes?

tdevelioglu:
It's partially merged (and released). The final commit is still pending, however. Sure, we can wait until it's finalized.

tdevelioglu:
@yastupin would it be wise to just bump it to 4.0.0 straight away (since we're bumping the puppet requirement to 4.7 anyway) ? (also see: https://github.com/puppetlabs/puppetlabs-concat/blob/master/CHANGELOG.md)

rhoml:
Hello folks any change this can get merged?

oranenj:
I don't really see a point in not merging this, since it's just a metadata change. Let's fix things later if there's something wrong with it...

PR #1082, opened 2017-04-30
Clean up nginx::resource::server

ekohl:


manifests/resource/server.pp
```diff
@@ -399,16 +395,14 @@
...
+  if $listen_port != $ssl_port {
     concat::fragment { "${name_sanitized}-header":
       target  => $config_file,
       content => template('nginx/server/server_header.erb'),
       order   => '001',
     }
-  }
 
-  # Create a proper file close stub.
-  if (($listen_port + 0) != ($ssl_port + 0)) {
```
yastupin: Hasn't this one accidentally been dropped and not replaced?

oranenj:
@yastupin: No. If I'm not reading the diff wrong, it duplicates the condition around the previous resource, so the cleanup was just to move the end of the if later.

yastupin:
@oranenj Yes thanks, I was the one reading it wrong.

PR #1091, opened 2017-06-17
Use correct scheme with rewrite_www_to_non_www

alfoeternia:
If `ssl_redirect` is enabled along with `rewrite_www_to_non_www`, then the user is redirected from the www version, to the http non-www one, and finally to https.


bastelfreak:
Thanks @alfoeternia!

PR #1093, opened 2017-07-11
Add http_raw_prepend and http_raw_append parameters

walkamongus:


manifests/init.pp
```diff
@@ -26,141 +26,143 @@
...
+  $gzip_disable                                              = 'msie6',
+  $gzip_min_length                                           = 20,
+  $gzip_http_version                                         = 1.1,
+  $gzip_proxied                                              = 'off',
+  $gzip_types                                                = undef,
+  $gzip_vary                                                 = 'off',
+  Optional[Variant[Hash, Array]] $http_cfg_prepend           = undef,
+  Optional[Variant[Hash, Array]] $http_cfg_append            = undef,
+  Optional[Variant[Array[String], String]] $http_raw_prepend = undef,
+  Optional[Variant[Array[String], String]] $http_raw_append  = undef,
```
wyardley: Just curious, in the templates, it seems like this is only putting it in if it's an array, so is the string type needed too? I see that this is how it's implemented in the server resource, just want to make sure this is done for a reason. Or am I misunderstanding how types work?

Just from a quick read, it seems like this would allow it to be defined as a string, but then not add anything in the resulting config?

walkamongus: Hi @wyardley  -- thanks for the comments! If a String is passed in, the template will cast it into an Array with one member e.g.

```
[1] pry(main)> v = 'String'
=> "String"
[2] pry(main)> Array(v)
=> ["String"]
```
I just used the added flexibility that was already at https://github.com/voxpupuli/puppet-nginx/blob/360e700d1aa98bdf9bb05f1379becdc01ed001ea/templates/server/location_header.erb#L67

wyardley: Got it, that makes sense, didn't realize it would do that!

Issue #1103, opened 2017-08-04
ensure => 'absent' on nginx::resource::server leaves file behind


dominics:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.0.1
- Module version: 0.7.0

## How to reproduce

Create a server with something like:

```
nginx::resource::server { "foo":
  server_name => ['foo.example.com'];
}
```

Then mark it absent:

```
nginx::resource::server { "foo":
  ensure => 'absent',
  server_name => ['foo.example.com'];
}
```

## What are you seeing

The locations are removed from the server, and the sites-enabled symlink is removed, but the emptied server definition still lives in sites-available. (As the locations are removed, it's kind-of useless; you couldn't "just" re-symlink it anyway.)

## What behaviour did you expect instead

I'd expect the server definition file to be removed from sites-available as well.

## Additional information

* I think this can be fixed just by passing the `$ensure` value to the concat_file definition at https://github.com/voxpupuli/puppet-nginx/blob/v0.7.0/manifests/resource/server.pp#L304 - that way the entire file will be absent, rather than just the locations within it.
* Not a super-important bug - already works functionally because the symlink is removed, which is the important bit. But just a bit annoying to have to define your own `file {}` to clean up after a server definition is removed.

wyardley:
I'm curious, if you remove the server resource rather than set `ensure => absent` (which is how I'd normally get rid of it), with `$server_purge` enabled, won't that ensure that the file is removed?  That's how I normally deal with it vs. doing `ensure => absent` on something I no longer need, so if you don't have unmanaged files that you need in those directories (or are using `$confd_only`), that might be an option.

That said, I agree with your assessment of what's happening and your proposed fix seems reasonable. Can you test #1104 and see if that works as you suggested?

wyardley:
ps - PRs also welcomed if that's something you're interested in / comfortable with.

wyardley:
Closed via #1104

alexjfisher:
[...]

PR #1104, opened 2017-08-04
Ensure absent on concat resource for server resource with ensure => absent (#1103)

wyardley:
While this doesn't seem like a super big deal (#1103), I agree with the proposed fix, and it's pretty easy to implement and to test for.


wyardley:
ps - I got that `Could not find semantic_puppet gem` error locally as well, maybe related to the ruby version requirements for rubocop in Gemfile?

Issue #1092, opened 2017-06-23
Redirect non-www to www


ogt00l:
We have numerous platforms where the preferred hostname is with the www. - this is either from a marketing point of view, or for technical reasons (uses an ELB or CDN that needs to be pointed to a CNAME record but the apex record can only be an A record, so it'll hit example.org (direct) and then get redirected to www.example.org which will come back through the CDN/ELB.

The rewrite_www_to_non_www does the exact reverse of what we're after and while we're currently solving by doing a separate vhost housing the custom redirect, this would be a much more graceful approach.

wyardley:
You know, I had always thought that it actually _was_ the opposite already.
I agree that this is definitely a common use-case, maybe the more common one.

I don't have time / energy to implement this personally, but I do think this is a good idea.

wyardley:
This turned out to be a much bigger headache than I thought it would be; that said, there is #1105 available for review. I'll see if I can get some suggestions to see if there's a way to simplify this or make it easier to read / follow.

If you'd like to do any functional testing from that branch (or apply the patch from it), feel free to test.

PR #1105, opened 2017-08-04
add new redirect_non_www_to_www option (#1092)

wyardley:
This was harder than I thought, mostly because of the way the existing feature was implemented. Welcome reviews, or suggestions on how to make this logic cleaner and more bulletproof.

I think this should work for the typical / simple use cases, but have not tested well. I don't really know what the behavior should be if someone put in a hostname like `www.www.example.com`.

Welcome feedback too on whether the existing tests are sufficient (and whether there are redundant tests), and specifically, whether I need to replicate https://github.com/wyardley/puppet-nginx/blob/issues_1092/spec/defines/resource_server_spec.rb#L965-L972 for SSL as well (the normal test for the other case doesn't really work well because the default hostname has www. in it).

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -69,8 +69,8 @@
 #   [*http2*]                      - Toggles HTTP/2 protocol.
 #   [*server_name*]                - List of servernames for which this server will respond. Default [$name].
 #   [*www_root*]                   - Specifies the location on disk for files to be read from. Cannot be set in conjunction with $proxy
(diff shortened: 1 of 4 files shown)


wyardley:
This one is a bit too complicated, maybe. Not sure if there's a better way to rework this to be easier to read or to be more re-usable. I'm going to rebase this against master, but also will probably close, as it seems like this one may be hard to review.

I think this should be possible to accomplish through the module using existing directives, though it does seem like a use case where it would be nice to have a default.

PR #1014, opened 2017-01-31
Fix for apt-transport-https

jas01:
The installation for the package apt-transport-https is put only for
the passenger case. But event with « just » nginx we need the
apt-transport-https because the url for nginx repot is on https://nginx

PR #1110, opened 2017-08-11
Fix dependency on apt-transport-https

rvdh:
Without this change, the nginx https repository url is added to sources.list.d
and apt-get update is executed immediately after, failing because
apt-transport-https is not yet installed. Subsequently all package
installs fail because the apt cache has been invalidated.

With this change we ensure the apt-transport-https package is present
before running apt-get update. Since all sources are now https, the
package is always necessary, which is why I moved it out of the 'case'
statement.

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -31,7 +31,8 @@ class nginx::package::debian {
 
   if $manage_repo {
     include '::apt'
-    Exec['apt_update'] -> Package['nginx']
+    ensure_packages([ 'apt-transport-https', 'ca-certificates' ])
+    Package['apt-transport-https','ca-certificates'] -> Exec['apt_update'] -> Package['nginx']
 
     case $package_source {
       'nginx', 'nginx-stable': {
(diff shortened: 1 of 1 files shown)


wyardley:
This seems sensible to me, and we can probably close #1014, apologies to the author of that one.

oxc:
I'm afraid the update of puppetlabs-apt to 4.4.0 (https://forge.puppet.com/puppetlabs/apt/4.4.0/changelog) breaks compatibility with this change, I'm getting the following dependency cycle:

>  Error: Found 1 dependency cycle:
(Package[apt-transport-https] => Apt::Source[nginx] => Package[apt-transport-https])

Does this mean the dependency should be removed again?

ekohl:
I do think it was incompatible before but you're right: we should remove all off the apt https code and rely on puppetlabs-apt 4.4.0 to do it right. @oxc would you be willing to submit a PR?

oxc:
@ekohl unfortunately I'm not totally sure what to remove, and how to test if it does not break anything. Especially the passenger-specific require parameter puzzles me a bit.

ekohl:
If you could test https://github.com/voxpupuli/puppet-nginx/pull/1163 that'd be great.

oxc:
Turns out I get the cycle even without manage_repos, so I won't be able to test your fix yet (because ther error even occurs without the explicit dependencies on apt-transport-https)

PR #1111, opened 2017-08-14
Add DragonFly BSD support

strangelittlemonkey:


wyardley:
I'm not sure how much demand there is, but this seems fairly harmless.

PR #1115, opened 2017-08-27
fix lint warnings

PascalBourdier:


manifests/resource/mailhost.pp
```diff
@@ -1,5 +1,7 @@
 # define: nginx::resource::mailhost
 #
+# lint:ignore:140chars
+#
```
wyardley: Can we just wrap these lines @ 140 char rather than ignoring the lint warnings? I realize it's just docs, but...

If you indent the following line one more space
```
# blah blah
#   blah
```
it will get continued as one lines in the YARD doc, I think. Also, we'll switch to Puppet strings for param docs at some point anyway.

manifests/resource/location.pp
```diff
@@ -251,7 +251,7 @@
   # Only try to manage these files if they're the default one (as you presumably
   # usually don't want the default template if you're using a custom file.
 
-  if $ensure == present and $fastcgi != undef and !defined(File[$fastcgi_params]) and $fastcgi_params == "${::nginx::conf_dir}/fastcgi.conf" {
+  if $ensure == present and $fastcgi != undef and !defined(File[$fastcgi_params]) and $fastcgi_params == "${::nginx::conf_dir}/fastcgi.conf" { # lint:ignore:140chars
```
wyardley: Can this conditional just be broken up into multiple lines as well?

wyardley:
Awesome. I squash-merged it rather than making you squash the commits. Thanks for the contribution.

Issue #1120, opened 2017-09-01
Location ssl and ssl_only checks do not match the description in the README.mf


mbelscher:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet:
- Ruby:
- Distribution:
- Module version: all

## How to reproduce (e.g Puppet code you use)
Set $ssl to false on location and $ssl_only to true
## What are you seeing
Code ignores that $ssl is disabled on the location and creates an invalid configuration.
## What behaviour did you expect instead
I expected like the README states set $ssl => false and then it will only put location in an HTTP server.

wyardley:
README also says `To add a location only to the HTTPS server, set both ssl => true and ssl_only => true on the location.` https://github.com/voxpupuli/puppet-nginx#http-and-https-server
so I don't think that `$ssl => false` and `$ssl_only => true` is a valid config.

What is the desired outcome of setting both `$ssl => false` and `$ssl_only => true`, or am I misunderstanding?

mbelscher:
There are 3 cases listed in the README for the handling of the location 
HTTP only:  ` $ssl => false`
HTTP and HTTPS: ` $ssl=>true` and` $ssl_only=>true`
HTTPS only:  `$ssl=>true` and` $ssl_only => true`


The way the current code is working is
HTTP only:  `$ssl=> false` and `$ssl_only->false`
HTTP and HTTPS: `$ssl=>true`  and `$ssl_only=>false`
HTTPS only: `$ssl_only=>true`  ; for this case it is irrelevant the setting of $ssl
  
 
Based on the Readme section you listed.` $ssl=> false` should only produce a location in an HTTP server is what I'm expecting and also am expecting an HTTPS only section if both $ssl =>true and `$ssl_only=>true`. By setting` $ssl=>false`.  You have told it that you are not supporting ssl  for the location . This should make the $ssl_only flag irrelevant,

wyardley:
_By setting$ssl=>false. You have told it that you are not supporting ssl for the location . This should make the $ssl_only flag irrelevant,_
I agree, but I think the module should error in this case. 
Your "how to reproduce" has:
"Set $ssl to false on location and $ssl_only to true"

If `ssl` is false, there should not be an SSL location block
if `ssl_only` is false, there  should not be a non-ssl location block. In other words, this would create no location blocks, so the module should just fail if someone sets this combination of settings IMHO.

wyardley:
[...]

PR #1121, opened 2017-09-01
Fix location ssl and ssl_only checks to match README.md

mbelscher:
This Addresses issue #1120.

The documentation states that if you set  $ssl to false it will only add to http server.

If $ssl is true it will then add to https and http if ssl_only is false.

This is causing a little confusion on how to set the values.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -275,7 +275,7 @@ define nginx::resource::location (
   if $ensure == present {
     ## Create stubs for server File Fragment Pattern
     $location_md5 = md5($location)
-    if ($ssl_only != true) {
+    if ($ssl != true or $ssl_only != true) {
       concat::fragment { "${server_sanitized}-${priority}-${location_md5}":
         target  => $config_file,
         content => template('nginx/server/location.erb'),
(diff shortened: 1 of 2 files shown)


mbelscher:
Where are these tests defined.  Are they in a separate. It seems that the same incorrect logic that doesn't match the description that was in the location.pp files was written by the same to produce a success.

Based on the documentation it is irrelevant what the setting if $ssl_only is if $ssl = false

wyardley:
@mbelscher The line that's failing is
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/defines/resource_location_spec.rb#L923

in
https://travis-ci.org/voxpupuli/puppet-nginx/jobs/270953616
it has:
```
  1) nginx::resource::location on debian-7-x86_64 os-independent items attribute resources when ssl_only => true should not contain Concat::Fragment[server1-500-33c6aa94600c830ad2d316bb4db36724]
     Failure/Error: it { is_expected.not_to contain_concat__fragment('server1-500-' + Digest::MD5.hexdigest('rspec-test')) }
       expected that the catalogue would not contain Concat::Fragment[server1-500-33c6aa94600c830ad2d316bb4db36724]
     # ./spec/defines/resource_location_spec.rb:923:in `block (7 levels) in <top (required)>'
```
https://github.com/voxpupuli/puppet-nginx/blob/master/.github/CONTRIBUTING.md has some more details about contributing and how to get tests setup.

bastelfreak:
Hi @mbelscher. thanks for the PR. Can you please rebase this against latest upstream master branch and take a look at the used email address in your commits? It isn't associated with your github account.

bastelfreak:
Hi. I'm going to close this PR due to inactivity. Please reopen it if you still want to work on it.

PR #1126, opened 2017-09-25
add settable nginx daemon group

miksercz:
Hello,
in a problem I was trying to solve with your module today, I needed to specify both `user` and `group` name for the nginx.conf `name` descriptor, as these were not the same. Simply setting the `daemon_user` param to `$user $group` doesn't work, because it is used below in config.pp as an owner for multiple directories:

```
file { $log_dir:
  ensure => directory,
  mode   => $log_mode,
  owner  => $daemon_user,
  group  => $log_group,
}
```

Therefore I believe I had to add this parameter, which fixes this issue nicely.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -23,6 +23,7 @@ class nginx::config {
   $conf_dir                       = $::nginx::conf_dir
   $daemon                         = $::nginx::daemon
   $daemon_user                    = $::nginx::daemon_user
+  $daemon_group                   = $::nginx::daemon_group
   $global_owner                   = $::nginx::global_owner
   $global_group                   = $::nginx::global_group
   $global_mode                    = $::nginx::global_mode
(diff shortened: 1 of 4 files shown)


wyardley:
Thanks, we had #647 which was going to implement this, however it never ended up getting rebased or updated, and we closed it.

This looks like the right fix to me, but it needs spec tests, please let us know if you need help figuring out how to add tests for this.

https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb#L341-L350
https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb#L1069-L1074
etc.... probably you'd want to actually test the behavior in that template, e.g., that the group is present in the generated config when specified, and suppressed when not.

miksercz:
Sorry for the long wait - I added two tests with help from people on slack, let me know if you want to test any additional cases.

wyardley:
Thanks for the follow-through @miksercz

Issue #1123, opened 2017-09-10
Making puppet-nginx work with puppet-letsencrypt in webroot mode


kwisatz:
I wanted to see if someone else already solved this problem or, if not, discuss it so that we may find a viable and functional solution.

We're in a HA setup were using any lenc plugin other than `webroot` for renewal and certificate creation is not an option.

However, getting to a working configuration with puppet is now a two-step process for us, which kind of defeats the purpose of using puppet in the first place:

1. Define the server in nginx to run on `port 80` _-> run puppet_
1. Define the letsencrypt configuration to have a certificate be generated and enable `ssl` on nginx _-> run puppet again_

This is necessary since we have a circular dependency here:

- let's encrypt with `webroot` depends on the .well-known location having been defined and being available
- nginx ssl depends on the certificate and key files already existing

To the best of my knowledge, both can't currently be satisfied in a single puppet agent run.

If I understand the anchor resource correctly – and I'm not sure I do – a possible solution would be to chain the execution flow like this:

1. realize nginx server on `port 80` including `webroot location`
1. realize letsencrypt certificate generation
1. realize nginx server `443 ssl;` configuration

I'm not entirely sure yet how one would go about implementing this or if this makes any sense in the first place, but I'm very keen on seeing if anyone has had the same problem and has either solved it (differently) or is interested in discussing and coming up with a solution.

ekohl:
You're right: this is not something puppet can easily do. You can think of some workarounds, for example create a custom fact ```enable_ssl``` which looks at if a certificate has been requested. Then you only create the SSL-enabled vhosts if that fact is true. Then you need 2 runs but at least it's automated.

A more complicated solution would be to chain the resources in such a way that you can start the server with an exec. It is possible that the following works but you will have double nginx reloads/restarts when a non-ssl vhost changes.

```puppet
exec { 'ensure nginx':
  command     => '/bin/systemctl reload-or-restart nginx':
  refreshonly => true,
  require     => Class['nginx::config'],
}

Nginx::Resource::Server <| ssl == false |> ~> Exec['ensure nginx'] -> Class['letsencrypt'] -> Nginx::Resource::Server <| ssl == true |>
```

kwisatz:
Hi, @ekohl, thanks for your input! 

Please CMIIW:

[...]

PR #1127, opened 2017-09-30
refactoring ssl_redirect block in template for a more flexible location-…

law:
…/ redirect

A redirect in this format is necessary to support sites who would like to use the 'webroot' functionality of LetsEncrypt (a common program/organization who provides signed SSL certificates for free).  

  (Exceedingly) long story short, the LE-webroot plugin makes an http (not https) request to the server, and expects a token to be in a specific directory.  Any attempts to blindly rewrite the http request to https, regardless of the current state of the server's cert, will fail.  Folks who expect LE certs to auto-renew via this method tend to get a little... excitable when they don't.  This refactor makes the :80 listener perform its redirect against a specific location (location /), instead of a global non-location-specific redirect.  From there, we can add a location-block for the Lets Encrypt directory that does NOT get picked up by a global redirect, for appropriate processing on the LE side.  

  For example, this will fail (LE-request gets a 301 redirect to https; tested on nginx 1.10.3 on Ubuntu 16.04.3 LTS, nginx 1.13.5-1 on Debian Testing/Buster): 
```
# nginx server definition in /etc/nginx/sites-enabled/example.com.conf
server {
  listen *:80 default_server;
  server_name           example.com;

  location /.well-known {
    root      /var/www/letsencrypt/;
    try_files $uri $uri/ =404;
  }
  
}
```

  This will succeed (LE-request succeeds over HTTP, everything else gets the 301 redirect to HTTPS)
```
# nginx server definition in /etc/nginx/sites-enabled/example.com.conf
server {
  listen *:80 default_server;
  server_name           example.com;

  location / {
  }

  location /.well-known {
    root      /var/www/letsencrypt/;
    try_files $uri $uri/ =404;
  }
}
```

  Supporting Puppet role definition: 
```
class role::webserver (
  $nginx_vhosts,
  $nginx_location,
) {
  class { 'nginx': }
    create_resources('nginx::resource::server', $nginx_vhosts)
    create_resources('nginx::resource::location', $nginx_location)
} 
```
  Supporting Hieradata: 
```
---
role::webserver::nginx_vhosts:
  "example-vhost":
    www_root: '/var/www'
    server_name: [ "example.com" ]
    access_log: '/var/log/nginx/example-access.log'
    error_log: '/var/log/nginx/example-error.log'
    ssl_redirect: true
    ssl: true
    ssl_cert: "/etc/letsencrypt/live/example.com/fullchain.pem"
    ssl_key: "/etc/letsencrypt/live/example.com/privkey.pem"

role::webserver::nginx_location:
  'letsencrypt-well-known':
    location: '/.well-known'
    www_root: '/var/www/letsencrypt'
    try_files: ['$uri $uri/ =404']
    ssl_only: false
    server: "example-vhost"
```

### DIFF
--- a/templates/server/server_header.erb
+++ b/templates/server/server_header.erb
@@ -31,7 +31,9 @@ server {
   <%- end -%>
   server_name  www.<%= s.gsub(/^www\./, '') %>;
   <%- if @ssl_redirect or @ssl_only -%>
-  return       301 https://<%= s.gsub(/^www\./, '') %><% if @_ssl_redirect_port.to_i != 443 %>:<%= @_ssl_redirect_port %><% end %>$request_uri;
+  location / {
+    return       301 https://<%= s.gsub(/^www\./, '') %><% if @_ssl_redirect_port.to_i != 443 %>:<%= @_ssl_redirect_port %><% end %>$request_uri;
+  }
   <%- else -%>
   return       301 http://<%= s.gsub(/^www\./, '') %>$request_uri;
   <%- end -%>
@@ -143,7 +145,9 @@ server {
   <%= @maintenance_value %>;
 <% end -%>
 <% if @ssl_redirect -%>
-  return 301 https://$host<% if @_ssl_redirect_port.to_i != 443 %>:<%= @_ssl_redirect_port %><% end %>$request_uri;
+  location / { 
+    return 301 https://$host<% if @_ssl_redirect_port.to_i != 443 %>:<%= @_ssl_redirect_port %><% end %>$request_uri;
+  }
 <% end -%>
 <% if @index_files and @index_files.count > 0 and not @ssl_only -%>
   index <% Array(@index_files).each do |i| %> <%= i %><% end %>;


wyardley:
My worry is that I think this kind of could subvert the purpose / behavior of the `ssl_redirect` parameter in allowing the user to create locations that SSL does not redirect.

FWIW, this works for me to create more or less what I think you're requesting:
[separately, not sure why setting `index_files` to `undef` isn't working in my tests for `nginx::resource::location`; will look into that]

```
class { 'nginx': }
nginx::resource::server { 'www.puppetlabs.com':
  ensure               => present,
  ssl                  => true,
  ssl_cert             => '/etc/pki/tls/certs/blah.cert',
  ssl_key              => '/etc/pki/tls/private/blah.key',
  www_root             => '/var/www/www.puppetlabs.com',
  use_default_location => false,
}
nginx::resource::location { "/":
  index_files => [],
  raw_prepend => 'return 301 https://$host$request_uri;',
  server      => 'www.puppetlabs.com',
}
nginx::resource::location { '/.well-known':
  index_files => [],
  try_files => ['$uri $uri/ =404'],
  www_root  => '/var/www/letsencrypt/',
  server    => 'www.puppetlabs.com',
}
```

```
[root@centos-7-x64 tmp]# curl -H "Host: www.puppetlabs.com" localhost/.well-known/
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.12.1</center>
</body>
</html>
[root@centos-7-x64 tmp]# 
[root@centos-7-x64 tmp]# curl -H "Host: www.puppetlabs.com" localhost
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.12.1</center>
</body>
</html>
```

wyardley:
@law were you able to accomplish what you needed to with the module's existing params? I'm still a bit concerned about the security implications of this one.

bastelfreak:
ping @law. Did you see the comments from @wyardley ?

law:
Hey, my apologies for the long reply-time.  I have not been able to accomplish my tasks within the existing params, I've resigned myself to manually updating LE certs until a better method comes along.

law:
Perhaps this could be contained in a 'LE compatibility => true' attribute?  Enable the attribute and it creates _exactly_ this scenario, with the only variable-change allowed is the path to the LE challenge-root (defaulting to /var/www/le-certs, or some such).  This would prevent @wyardley's concern of arbitrary/accidental/unintended ssl-redirect-disabling, but fix my concern of "I can't make LE do the needful when I use this module".

wyardley:
I know it's already linked above, but #1123 did have some discussion of the broader issues around getting LE to play nicely with Puppet / nginx. The whole issue of handing LE in an idempotent way is kind of tricky. 
https://github.com/voxpupuli/puppet-letsencrypt/issues/66 also has some discussion (more focused on Apache, but same issues).

I do think that my workaround in https://github.com/voxpupuli/puppet-nginx/pull/1127#issuecomment-333285459 (or something like it) accomplishes the main intent in the issue under discussion here, though, unless I'm missing something (in terms of redirecting requests other than, e.g., `/.well-known`). @law: is there a reason that doesn't work for your use case?

law:
It's been a minute since I've tried it, but IIRC I attempted to duplicate your config, and couldn't get it to function.  Give me a little bit (under the gun for some projects ATM) and I'll try to duplicate again, posting specifics, etc, if it continues to fail.

anthonysomerset:
I will +1 this request because it would allow the needed manual overrides of locations to bypass the HTTPS redirect for the edge cases that need it such as LetsEncrypt

The default behaviour would still redirect all non matched traffic properly and anyone adding HTTP(S) only locations for SSL bypass should at least have read the module documentation enough to understand the ramifications of such. one could also extend the documentation around this specific area accordingly to mitigate such concerns as well.

wyardley:
@anthonysomerset is there a reason you believe the existing ways of doing this don’t work (as outlined in https://github.com/voxpupuli/puppet-nginx/pull/1127#issuecomment-333285459)?

anthonysomerset:
i get duplicate location define errors for location / when i try the same approach without setting ssl_redirect => true

if i set ssl_redirect => true then the location goes in but the straight redirect rule takes precedence because its outside of any location block

wyardley:
Are you defining `/` twice, or getting that error because you don’t have `use_default_location` set to `false`? You should only have to define that location once if you’re defining a more specific location to bypass TLS.

anthonysomerset:
thanks @wyardley  - i was missing `use_default_location` , once i set that false, i was able to specify 2 seperate / location blocks - first for http only for the https redirect and a second one for https only to set the proxy to backend and other params that i had previously when i had default location enabled.

my example config (other elements excluded for brevity):
```
  nginx::resource::server { $server_name:
      ssl              => true,
      use_default_location => false,
      ssl_cert         => "/etc/letsencrypt/live/$server_name/fullchain.pem",
      ssl_key          => "/etc/letsencrypt/live/$server_name/privkey.pem",
      http2            => 'on',
  }
  
  nginx::resource::location{ 'https-redirect': 
      ensure     => present, 
      server      => $server_name, 
      location   => '/', 
      www_root   => '/var/www/', 
      ssl   => false, 
      location_cfg_append => { 
        'return' => '301 https://$host$request_uri' 
      } 
  }
  
  nginx::resource::location{ 'https-proxy': 
      ensure     => present, 
      server      => $server_name, 
      location   => '/', 
      proxy =>  $upstream,
      ssl   => true,
      ssl_only   => true, 
  } 

  nginx::resource::location{ 'letsencrypt-well-known':
      ensure     => present,
      server      => $server_name,
      location   => '/.well-known',
      www_root   => '/var/www/letsencrypt',
      ssl   => true,
  }
```

wyardley:
@anthonysomerset Yes, thanks, this is a common source of confusion w/ the project, even though it is documented, I believe there's room for improvement there. I think the original intent was to have the very simple use case work "out of the box". There have been some discussions about ways to make this a little more optimal (for example, just overriding the default location when `/` is defined), but I don't think anyone's had a chance to work on it. Given that, I'd like to close this. However, if someone wants to open a docs fix, or a fix that has a lower chance of creating an insecure configuration, I'm happy to get some eyes on it.

Issue #1128, opened 2017-09-30
setting index files to undef doesn't work as expected


wyardley:
For some reason, this works to suppress the index directive:

```
nginx::resource::server { 'www.puppetlabs.com':
  index_files          => [],  
  use_default_location => false,
}
nginx::resource::location { "/":
  index_files => [],
}
```

But this doesn't:
```
nginx::resource::server { 'www.puppetlabs.com':
  index_files          => undef,
  use_default_location => false,
}
nginx::resource::location { "/":
  index_files => undef,
}
```

wyardley:
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/server.pp#L203-L206
and https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/location.pp#L162-L165 are where they're defined; see the linked PR for where the test cases are.

wyardley:
@walkamongus so it turns out this isn't actually valid

passing in
```
nginx::resource::server { 'foo':
  index_files => undef,
}
```
gets the default behavior, vs. suppressing it.

So this creates unexpected behavior. Also, the tests don't work as expected. So I'll probably more or less revert #1094 in #1129, as well as updating the tests.

walkamongus:
@wyardley yep -- that looks to be correct. I found some discussion of this at https://tickets.puppetlabs.com/browse/PUP-5295. It looks like there is a desire to make the distinction between 'default value' and  undef easier, but it hasn't been done due to backwards incompatibility issues.

Interestingly enough, after looking at https://docs.puppet.com/puppet/5.2/lang_data_default.html#anywhere-else and knocking up a test case, setting the param to `default` instead of `undef` resulted in the expected behavior of "ignoring" the default value:

```
~$ cat test.pp
class testclass (
  Variant[Undef,Default,Array] $files = ['one', 'two', 'three']
) {
  if $files { notify { $files: } }
}

class {'testclass':
  files => undef,
}
~$ puppet apply test.pp
Notice: Compiled catalog for localhost.localdomain in environment production in 0.10 seconds
Notice: one
Notice: /Stage[main]/Testclass/Notify[one]/message: defined 'message' as 'one'
Notice: two
Notice: /Stage[main]/Testclass/Notify[two]/message: defined 'message' as 'two'
Notice: three
Notice: /Stage[main]/Testclass/Notify[three]/message: defined 'message' as 'three'
Notice: Applied catalog in 0.02 seconds


[...]

PR #1130, opened 2017-10-03
Remove 'Optional' for resources with default settings

wyardley:
See #1128, but basically, having default settings for Optional params gives some unexpected results (because passing in undef gives you the default vs. setting undef)


ekohl:
Are the acceptance tests failing in ```beaker-rspec```?

ekohl:
Looked like https://github.com/puppetlabs/beaker-rspec/commit/6809f32c4fc684197eaf0d7e199a93d54c4b8eff fixed it.

wyardley:
Technically, it is possible to set undef in hiera (though not directly from the DSL). Not sure worth re-allowing this, though, as it could cause confusion in terms of how to use the module.

PR #1141, opened 2017-11-02
Update upstream.pp

jabreity:
Per the ERB template, the value is inserted even if it's not a hash.
```
  <%- @upstream_cfg_prepend.each do |key,value| -%>
    <%- if value.is_a?(Hash) -%>
      <%- value.each do |subkey,subvalue| -%>
        <%- Array(subvalue).each do |asubvalue| -%>
  <%= key %> <%= subkey %> <%= asubvalue %>;
        <%- end -%>
      <%- end -%>
    <%- else -%>
      <%- Array(value).each do |asubvalue| -%>
  <%= key %> <%= asubvalue %>;
      <%- end -%>
```

### DIFF
--- a/manifests/resource/upstream.pp
+++ b/manifests/resource/upstream.pp
@@ -6,7 +6,7 @@
 #   [*members*]               - Array of member URIs for NGINX to connect to. Must follow valid NGINX syntax.
 #                               If omitted, individual members should be defined with nginx::resource::upstream::member
 #   [*ensure*]                - Enables or disables the specified location (present|absent)
-#   [*upstream_cfg_append*]   - Hash of custom directives to put after other directives in upstream
+#   [*upstream_cfg_append*]   - Hash or value of custom directives to put after other directives in upstream
 #   [*upstream_cfg_prepend*]  - It expects a hash with custom directives to put before anything else inside upstream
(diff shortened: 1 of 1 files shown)


wyardley:
Thanks, this is helpful.
I think maybe it should be 'Hash or Array'?  Would you mind adjusting that?

This would also get fixed in a better way when we convert everything to data types and use Puppet Strings for documentation. Let me know if you're interested in doing any further work in improving the module docs!

ekohl:
It does have a data type that prevents from allowing this: https://github.com/voxpupuli/puppet-nginx/blob/8c49c4776efe8b11399a92a02089415f8c11773e/manifests/resource/upstream.pp#L47

wyardley:
> It does have a data type that prevents from allowing this: 
Good point. So that would have to be updated to be `Optional[Variant[Hash, Array]]` or something before we could update the docs.

wyardley:
@jabreity do you want to make the suggested changes to docs and to the data type?

bastelfreak:
Hi @jabreity are you still interested in this PR?

jabreity:
Indeed, and I'm depending upon the ability to pass more configuration details in without extending the puppet module.  I realized the hackery I was depending upon wasn't ideal at all.  Let me revisit my code and validate my thoughts tonight!

bastelfreak:
Hi. I'm going to close this PR. Please reopen it if you're still interested.

PR #1145, opened 2017-11-09
Adding support for include in map and support some proxy_cache options

ceonizm:
Hello, 
I needed to use the include directive in a map block to handle some redirections that were dynamically generated by an other app. I noticed this feature was missing from your module so I've added it.

I've also added the support for proxy_cache_bypass and proxy_cache_lock 

Hopes it helps.

Best Regards
François Boukhalfa


templates/conf.d/map.erb
```diff
@@ -6,6 +6,11 @@ map <%= @string %> $<%= @name %> {
 <% if @default -%>
   default <%= @default %>;
 <% end -%>
+<% if @includes.is_a?(Array) -%>
```
alexjfisher: Can it ever not be an array?

templates/server/locations/proxy.erb
```diff
@@ -46,4 +46,12 @@
 <% if @proxy_cache_key -%>
     proxy_cache_key         <%= @proxy_cache_key %>;
 <% end -%>
+<% if @proxy_cache_bypass && Array(@proxy_cache_bypass).size > 0 -%>
```
alexjfisher: Is the size check needed?

alexjfisher:
@ceonizm thanks! Would it be possible to split this PR up by feature you’re adding? Eg support for access log in a location seems unrelated to the proxy_cache changes.

templates/server/locations/proxy.erb
ceonizm: Initially when I've added theses features I've started from the 0.5.0 version which was published on forge and followed the same coding style, that's why there is this check I believe: 
I totally agree with you it seems useless now.

ceonizm:
I close this PR and open some new one from different branches in order to get all features splitted.
- #1149 
- #1150 
- #1151

alexjfisher:
@ceonizm Thanks.  This will make it easier to review and get your additions merged.

PR #1149, opened 2017-11-16
adding support for include directive in map

ceonizm:
As requested in previous PR (#1145) I've splitted features in multiples branches and PR:

Hello,
I needed to use the include directive in a map block to handle some redirections that were dynamically generated by an other app. I noticed this feature was missing from your module so I've added it.

Best Regards
François.

### DIFF
--- a/manifests/resource/map.pp
+++ b/manifests/resource/map.pp
@@ -10,6 +10,7 @@
 #   [*mappings*]   - Hash of map lookup keys and resultant values
 #   [*hostnames*]  - Indicates that source values can be hostnames with a
 #                    prefix or suffix mask.
+#   [*includes*]   - An array of external files to include
 #
 # Actions:
 #
@@ -37,6 +38,14 @@
 #      { 'key' => '*.nyc.example.com', 'value' => 'ny-pool-1' },
 #    ]
 #  }
+#
+# Sample Usage (using external include)
+#
+# nginx::resource::map { 'redirections':
+#
+#    includes => [ '/etc/nginx/conf.d/redirections.map']
+#
+# }
 #
 # Sample Hiera usage:
 #
@@ -67,6 +76,7 @@ define nginx::resource::map (
   Variant[Array, Hash] $mappings,
   Optional[String] $default         = undef,
   Enum['absent', 'present'] $ensure = 'present',
+  Array[String] $includes           = [],
   Boolean $hostnames                = false
 ) {
   if ! defined(Class['nginx']) {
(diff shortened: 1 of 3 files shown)


manifests/resource/map.pp
```diff
@@ -67,6 +76,7 @@
   Variant[Array, Hash] $mappings,
   Optional[String] $default         = undef,
   Enum['absent', 'present'] $ensure = 'present',
+  Array[String] $includes           = [],
```
wyardley: Will this work if it's an empty array, since you're not checking if it's empty in the template? Will this create an empty `includes` line?
Maybe this should be `Optional[Array[String]] $includes`?

wyardley:
Can you add spec tests for this?
And is the case where the array is empty safe? Maybe this should be optional (see review comment)?

wyardley:
FWIW, there's a consistency issue here as well, as server resource uses `include_files`:
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/server.pp#L100
https://github.com/voxpupuli/puppet-nginx/blob/master/templates/server/server_footer.erb#L1-L4
Personally, I actually like `include` better, since it matches the directive, as long as there's no conflict, but IMHO, it's best if server and location use like parameters for like directives, if that makes sense.

manifests/resource/map.pp
ceonizm: Because there's a default value you can omit it ; it acts as an optional parameter. 
So if you pass an empty Array it will be the same as passing nothing and use the default value: it won't iterate nothing into the loop in template so nothing will be written (no empty includes line).
I agree with you since it is finally an optional parameter that declaring it explicitly as Optional would be better (but in that case the test in the template will became mandatory).

alexjfisher:
I much prefer *not* making it `Optional` and defaulting to the empty array instead.  As @ceonizm points out, there will be no output in the template when the array is empty.

For consistency with `server.pp` I agree the parameter name should be `include_files` (even if `includes` would have otherwise been a better name).

I'd be tempted to change the `server.pp` `include_files` to also just default to an empty array and simplify the template.  https://github.com/voxpupuli/puppet-nginx/blob/9a5e444ea518e5f903b620f165aac606e2db873d/templates/server/server_footer.erb#L1 and https://github.com/voxpupuli/puppet-nginx/blob/9a5e444ea518e5f903b620f165aac606e2db873d/templates/server/server_footer.erb#L5 could then be safely removed.

wyardley:
I squash-merged, thanks for the contribution @ceonizm! With the other PRs, if you're able to squash / force-push (or break into logical commits), that would be great.

PR #1154, opened 2017-11-18
Enhance distro support

fnoop:
Add explicit Debian 9 (stretch) support, and add support for Ubilinux.

Ubilinux 4 (that supports all variants of the Up boards) is a derivative of Debian 9. However it reports dolcetto as the lsbdistcodename instead of stretch so the nginx repo fails. This PR adds support for a $nginx::repo_release parameter that allows the distro code name to be specified rather than blindly taking it from lsbdistcodename.

```
    class { 'nginx':
        manage_repo     => true,
        repo_release    => "stretch",
    }
```

See:
https://github.com/voxpupuli/puppet-nodejs/pull/342

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -21,6 +21,7 @@ class nginx::package::debian {
   $package_flavor           = $::nginx::package_flavor
   $passenger_package_ensure = $::nginx::passenger_package_ensure
   $manage_repo              = $::nginx::manage_repo
+  $release                  = $::nginx::repo_release
 
   $distro = downcase($::operatingsystem)
 
@@ -40,6 +41,7 @@ class nginx::package::debian {
           location => "https://nginx.org/packages/${distro}",
           repos    => 'nginx',
           key      => {'id' => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62'},
+          release  => pick($release, $::lsbdistcodename),
         }
       }
       'nginx-mainline': {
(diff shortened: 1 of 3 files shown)


wyardley:
I wonder if it would maybe be better to approach this just with another case in params (or with data in modules if #1155 gets merged?

manifests/package/debian.pp
```diff
@@ -40,13 +41,15 @@
           location => "https://nginx.org/packages/${distro}",
           repos    => 'nginx',
           key      => {'id' => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62'},
+          release  => pick($release, $::lsbdistcodename),
```
ekohl: You can actually pass ```undef``` to puppetlabs-apt and let the module handle it meaning you don't need to pick: using ```$release``` will just work.

fnoop: Good point, changed in https://github.com/voxpupuli/puppet-nginx/pull/1154/commits/b85c0607cfc0081edbde4f48496b543480004667

bastelfreak:
@fnoop can you add Ubilinux and Debian 9 to the metadata.json? Are you also interested in adding Ubilinux to [facterdb](https://github.com/camptocamp/facterdb/)? This is a gem with a list of mocked facts. This is heavily used for our unit testing (without that we can't test for Ubilinux).

bastelfreak:
We should also add debian 9 to the acceptance test matrix if this is now working.

bastelfreak:
@fnoop can you please rebase?

wyardley:
Closing via #1200; thanks for the contribution.

fnoop:
Apologies wasn't able to find time to rebase and finish this off, thanks for doing it :)

PR #1155, opened 2017-11-18
Use hiera data

ekohl:


wyardley:
I think it would definitely be useful if we had a working prototype of "data in modules" if that's the direction things are heading.

bastelfreak:
data in modules \o/

ekohl:
Using .yml extensions while hiera looks for .yaml makes rspec sad. Should be better now.

data/ArchLinux.yaml
```diff
@@ -0,0 +1,4 @@
+---
+nginx::daemon_user: http
```
bastelfreak: Archlinux \o/

I'm pretty sure filename should be Archlinux.yaml

ekohl: I copied it from ```params.pp``` where it's also ```ArchLinux```. Might be good to verify there's tests for it.

ekohl:
This is still WIP because we'll need to bump the minimal puppet version to something 4.9+ or 4.10+. puppetlabs-ntp has 4.9.4 as minimal version and uses hiera 5.

manifests/params.pp
```diff
@@ -1,124 +0,0 @@
-# Class: nginx::params
-# ====================
-#
-# nginx default settings and according to operating system
-#
-class nginx::params {
-  ### Operating System Configuration
-  ## This is my hacky... no hiera system. Oh well. :)
```
ekohl: Fixed! :)

data/Solaris.yaml
```diff
@@ -0,0 +1,3 @@
+---
+nginx::daemon_user: webservd
+nginx::package_name: undef
```
ekohl: I'm pretty sure this is not the way to unset it and need a better solution.

metadata.json
```diff
@@ -20,7 +20,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.7.0 < 6.0.0"
+      "version_requirement": ">= 4.9.4 < 6.0.0"
```
bastelfreak: We had a discussion yesterday on the IRC about lowest version of puppet in modules with data-in-modules. We decided to go with 4.10.9 as there were soo many bug fixes for the hiera implementation.

ekohl:
Updated

bastelfreak:
Should we merge this now as is?

ekohl:
Should we also consider this a major version bump? I'd like to get some more reviews on this.

bastelfreak:
yeah a major bump is probably needed.

dhollinger:
puppet-rsyslog uses data in modules if you want some other examples

ekohl:
I based it on theforeman/tftp wich also uses data in modules. I'm fairly confident it works but maybe the bigger point is that I want an ACK that this is something we want in our modules and maybe a new goal to work towards.

bastelfreak:
@ekohl can you rebase this? I would like to the the 1.0.0 release next and include this. Bumping lowest puppet version boundary to 4.10.10 is fine.

ekohl:
Rebased

manifests/init.pp
```diff
@@ -26,33 +26,33 @@
 # }
 class nginx (
   ### START Nginx Configuration ###
-  $client_body_temp_path                                     = $::nginx::params::client_body_temp_path,
-  Boolean $confd_only                                        = false,
-  Boolean $confd_purge                                       = false,
-  $conf_dir                                                  = $::nginx::params::conf_dir,
+  Stdlib::Absolutepath $client_body_temp_path                = undef,
```
alexjfisher: `undef` isn't a valid default for the data type (and the default is overridden in the module data anyway).

ekohl:
Moved a few more parameters and added a few more data types.

wyardley:
I'm going to restart just in case, but looks like multiple tests are failing?

bastelfreak:
@ekohl the parameter ordering is still wrong

ekohl:
> @ekohl the parameter ordering is still wrong

Yes, that's why I kept the ```= undef``` in the initial version. I don't have much time to complete this.

bastelfreak:
hey @ekohl, are you still interested in this?

ekohl:
I think this is still a good idea but I don't have time to finish this. I've also come around on the common.yaml and I now prefer defaults where possible in the .pp file because of puppet-strings.

ekohl:
I don't have time to finish this.

PR #1156, opened 2017-11-19
regenerate puppet-strings docs

bastelfreak:


bastelfreak:
We should do this during the release process

PR #1159, opened 2017-11-21
Use $service_name for service resource title.

### LINKED ISSUE #1158: Service resource name conflicts with system service

- Puppet: 5.2
- Ruby: 2.3.3
- Distribution: Debian 9
- Module version: master

```
class { 'nginx':
        service_manage  => true,
        service_name    => "maverick-nginx",
    }
```

```
service { "nginx":
 ensure => stopped,
 enable => false
}
```

## What are you seeing
Because the puppet-nginx service manifest calls `service { 'nginx':`, if you want to use puppet-nginx to manage a custom nginx service (eg. 'maverick-nginx' in this instance), you cannot also control the default system nginx service (eg. to turn it off as in the fragment above).  This is a perfectly valid scenario.

```
Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Cannot alias Service[system-nginx] to ["nginx"] at /srv/maverick/software/maverick/manifests/init.pp:205; resource ["Service", "nginx"] already declared at /srv/maverick/software/maverick/manifests/puppet-modules/nginx/manifests/service.pp:53 at /srv/maverick/software/maverick/manifests/init.pp:205:13  at /srv/maverick/software/maverick/manifests/maverick-modules/maverick_web/manifests/nginx.pp:26 on node maverick-up.home
```

Control of both services should be possible.

fnoop:
This allows to specify a custom service name such as 'custom-nginx', while also separately controlling the system nginx service outside of puppet-nginx module.

Closes #1158

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -150,7 +150,7 @@ class nginx (
   $service_ensure                                            = running,
   $service_flags                                             = undef,
   $service_restart                                           = undef,
-  $service_name                                              = "nginx",
+  $service_name                                              = 'nginx',
   $service_manage                                            = true,
   ### END Service Configuration ###
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -150,7 +150,7 @@ class nginx (
   $service_ensure                                            = running,
   $service_flags                                             = undef,
   $service_restart                                           = undef,
-  $service_name                                              = undef,
+  $service_name                                              = "nginx",
   $service_manage                                            = true,
   ### END Service Configuration ###
 
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -40,7 +40,7 @@ class nginx::service(
   if $service_manage {
     case $::osfamily {
       'OpenBSD': {
-        service { 'nginx':
+        service { 'puppet_nginx':
           ensure     => $service_ensure_real,
           name       => $service_name,
           enable     => $service_enable,
@@ -50,7 +50,7 @@ class nginx::service(
         }
       }
       default: {
-        service { 'nginx':
+        service { 'puppet_nginx':
           ensure     => $service_ensure_real,
           name       => $service_name,
           enable     => $service_enable,
@@ -63,7 +63,7 @@ class nginx::service(
 
   # Allow overriding of 'restart' of Service resource; not used by default
   if $service_restart {
-    Service['nginx'] {
+    Service['puppet_nginx'] {
       restart => $service_restart,
     }
   }
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -40,9 +40,8 @@ class nginx::service(
   if $service_manage {
     case $::osfamily {
       'OpenBSD': {
-        service { 'puppet_nginx':
+        service { $service_name:
           ensure     => $service_ensure_real,
-          name       => $service_name,
           enable     => $service_enable,
           flags      => $service_flags,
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -150,7 +150,7 @@
   $service_ensure                                            = running,
   $service_flags                                             = undef,
   $service_restart                                           = undef,
-  $service_name                                              = undef,
+  $service_name                                              = "nginx",
```
ekohl: This should be in single quotes

manifests/service.pp
```diff
@@ -40,7 +40,7 @@
   if $service_manage {
     case $::osfamily {
       'OpenBSD': {
-        service { 'nginx':
+        service { 'puppet_nginx':
```
ekohl: I wonder if we should just use ```$service_name``` and drop the ```name => $service_name```.

manifests/init.pp
fnoop: Why is that - for security, to stop string evaluation?

manifests/service.pp
fnoop: Same thing as name defaults to the resource name, just more explicit.  Would be tidier that way, I agree.

ekohl: IMHO the biggest benefit is that there's no risk of overlapping resource names for different services.

manifests/init.pp
ekohl: It's a lint rule we have. Strings without variables should be in single quotes. I think it was copied from ruby.

manifests/service.pp
fnoop: Agreed, updated with https://github.com/voxpupuli/puppet-nginx/pull/1159/commits/7e7dbf66deeae822ec3d99174516684b15d7be32

manifests/init.pp
ekohl: If you can fix this then the build will pass and I think it's good to go.

fnoop: Done with https://github.com/voxpupuli/puppet-nginx/pull/1159/commits/e557db0dc410d23525731aea15bd6530899395b4

manifests/service.pp
```diff
@@ -40,17 +40,16 @@
...
           flags      => $service_flags,
           hasstatus  => true,
           hasrestart => true,
         }
       }
       default: {
-        service { 'nginx':
+        service { $service_name:
           ensure     => $service_ensure_real,
           name       => $service_name,
```
ekohl: Oh, and this line can be removed too

fnoop: Oops.  https://github.com/voxpupuli/puppet-nginx/pull/1159/commits/ad41e15349a61d357d5c41ecb795385f148b3555

fnoop:
@alexjfisher Now the service name is dynamic, I don't know how to do that.  Sorry, I don't know how tests work.

ekohl:
It should be ```it { is_expected.to contain_service('nginx14') }```

fnoop:
Hurrah, travis passes now :)

alexjfisher:
@ekohl If I squash and merge, do you know what ends up in the commit message?

ekohl:
@alexjfisher Github shows a dialog when you squash where it prefills a commit message based on all the commits but you can edit it.

alexjfisher:
Thanks @fnoop

PR #1161, opened 2017-11-24
Add advanced nginx mapping support

desertkun:
So map key/values can be defined in different places.

This example is self-explanatory:
```
nginx::resource::map { 'backend_pool':
  ensure    => present,
  hostnames => true,
  default   => 'ny-pool-1,
  string    => '$http_host'
}

# in some classes far away from map
nginx::resource::map::entry { 'backend_pool_nyc':
  map => 'backend_pool',
  key => '*.nyc.example.com',
  value => 'ny-pool-1'
}

nginx::resource::map::entry { 'backend_pool_sf':
  map => 'backend_pool',
  key => '*.sf.example.com',
  value => 'sf-pool-1'
}
```
This PR makes `mappings` argument optional, so if defined, the map will use the hash instead, otherwise it would expect for appropriate `nginx::resource::map::entry` definitions.


manifests/resource/map/entry.pp
```diff
@@ -0,0 +1,82 @@
...
+    fail("Map ${map} has \$mappings defined, cannot use entry.")
+  }
+
+  concat::fragment { "${::nginx::conf_dir}/conf.d/${map}_${key}":
+    target => "${::nginx::conf_dir}/conf.d/${map}-map.conf",
+    content => "  ${key} ${value};\n",
+    order => "5_${order}"
+  }
+
+}
```
bastelfreak: can you please add the missing newline?

manifests/resource/map/entry.pp
```diff
@@ -0,0 +1,82 @@
...
+#    value => 'sf-pool-1',
+#    order => '1'
+#  }
+#
+
+define nginx::resource::map::entry (
+  String $map,
+  String $key,
+  String $value,
+  String $order                = $key
```
bastelfreak: you can align the = directly to $order

templates/conf.d/map/header.erb
```diff
@@ -0,0 +1,11 @@
+# MANAGED BY PUPPET
```
bastelfreak: can you rewrite this to an epp template?

templates/conf.d/map/footer.erb
```diff
@@ -0,0 +1 @@
+}
```
bastelfreak: do we really need a template for this? You can directly set the `}` in the concat::fragment resource.

bastelfreak:
Hi @desertkun, thanks for the great PR!
I added a few inline comments, can you take a look at them? Can you also add unit tests to verify that this is working?

templates/conf.d/map/header.erb
ekohl: I'd disagree with this. All other templates are ERB and I'd prefer to keep the templating language used in a module consistent. Rewriting everything to EPP is fine by me, but should happen in another PR for every template.

desertkun:
I have added missing newlines and removed the `footer.erb`. As for the unit tests, it's the thing I am not very familiar with. Are you implying I should add a complete example using this feature in `examples` folder?

bastelfreak:
Hi @desertkun, are you still interested in this PR? If so, please rebase it. We currently use beaker to run acceptance tests, please checkout https://github.com/voxpupuli/puppet-nginx/tree/master/spec/acceptance for examples. If you need a walk-through, please ping us at [slack](http://slack.puppet.com) or on our IRC channel #voxpupuli on freenode.

bastelfreak:
Hi. I'm going to close this PR due to inactivity. If you are still interested in it please reopen and rebase.

PR #1163, opened 2017-12-10
Compatibility with puppetlabs-apt 4.4.0

ekohl:
It's better to have this handled in puppetlabs-apt since this will also work for Debian 10 which has removed the apt-transport-https package.


wyardley:
Looks good to me. Noting here that this essentially reverts #1110, just so the issues are linked.

PR #1165, opened 2017-12-21
#1164 fix the member_max_fails missing param

### LINKED ISSUE #1164: Upstream member:  member_max_fails param is missing

- Puppet: 5.3.3-1
- Ruby:  -
- Distribution: Debian Jessie
- Module version: 0.9.0

Call the `nginx::resource::upstream::member` with the parameter:  `upstream_max_fails` 

## What are you seeing

Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Nginx::Resource::Upstream::Member[test]: has no parameter named 'upstream_max_fails' at 

Setting the upstream member to the right max fails counter.

the parameter is already available in the member template.
it's only missing in the manifest. 
I've created a PR.

khaefeli:
fix #1164

### DIFF
--- a/manifests/resource/upstream/member.pp
+++ b/manifests/resource/upstream/member.pp
@@ -14,6 +14,7 @@
 #   [*server*]                  - Hostname or IP of the upstream member server
 #   [*port*]                    - Port of the listening service on the upstream member
 #   [*upstream_fail_timeout*]   - Set the fail_timeout for the upstream. Default is 10 seconds
+#   [*upstream_max_fails*]      - Set the max_fails for the upstream member. Default is to use nginx default value which is 1.
 #
 #
 # Examples:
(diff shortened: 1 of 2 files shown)


manifests/resource/upstream/member.pp
```diff
@@ -39,7 +40,8 @@
   $server,
   Enum['present', 'absent'] $ensure = 'present',
   Integer $port                     = 80,
-  $upstream_fail_timeout  = '10s',
+  $upstream_fail_timeout            = '10s',
```
juniorsysadmin: Given these are new, can you use data types?

khaefeli: done

joekohlsdorf:
I don't agree with the use of String datatypes for time values, I suggest the following:
```Optional[Variant[Integer,Pattern[/^\d+[smhdwMy]?$/]]]```

bastelfreak:
Hi @khaefeli, thanks for the PR. Can you take a look at the comment from @joekohlsdorf? Can you please add a unit test?

saz:
I don't think this PR is required anymore: https://github.com/voxpupuli/puppet-nginx/blob/52ecfd7e5f08cf3c7046cc37a3af2fc069763732/manifests/resource/upstream/member.pp#L57

khaefeli:
closed. not needed anymore

PR #1169, opened 2018-01-10
Add more per-location proxy options: proxy_send_timeout, proxy_ignore…

merclangrat:
Added more options configurable per-location to proxy template and manifests:
- proxy_send_timeout
- proxy_ignore_header (nginx config option is proxy_ignore_headers, but I used 'proxy_ignore_header' name to be compatible with other options)
- proxy_next_upstream

All these options are supported by Nginx.


bastelfreak:
Thanks @merclangrat !

PR #1174, opened 2018-01-28
add missing autoindex parameter in template of server resource

joekohlsdorf:
The server resource has an ```autoindex``` parameter but the templates did not reflect this. There was also no testcase for this parameter.

### DIFF
--- a/spec/defines/resource_server_spec.rb
+++ b/spec/defines/resource_server_spec.rb
@@ -331,6 +331,12 @@ describe 'nginx::resource::server' do
               attr: 'index_files',
               value: [],
               notmatch: %r{\s+index\s+}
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @joekohlsdorf, thanks for the PR. Can you squash the commits into one?

joekohlsdorf:
Hi @bastelfreak, I squashed the commits into one.
PS for the future: You can also squash on merge, its a GitHub feature.

bastelfreak:
I'm aware of the feature, but unless it's something support urgent we want encourage contributors to do it. Many don't know how it works and we then are happy to provide assistance via slack or via our IRC channel.

PR #1180, opened 2018-02-13
Add support for dynamic modules.

sevencastles:
Allow nginx to load dynamic modules by specifying an absolute or a
relative ("modules/") path. Modules must be installed separately and
before loading the puppet nginx module.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,6 +33,7 @@ class nginx (
   Optional[Enum['on', 'off']] $daemon                        = undef,
   $daemon_user                                               = $nginx::params::daemon_user,
   $daemon_group                                              = undef,
+  Array[String] $dynamic_modules                             = [],
   $global_owner                                              = $nginx::params::global_owner,
   $global_group                                              = $nginx::params::global_group,
   $global_mode                                               = $nginx::params::global_mode,
--- a/spec/classes/nginx_spec.rb
+++ b/spec/classes/nginx_spec.rb
@@ -360,6 +360,12 @@ describe 'nginx' do
           describe 'nginx.conf template content' do
             [
               {
+                title: 'should not set load_module',
+                attr: 'dynamic_modules',
+                value: :undef,
+                notmatch: %r{load_module}
+              },
+              {
                 title: 'should not set user',
                 attr: 'super_user',
                 value: false,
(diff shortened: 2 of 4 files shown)


manifests/init.pp
```diff
@@ -33,6 +33,7 @@
   Optional[Enum['on', 'off']] $daemon                        = undef,
   $daemon_user                                               = $::nginx::params::daemon_user,
   $daemon_group                                              = undef,
+  Optional[Array[String]] $dynamic_module                    = undef,
```
bastelfreak: Can you update it so it defaults to an empty array?

spec/classes/nginx_spec.rb
```diff
@@ -337,6 +337,12 @@
 
           describe 'nginx.conf template content' do
             [
+              {
+                title: 'should not set load_module',
```
bastelfreak: can you add another test that sets a module and checks if it is present in the config file?

templates/conf.d/nginx.conf.erb
```diff
@@ -1,4 +1,13 @@
 # MANAGED BY PUPPET
+<% if @dynamic_module.is_a?(Array) -%>
```
bastelfreak: the variable is always an array, so you don't need this check

bastelfreak:
Hi @sevencastles, thanks for the PR! I added a few inline comments, please take a look at them. The used email address in your commit doesn't match the one you've configured in your github settings. Now the commit isn't associated with your account. Can you please fix that?

manifests/init.pp
ekohl: I'd also prefer it to be named ```dynamic_modules``` since it's an array.

sevencastles:
Hi, 
thanks a lot for your suggestions. I have applied all of them.

bastelfreak:
Thanks for the updates @sevencastles . Can you take a look at https://github.com/voxpupuli/puppet-nginx/pull/1180#discussion_r172012298 ?
Also the email address in the first commit isn't linked with your github account. You can squash those commits into one and reset the author. Some explanations: https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

spec/classes/nginx_spec.rb
sevencastles: This is done, check diff below.

sevencastles: This is done, check lines 891-910

sevencastles:
Hi @bastelfreak. I've squashed the commits into one. Also, I've (already) added the tests you've asked, if I have understood you right.

templates/conf.d/nginx.conf.erb
```diff
@@ -1,4 +1,11 @@
 # MANAGED BY PUPPET
+<% Array(@dynamic_modules).each do |mod_item| -%>
```
bastelfreak: Can you get rid of the Array()?
That isn't needed since it's always an array.

sevencastles: I will do. Sorry, I am not very good with ERB

sevencastles:
Hi,
do you have any news? I have implemented all the requested changes.

sevencastles:
I've updated my pull request to be compatible with the latest master. Can you have a look and merge it? @bastelfreak 

Thanks

manifests/init.pp
```diff
@@ -33,6 +33,7 @@
   Optional[Enum['on', 'off']] $daemon                        = undef,
   $daemon_user                                               = $nginx::params::daemon_user,
   $daemon_group                                              = undef,
+  Optional[Array[String]] $dynamic_modules                   = [],
```
alexjfisher: If not defaulting to `undef`, do you really need to use `Optional`?

PR #1182, opened 2018-02-26
add etag support at the http level

bryangwilliam:
This change adds support for enabling/disabling the etag nginx directive in the http context. http://nginx.org/en/docs/http/ngx_http_core_module.html#etag

If no value is supplied it will not add anything and allow the nginx-defined default to take effect.


bryangwilliam:
I had to amend the git log and so I closed and recreated a new PR with the changes. Apologies for the duplication.

PR #1183, opened 2018-02-26
add etag support at the http level

bryangwilliam:
This change adds support for enabling/disabling the etag nginx directive in the http context. http://nginx.org/en/docs/http/ngx_http_core_module.html#etag

If no value is supplied it will not add anything and allow the nginx-defined default to take effect.

### DIFF
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -96,6 +96,10 @@ http {
   lingering_timeout   <%= @lingering_timeout %>;
   tcp_nodelay         <%= @http_tcp_nodelay %>;
 
+<% if @etag -%>
+  etag <%= @etag %>;
+
+<% end -%>
 <% if @gzip == 'on' -%>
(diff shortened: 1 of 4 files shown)


templates/conf.d/nginx.conf.erb
```diff
@@ -96,6 +96,10 @@ http {
   lingering_timeout   <%= @lingering_timeout %>;
   tcp_nodelay         <%= @http_tcp_nodelay %>;
 
+<% if defined? @etag -%>
```
alexjfisher: `<% if @etag -%>` is fine here since etag is either `undef` or a `yes\no` string, not a boolean.

bryangwilliam: Sure thing. I just pushed a fix for this, along with a small formatting change for the whitespace so the config file will not get an extra newline when no value is set.

alexjfisher:
@bryangwilliam Thanks!

PR #1189, opened 2018-03-15
Add explicit parameter types to all parameters

JacobHenner:
This is a WIP to specify parameter datatypes where missing, and reconsider some potentially inconsistent types.


JacobHenner:
As I continue to work through this, I notice that there are a bunch of odd situations that should probably discussed. 

- Intermingling boolean and 'on'/'off' strings.
- strings for fields that could use integers
- The implicit use of Enum['opt 1', 'opt 2', false] rather than Optional[Enum['opt1 ', 'opt 2']
- Optional[Boolean] 

I'll add examples in contextual comments once I get my branch passing unit tests.

ekohl:
In https://github.com/voxpupuli/puppet-nginx/pull/1155 I started doing the same. Could be good to align the two.

JacobHenner:
Agreed. After I update my branch to pass all tests, I think it would be
best to gather consensus regarding the odd situations I mentioned, to
determine if the definition or behavior within the module needs to change.

ekohl:
For now I added the backwards-incompatible label because I don't know if any type changed. This may not be true and we could remove it later. I'd be happy to change some parameters so the types are consistent and parameters aligned.

> As I continue to work through this, I notice that there are a bunch of odd situations that should probably discussed.

I noticed those as well when I started to work on it. I'd be happy if we could be consistent.

JacobHenner:
Now that #1190 is fixed, it looks like all tests are passing. When I have some time (hopefully this weekend), I will comment on the cases that should be discussed.

JacobHenner:
One pattern I've noticed a bunch of times is the use of strings "on" and "off" to match nginx's config representation, rather than booleans. Do we want to standardize these values, and restrict these parameters to Enum["on","off"], or should a breaking change be introduced to set these as boolean parameters?

ekohl:
That's something I've struggled with as well. I believe there are option that accept ```on```, ```off``` and others which would argue in favor of the ```Enum``` solution. For now I've followed that, but I'd be open to other solutions.

JacobHenner:
Is there any consensus on whether Enums with ["on","off"], booleans, or the current hybrid of the two and the optional type should be used? Since this will be a breaking change, I want to make sure there's some agreement before standardizing it one way or another.

JacobHenner:
I believe this is ready for review and discussion.

manifests/init.pp
```diff
@@ -26,148 +26,148 @@
...
+  String[1] $client_body_temp_path                                              = $::nginx::params::client_body_temp_path,
+  Boolean $confd_only                                                           = false,
+  Boolean $confd_purge                                                          = false,
+  String[1] $conf_dir                                                           = $::nginx::params::conf_dir,
+  Optional[Nginx::Toggle] $daemon                                               = undef,
+  String[1] $daemon_user                                                        = $::nginx::params::daemon_user,
+  Optional[String[1]] $daemon_group                                             = undef,
+  String[1] $global_owner                                                       = $::nginx::params::global_owner,
+  String[1] $global_group                                                       = $::nginx::params::global_group,
+  Stdlib::Filemode $global_mode                                                 = $::nginx::params::global_mode,
```
bastelfreak: great that we finally have a datatype for this

bastelfreak: Can you, in a seperate PR, bump the stdlib requirement to at least 4.22.0 in metadata.json?

2018-04-21-205849_608x197_scrot

manifests/init.pp
```diff
@@ -26,148 +26,148 @@
...
+  String[1] $log_group                                                          = $::nginx::params::log_group,
+  Stdlib::Filemode $log_mode                                                    = '0750',
+  Variant[String[1], Array[String[1]]] $http_access_log                         = "${log_dir}/${::nginx::params::http_access_log_file}",
+  Optional[String[1]] $http_format_log                                          = undef,
+  Variant[String[1], Array[String[1]]] $nginx_error_log                         = "${log_dir}/${::nginx::params::nginx_error_log_file}",
+  Nginx::ErrorLogSeverity $nginx_error_log_severity                             = 'error',
+  Variant[String[1], Boolean] $pid                                              = $::nginx::params::pid,
+  String[1] $proxy_temp_path                                                    = $::nginx::params::proxy_temp_path,
+  String[1] $root_group                                                         = $::nginx::params::root_group,
+  String[1] $run_dir                                                            = $::nginx::params::run_dir,
```
bastelfreak: We can you Stdlib::Absolutepath here?

manifests/resource/server.pp
```diff
@@ -300,10 +300,10 @@
   # Check to see if SSL Certificates are properly defined.
   if $ssl {
     if $ssl_cert == undef {
-      fail('nginx: ssl_cert must be set to false or to a fully qualified path')
+      warning('nginx: ssl enabled but ssl_cert undef')
```
bastelfreak: why did you change this? Doesn't a fail() make more sense?

manifests/resource/streamhost.pp
```diff
@@ -47,23 +47,23 @@
...
+  Stdlib::Port $listen_port                                   = 80,
+  Optional[String[1]] $listen_options                         = undef,
+  Boolean $ipv6_enable                                        = false,
+  Variant[Array[Stdlib::Ipv6], Stdlib::Ipv6] $ipv6_listen_ip  = '::',
+  Stdlib::Port $ipv6_listen_port                              = 80,
+  String[1] $ipv6_listen_options                              = 'default ipv6only=on',
+  Optional[String[1]] $proxy                                  = undef,
+  Nginx::Duration $proxy_read_timeout                         = $::nginx::proxy_read_timeout,
+  Nginx::Duration $proxy_connect_timeout                      = $::nginx::proxy_connect_timeout,
+  Array[String[1]] $resolver                                  = [],
```
bastelfreak: now resolver is a required parameter, it needs to contain a string, this wasn't the case before. Should we change it back?

JacobHenner:
Is there still interest in this?

bastelfreak:
There sure is! I raised a few inline questions within the code, can you take a look? In the meantime we also defined some guidelines for a consistent code style: https://voxpupuli.org/docs/#reviewing-a-module-pr

ekohl:
To get this merged I think it's easier if we split this up into manageable pieces. Would you be willing to submit a PR per file?

bastelfreak:
Ping @JacobHenner :)

ekohl:
Replacement in https://github.com/voxpupuli/puppet-nginx/pull/1413

PR #1198, opened 2018-03-24
add acceptance test to verify default values

bastelfreak:


ekohl:
You can add [port](http://serverspec.org/resource_types.html#port) to verify it's listening on port 80 with tcp.

bastelfreak:
@ekohl implemented
@oranenj fixed

PR #1209, opened 2018-04-24
Remove unnecessary class parameters from nginx::service

tdevelioglu:
Class `nginx::service` is private, declared once and should reference
variables only from the main class.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -13,32 +13,25 @@
 # Sample Usage:
 #
 # This class file is not called directly
-class nginx::service(
-  $service_restart = $nginx::service_restart,
-  $service_ensure  = $nginx::service_ensure,
-  $service_enable  = $nginx::service_enable,
-  $service_name    = $nginx::service_name,
-  $service_flags   = $nginx::service_flags,
-  $service_manage  = $nginx::service_manage,
-) {
+class nginx::service {
 
   assert_private()
 
-  if $service_manage {
+  if $nginx::service_manage {
     case $facts['os']['name'] {
       'OpenBSD': {
-        service { $service_name:
-          ensure     => $service_ensure,
-          enable     => $service_enable,
-          flags      => $service_flags,
+        service { $nginx::service_name:
+          ensure     => $nginx::service_ensure,
+          enable     => $nginx::service_enable,
+          flags      => $nginx::service_flags,
           hasstatus  => true,
           hasrestart => true,
         }
       }
       default: {
-        service { $service_name:
-          ensure     => $service_ensure,
-          enable     => $service_enable,
+        service { $nginx::service_name:
+          ensure     => $nginx::service_ensure,
+          enable     => $nginx::service_enable,
           hasstatus  => true,
           hasrestart => true,
         }
@@ -47,9 +40,9 @@ class nginx::service(
   }
 
   # Allow overriding of 'restart' of Service resource; not used by default
-  if $service_restart {
-    Service[$service_name] {
-      restart => $service_restart,
+  if $nginx::service_restart {
+    Service[$nginx::service_name] {
+      restart => $nginx::service_restart,
     }
   }
 }


ekohl:
I disagree with this change. The biggest benefit of explicit class parameters is that you can write tests without including the entire module but just the class by explicitly passing in parameters. This provides huge benefits in performance and allows writing real unit tests. IMHO every class should explicitly declare its variables and only use local scope vars but I do realize it's a lot of duplication.

binford2k:
Fwiw, it's pretty easy to mock out the variables in a spec even without passing them directly to the class. For example,

```ruby
   let(:pre_condition) {
        class nginx {
           $service_restart = true,
           $service_ensure  = 'installed',
           $service_name    = 'nginx',
           $service_flags   = undef,
           $service_manage  = true,
        }
        include nginx
    }
```

ekohl:
@binford2k that still includes the entire class which is very slow. Especially on big modules this really slows you down.

tdevelioglu:
It allows setting the same variables in two places though and in a private class even. But what I really dislike is the unnecessary duplicate lookups it causes.

bastelfreak:
I'm not really sure if we should go that path or not. @binford2k is there an official style that Puppetlabs suggests for private classes? @voxpupuli/collaborators how do you all think about this?

jhoblitt:
I generally agree with @tdevelioglu's PR comment.  I also conceded that sometimes sub-ideal patterns are a necessary concession to practical testing. *However*, I think that directly testing a private class is best avoided.  From a behavioral standpoint, the only concerns should be the [public] interface and the resulting resources in the catalog; what goes on inside the sausage factory should be considered fairly opaque.

wyardley:
I can see both sides of this; don't have a string feeling -- I've written and tested stuff both ways. I tend to lean towards taking them out, but it does make testing a bit trickier.

 _But_, I think `$nginx::foo` is now preferred over `$::nginx::foo`, so if we do go this way, we should at least use that.

ekohl:
> But, I think $nginx::foo is now preferred over $::nginx::foo, so if we do go this way, we should at least use that.

IMHO we should enforce this via a linter rule and explicitly disable the old style rule.

wyardley:
> I've been bitten by $foo::bar being ambiguous in site modules -- has the variable resolution logic changed to make the :: implicit?

Short answer, yes
Some discussion in https://github.com/voxpupuli/modulesync_config/pull/369

binford2k:
My preference is also to test the public interface only, except in specific edge cases. As a module developer, I should be free to change the implementation details as long as it maintains the public interface.

And for the edge cases, I'm quite ok with needing to write a precondition mock to enable the test.

LongLiveCHIEF:
I'm also on board with testing through the public interface. Regardless of how quick/slow it is,it's best to test the way implementors implement.

I actually like this PR. @bastelfreak and I were having some discussion about redundant parameter validation inside a private class, and this practice completely eliminates the problems behind our discussion.

:thumbsup:

bastelfreak:
I agree with @binford2k and @dhollinger here. to quote one of my teachers from university 'only test the public interface you provide to your users. (Also try to test each different failure case, not success case)'.

ekohl:
I still see this class as a private class but I have applied this pattern because it allows unit testing. Testing the public interface is integration testing. Ideally you do both if there's complex logic in a class. The Puppet testing stack doesn't allow me to do that but this pattern is a workaround. We've started to apply this pattern in some places when our Travis tests started to take 2 hours and were getting killed.

If someone showed me how I could mock out the main class so it's just a data container then this pattern would be redundant

bastelfreak:
@ekohl isn't that what you are looking for? https://github.com/voxpupuli/puppet-nginx/pull/1209#issuecomment-383990224

bastelfreak:
The majority of the people seem to vote for removing the params. I will leave this open until Tuesday evening for more feedback. In my opinion we can merge this if there are no new objections.

ekohl:
@bastelfreak if that would work, that'd be great but in the past I already tried that. Just to check I just retried it in another module:

```
Class 'foreman' is already defined (line: 2); cannot redefine (file: /home/ekohl/dev/puppet-foreman/spec/fixtures/modules/foreman/manifests/init.pp, line: 230)
```

In general I don't particularly care what other modules do and I'd be fine with either solution, but I am arguing here in favor of better testability of modules. Especially when you apply the rspec-puppet-facts pattern of testing every class with all supported operating systems you quickly explode in testing time. patterns like these help you to reduce that.

LongLiveCHIEF:
@ekohl the `nginx::service` class should not be unit tested outside of the `nginx` module, since it should only ever be called _from_ the `nginx` module.  That's the very definition of a private class.

There's no need to _unit_ test another module from your own module. You test these concerns by integration/behavioral testing of _your_ module, combined with unit tests of _your_ module that confirm the right _public_ attributes are sent to the resources you've included in your modules manifests.

If you feel like there is a test that is missing for `nginx::service`, then you should add that unit test as a PR to _this_ module.

ekohl:
@LongLiveCHIEF I think you you're missing ym point. There are cases where I want to test a private class inside my own module. I want to unit test `nginx::service` **inside** this module. Sometimes you do have complex logic or want to test a combination of parameters and ignore the many other classes.

Note that when I talk about `nginx::service` it's just an example. This pattern can be applied to many modules.

Let me state it another way: I know of no other way than this pattern to properly unit test my own private classes.

dhollinger:
@ekohl Yet with your example, you are now moving outside of the intended functionality of that class and module. If that class is already tested within it's own development process, it's redundant to test it again within yours.... unless you've changed the functionality or using it in an unintended way.

There is such a thing as too much testing. We test `nginx::service` here with Unit and Beaker testing (through the main init) which validates that the class does what it is intended to do. At that point, all your tests need to do is validate your code and use your integration/beaker tests to validate that the service resource is placed as expected. No need to re-test what's already been tested.

ekohl:
> @ekohl Yet with your example, you are now moving outside of the intended functionality of that class and module. If that class is already tested within it's own development process, it's redundant to test it again within yours.... unless you've changed the functionality or using it in an unintended way.

Just to be clear, the way I've been talking about `nginx::service` here is as if I was a developer working on the `nginx` module and writing tests inside the `nginx` module. At no point did I think about writing anything from another module that uses `nginx::service` directly.

> There is such a thing as too much testing. We test nginx::service here with Unit and Beaker testing (through the main init) which validates that the class does what it is intended to do. At that point, all your tests need to do is validate your code and use your integration/beaker tests to validate that the service resource is placed as expected. No need to re-test what's already been tested.

Any my point was that it can be easier to unit test a class without all the other classes.

IMHO you can't unit test a class through init, that's an integration test. I see beaker tests as smoke tests. They all have their value.

My point has been that the pattern of moving all parameters to local scope variables allows you to make a class standalone which allows unit testing. Otherwise you always have to include other classes. It removes the need to go through the full matrix of operating systems for that particular class. You can still verify the OS behavior in an integration test through init.

LongLiveCHIEF:
> my point was that it can be easier to unit test a class without all the other classes

The problem with that technique is that the class is _never_ called by itself, so you're unit testing under conditions that can literally _never_ happen, introducing the possibility of false positives into your tests.

I think we're ready to merge this at this point, right?

LongLiveCHIEF:
Looks like all we're waiting on is changing `$::foo` to `$foo`.

ekohl:
> The problem with that technique is that the class is never called by itself, so you're unit testing under conditions that can literally never happen, introducing the possibility of false positives into your tests.

This is not true. Suppose you have a class with all variables that all inherit from a main class. Sometimes you just want to know if certain combinations of those variables end up with the correct result. You have simple if/else/case logic without complex inheritance, in other words a unit test. Testing it via the main class would be an integration test.

However, I guess I'm not going to convince anyone there's a gap in testing capabilities so I'll just give up. Because I'm going to need time to migrate away from Puppet I'm just going to do enough to keep my envs running but I'm not going to invest in getting infrastructure like that to run.

LongLiveCHIEF:
@ekohl the `nginx::service` class isn't just a private class, it's [contained within the `init.pp`](https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/init.pp#L181). This means it is essentially _part_ of _that_ class, and is only  it's own class as an organizational technique in puppet to group similar resources and treat them as one resource in the containing manifest (with pointers at the top and bottom of the contained class).

If `nginx::service` was a private class that _wasn't_ contained, and was instead `include`'ed by multiple other classes in this module, _then_ you are absolutely right in that it would be beneficial to test this class in isolation.

The nature of the `contain` however changes everything, making `nginx::service` less of a "class", and more like a distinct group of resources embedded in the modules main class. _That_ is why everyone here is insisting that it not be tested by itself. That particular class isn't a class because it's an entry point, it's a class for internal module organization and resource ordering purposes.

I realize now that this is a key distinction, and could have caused some of the frustration you've had in trying to express your point.

ekohl:
> I realize now that this is a key distinction, and could have caused some of the frustration you've had in trying to express your point.

It is one distinction, but I'd like to emphasize again that I don't care about this particular class. This is simple enough that unit testing doesn't add a lot of value. My point was about the bigger picture. A class like https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/proxydhcp.pp which uses a parameter to extract values from facts. That is pure logic that can be tested without all the other classes included. The also needs to be updated to support the new facts.

Of course one can argue that that class is more of a profile class inside a module and you'd probably be right.

> Please don't let one experience with individuals factor into your decision about what tools are best for managing your infrastructure.

Don't worry, it's been leading up to this for months. My biggest source of frustration is that puppetlabs' modules almost feel unmaintained and getting any change in takes so much effort.

bastelfreak:
Hey everybody. I created https://tickets.puppetlabs.com/browse/MODULES-7550 to get some official feedback from Puppet.

wyardley:
@bastelfreak FWIW, I have had some discussions with module team members about this before. IIRC, they leaned towards testing included or contained private classes via the main class, and referencing `$foo::bar` (for module foo) directly in the included classes.

jorhett:
So I think the key question is what is the module's public interface?

If the module allows / documents direct calls to child classes then each of them should be tested independently. If the module only provides a single public interface via the main class, then that's how it should be tested.

A private internal class has no responsibility for preserving its interface for public use. You can choose to test its interfaces, or you can choose to test only the public interface. That's your call.

IMHO the pattern of passing all variables into a class is important and useful when that class can be called directly. It avoids a loop of the child class having to include the parent class for testing purposes. If a class is an internal, private class there is absolutely zero need to allow it to function without the parent class dependencies, and it will GREATLY simplify testing and code maintainability. I've seen a 90% reduction in test artifacts without losing a simple bit of test coverage with that simplification.

IMHO the proposed change would greatly simplify testing and improve maintainability.

I don't know that @ekohl is still paying attention here, but if so I have some thoughts that might help you understand why what you're testing doesn't provide the value you are seeking. If you're not already on the #testing channel in Puppet community, let's talk about that there. It's a good conversation to have.

ekohl:
In https://github.com/voxpupuli/puppet-nginx/pull/1414 I go a step further. Since that removes almost all logic, there is no benefit to unit testing anymore.

PR #1210, opened 2018-04-25
Support setting `ssl_verify_depth` in nginx::resource::server

tdevelioglu:
This adds support for specifying the `ssl_verify_depth` option in server sections.


wyardley:
This looks good to me; small, sane, and has tests.

Ideally, we'd add the matching param to the mailhost section, but I'm going to merge this as-is.

PR #1213, opened 2018-05-10
migrate vars from topscope to relative scope

bastelfreak:


manifests/package/debian.pp
```diff
@@ -31,7 +31,7 @@
   }
 
   if $manage_repo {
-    include '::apt'
+    include 'apt'
```
alexjfisher: Stylewise... Do you prefer
```puppet
include 'apt'
```
or
```puppet
include apt
```

bastelfreak: I've no objections on this. In most cases I write it with the quotes.

PR #1211, opened 2018-05-03
Add explicit service_enable parameter, #1208

fnoop:
#### Pull Request (PR) description
Adds an explicit $service_enable parameter.  Sets to undef by default, which then follows the previous behaviour of auto-setting from $service_ensure.

#### This Pull Request (PR) fixes the following issues
Fixes #1208

PR #1217, opened 2018-05-17
Add service_enable and simplify service_ensure, #1208

### LINKED ISSUE #1208: nginx::service::service_enable does not exist

- Puppet: 5.5.0
- Ruby: 2.3.3p222
- Distribution: Raspbian (Debian stretch)
- Module version: 0.9.1-rc0

```
    class { 'nginx':
        service_manage  => true,
        service_ensure  => $service_ensure,
        service_enable  => $service_enable,
    }
```

## What are you seeing
```
Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Nginx]: has no parameter named 'service_enable' (file: /srv/maverick/software/maverick/manifests/maverick-modules/maverick_web/manifests/nginx.pp, line: 55, column: 5) on node maverick-raspberrylite.home
```

To control the service at boot

Adding the service_enable parameter to index.pp and service.pp makes it behave as expected

fnoop:
Adds an explicit $service_enable parameter. Sets to undef by default, which then follows the previous behaviour of auto-setting from $service_ensure.

This replaces a previous effort #1211

Fixes #1208

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -151,6 +151,7 @@ class nginx (
 
   ### START Service Configuation ###
   $service_ensure                                            = running,
+  $service_enable                                            = true,
   $service_flags                                             = undef,
   $service_restart                                           = undef,
   $service_name                                              = 'nginx',
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -16,6 +16,7 @@
 class nginx::service(
   $service_restart = $nginx::service_restart,
   $service_ensure  = $nginx::service_ensure,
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -151,6 +151,7 @@
 
   ### START Service Configuation ###
   $service_ensure                                            = running,
+  $service_enable                                            = true,
```
ekohl: Should this be a `Boolean` as well?

PR #1225, opened 2018-07-07
dont deploy "ssl on" on nginx 1.15 or newer

### LINKED ISSUE #1224: [warn] the "ssl" directive is deprecated, use the "listen ... ssl" directive instead
The `ssl on` syntax is deprecated since nginx 1.15. We should only add it if we use nginx 1.14 or older.

bastelfreak:
fixes #1224. the option 'ssl on' within a server block is deprecated since nginx 1.15.0.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -121,5 +121,11 @@ class nginx::params {
   $sites_available_group = $_module_parameters['root_group']
   $sites_available_mode  = '0644'
   $super_user            = true
+  if fact('nginx_version') {
+    # enable only for releases that are older than 1.15.0
+    $add_listen_directive = versioncmp(fact('nginx_version'), '1.15.0') < 0
+  } else {
+    $add_listen_directive = true
+  }
   ### END Referenced Variables
 }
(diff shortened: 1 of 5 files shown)


manifests/params.pp
```diff
@@ -121,5 +121,13 @@
   $sites_available_group = $_module_parameters['root_group']
   $sites_available_mode  = '0644'
   $super_user            = true
+  if fact('nginx_version') {
+    $add_listen_directive = versioncmp(fact('nginx_version'), '1.15.0') ? {
```
ekohl: Isn't this equal to `$add_listen_directive = versioncmp(fact('nginx_version'), '1.15.0') < 0`?

bastelfreak: I has some issues in the past with versioncmp() when the first param is undef, that's why I wrapped it within the if statement.

ekohl: I meant the `? { ...}` part. The full oneliner would be `$add_listen_directive = !fact('nginx_version') or versioncmp(fact('nginx_version'), '1.15.0') < 0` but that's not really readable.

PR #1228, opened 2018-07-08
add absolute_redirect support

bryangwilliam:
Add absolute_redirect directive support at all applicable levels
http://nginx.org/en/docs/http/ngx_http_core_module.html#absolute_redirect


bastelfreak:
Thanks for the patch @bryangwilliam !

PR #1230, opened 2018-07-09
Remove fact function and normalize using $facts

jabezk:
fact() was causing an error when using Puppet 5.5.2. 

N/A


ekohl:
Which error? `fact()` was introduced in stdlib 4.19.0 and should work.

PR #1231, opened 2018-07-10
initial support for snippets

### LINKED ISSUE #644: Same location on multiple vhosts
Hi.

I would like to have one site on several domains with every domain with it's own certificate. With current version, I have to duplicate all locations for every vhost.

It would then (probably) look something similar to:

```
'/':
  vhost:
    - vhost1
    - vhost2
  www_root: /var/www/my_host
```

Best regards.

### LINKED ISSUE #1135: Feature request: assign nginx location to multiple servers
Right now an nginx location config can only be assigned to one single nginx server passing it as a String. It would be nice to be able to define one nginx location and re-use it for multiple nginx servers. This could be done by passing an Array instead of a String. Here is an example of what I mean using hiera YAML config:  

```
nginx::nginx_locations:
  'acme-challenge':
    location: '/.well-known/acme-challenge'
    server: ['www.domain1.com', 'www.domain2.com']
```

This would avoid having to duplicate nginx location config blocks for multiple websites and hence have a shorter and cleaner config following DRY.

bryangwilliam:
Add basic support for creating snippets (reusable configuration files) that can be included by other resources. In phase 1 of this effort I am just introducing the ability to create include files by providing the raw contents, but future phases will include writing server::resource::location output into a snippet for reusability.

Please let me know your thoughts on the direction this is going.

Fixes #644
Fixes #1135

### DIFF
--- a/manifests/resource/snippet.pp
+++ b/manifests/resource/snippet.pp
@@ -1,0 +1,37 @@
+# This definition creates a reusable config snippet that can be included by other resources
+#
+# @param ensure Enables or disables the specified snippet
+# @param owner Defines owner of the .conf file
+# @param group Defines group of the .conf file
+# @param mode Defines mode of the .conf file
+# @param raw_content Raw content that will be inserted into the snipped as-is
+
+define nginx::resource::snippet (
+  String[1] $raw_content,
+  Enum['absent', 'present'] $ensure = 'present',
+  String $owner                     = $nginx::global_owner,
+  String $group                     = $nginx::global_group,
+  Stdlib::Filemode $mode            = $nginx::global_mode,
+) {
+  if ! defined(Class['nginx']) {
+    fail('You must include the nginx base class before using any defined resources')
+  }
+
+  $name_sanitized = regsubst($name, ' ', '_', 'G')
+  $config_file = "${nginx::snippets_dir}/${name_sanitized}.conf"
+
+  concat { $config_file:
+    ensure  => $ensure,
+    owner   => $owner,
+    group   => $group,
+    mode    => $mode,
+    notify  => Class['nginx::service'],
+    require => File[$nginx::snippets_dir],
+  }
+
+  concat::fragment { "snippet-${name_sanitized}-header":
+    target  => $config_file,
+    content => epp("${module_name}/snippet/snippet_header.epp", { 'raw_content' => $raw_content }),
+    order   => '001',
+  }
+}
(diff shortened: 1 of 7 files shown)


LongLiveCHIEF:
If we're going to add this in, we need to have a way where we can also ensure the snippets directory is `absent` when manage snippets is set to true.

We will also need tests added for this functionality.

LongLiveCHIEF:
Would you also need to notify the service for any new snippets added? This isn't a feature I've used from nginx before, so just asking. If yes, then you'll want to work that in before we pull this.

templates/snippet/snippet_header.erb
```diff
@@ -0,0 +1,8 @@
+# MANAGED BY PUPPET
```
LongLiveCHIEF: Any new templates added have to be `.epp`. Can you make this change?

LongLiveCHIEF:
> Any new templates added have to be `.epp`. Can you make this change?
Nevermind, you got it covered, i just missed it in my first read-through.

LongLiveCHIEF:
Ok, so a recap of changes needed before we can pull:

- [ ] remove `::nginx` and substitute with `nginx`, per current standards
- [ ] switch the new template introduced from an `erb` to an `epp` (again current standards)
- [ ] add a way for users to set `absent` when snippets dir is managed
- [ ] add tests
- [ ] add docs

manifests/resource/snippet.pp
```diff
@@ -0,0 +1,40 @@
+# define: nginx::resource::snippet
```
dhollinger: Rework the documentation comments to be `puppet-strings` compatible/compliant.

manifests/resource/snippet.pp
```diff
@@ -0,0 +1,40 @@
...
+  Stdlib::Filemode $mode            = $nginx::global_mode,
+) {
+  if ! defined(Class['nginx']) {
+    fail('You must include the nginx base class before using any defined resources')
+  }
+
+  $name_sanitized = regsubst($name, ' ', '_', 'G')
+  $config_file = "${nginx::snippets_dir}/${name_sanitized}.conf"
+
+  concat { $config_file:
```
dhollinger: Probably need a check in here to ensure that `$config_file` isn't already defined as a concat resource. If it is and you try to add another snippet to the same `$config_file`, it will fail when it tries to create this resource.

bastelfreak: I'm not sure about this. $config_file contains the name of the defined resource, so it's already unique in the catalog.

manifests/resource/snippet.pp
bastelfreak: updated it.

Issue #1222, opened 2018-07-02
nginx::resource::upstream make consistent use of nginx::resource::upstream::member


baurmatt:
We'd like to receive feedback on our plan to restructure the current handling of upstream members in this module.

# Current implementation
Two different ways to configure upstreams:
- Use `nginx::resource::upstream::member` and point the `upstream` parameter to the matching upstream name which then uses `templates/upstream/upstream_member.erb`
- Use nginx::resource::upstream parameters `members`/`upstream_fail_timeout`/`upstream_max_fails`/`upstream_context` which then uses `templates/upstream/upstream_members.erb`

# Suggested implementation
First: This is definitely a break change but should be easily migratable.

## `nginx::resource::upstream::member`
- Rename `upstream_fail_timeout` to `fail_timeout`
- Add all possible configuration options of a Nginx upstream member as defined in http://nginx.org/en/docs/http/ngx_http_upstream_module.html#server

## `nginx::resource::upstream`
- Remove `upstream_fail_timeout`/`upstream_max_fails`/`upstream_context` from `nginx::resource::upstream`
- Remove `templates/upstream/upstream_members.erb`
- Make `$members` of `nginx::resource::upstream` a Hash which gets later translated into a `nginx::resource::upstream::member` resource
- Add a `$member_defaults` parameter which can be used to set defaults for all configured `$members`
- Add all possible configuration options of a Nginx upstream as definied in http://nginx.org/en/docs/http/ngx_http_upstream_module.html as parameters of `nginx::resource::upstream`

Thanks in advance for taking the time to give us feedback!

/CC @SaschaDoering

PR #1233, opened 2018-07-16
Add parameters to upstream and upstreammembers

SaschaDoering:
##### Add parameters to upstream and upstreammembers

Add parameters to nginx::resource::upstream and
nginx::resource::upstream::member which allows more configuration on
upstreams as before. The only thing that we broke is that the
members of an upstream must now be passed as a hash rather than an
array. This also makes the sorting of keepalive to the end no
longer necessary because there is now a parameter for it. And
values for a nginx::resource::upstream::member can now be set as
default for all members of an upstream or individually for each
member inside the members hash. Of course, the explicit
specification overrides the defaults. In general the changes have
made more parameters available to nginx::resource::upstream and
nginx::resource::upstream::member. In addition, one of the two
templates for nginx::resource::upstream::member was disposed since
it is no longer needed.

Fixes GH-1222


SaschaDoering:
@juniorsysadmin Which kind of feedback do you need from me?

juniorsysadmin:
@SaschaDoering Nothing required from you just yet, but given this is a breaking change it would good to get some approval on the design from other users before merging.

SaschaDoering:
Thanks for the feedback, I thought something was missing from my side :)

baurmatt:
LGTM! :)

bastelfreak:
Is the only breaking change the new datatype for `$members`? Then it should be possible to implement this without backwards incompatible changes?

SaschaDoering:
Yes, this should be the only breaking change. In addition, it is possible to set different values for members of the upstream which is necessary for things like weight. I also don't see a reason for adding extra code for the old method given the _README.md_ states `Please note: This module is undergoing some structural maintenance. You may experience breaking changes between minor versions.` Especially since the changes to the configuration will be very minimal. On the other hand, the changes in the code would not be so minimal. Would it help to expand the _README.md_ with a migration example?

bastelfreak:
sure this is mentioned in the readme and we still didn't do a 1.0.0 release, but still backwards compatibility is always nice. Can you add backwards support or update the READMe.md?

manifests/resource/upstream.pp
```diff
@@ -18,82 +31,116 @@
...
-  $members_tag                              = undef,
-  Enum['present', 'absent'] $ensure         = 'present',
-  Optional[Hash] $upstream_cfg_append       = undef,
-  Optional[Hash] $upstream_cfg_prepend      = undef,
-  $upstream_fail_timeout                    = '10s',
-  $upstream_max_fails                       = undef,
-  Enum['http', 'stream'] $upstream_context  = 'http',
+  Enum['present', 'absent']                 $ensure                 = 'present',
+  Enum['http', 'stream']                    $context                = 'http',
+  Optional[Hash]                            $members                = undef,
```
bastelfreak: can this default to an empty hash instead of optional? We have some guidelines for datatypes at https://voxpupuli.org/docs/#reviewing-a-module-pr

> * Are there new params with datatype Hash or Array? If possible, they should default to empty Hash/Array instead of undef. You can also enforce the datastructure like Array[String]

and:

> * If you can supply one or multiple values for an attribute it’s common practice to enforce the datatype for one value and an array of that datatype. An example for string is Variant[String[1],Array[String[1]]]. This can be used in the Puppet code as [$var].flatten()

can you also define the datatypes within the hash?

manifests/resource/upstream.pp
```diff
@@ -18,82 +31,116 @@
...
+  Optional[String]                          $members_tag            = undef,
+  Optional[Nginx::UpstreamMemberDefaults]   $member_defaults        = undef,
+  Optional[String]                          $hash                   = undef,
+  Optional[Boolean]                         $ip_hash                = undef,
```
bastelfreak: Please check all parameters with the `Boolean` dataype. If it makes sense, remove the `Optional` and let them default to `True` or `False`

templates/upstream/upstream_footer.erb
```diff
@@ -1,10 +1,45 @@
-<% if @upstream_cfg_append %>
-<%# Slightly less obtuse way to sort but put keepalive at end -%>
-<%
-  @upstream_cfg_append = Hash[@upstream_cfg_append.sort]
-  @upstream_cfg_append['keepalive'] = @upstream_cfg_append.delete('keepalive')
--%>
-  <%- @upstream_cfg_append.each do |key,value| -%>
+<% if @hash -%>
```
bastelfreak: are you interested in turning this into an epp template? From our docs:

> * Is a new template added? The preferred language is [epp](https://puppet.com/docs/puppet/latest/lang_template_epp.html), not [erb](https://puppet.com/docs/puppet/latest/lang_template_erb.html)

it isn't required to turn this into an epp template, but it would be nice to have.

templates/upstream/upstream_member.erb
```diff
@@ -1 +1,12 @@
-  server     <%= @server %>:<%= @port %>  fail_timeout=<%= @upstream_fail_timeout %><% if @upstream_max_fails -%> max_fails=<%=@upstream_max_fails %><% end %>;
+  server <%= @_server %><% if @params_prepend -%> <%= @params_prepend %><% end -%>
```
bastelfreak: please convert this into an epp template.

manifests/resource/upstream.pp
SaschaDoering: You're right, I'll fix that.

manifests/resource/upstream.pp
SaschaDoering: You're right, again, I'll fix that, too.

templates/upstream/upstream_footer.erb
SaschaDoering: If I convert upstream_member.erb I will convert this too. It makes no sense to use different templatetypes for one file I think.

templates/upstream/upstream_member.erb
SaschaDoering: I'll convert the template.

SaschaDoering:
I pushed a new version a few moments ago which should fix the issues you commented. Please let me know if I missed something, thanks.

manifests/resource/upstream.pp
```diff
@@ -18,82 +33,121 @@
...
+  Optional[Nginx::UpstreamLeastTime]  $least_time             = undef,
+  Boolean                             $ntlm                   = false,
+  Optional[Integer]                   $queue_max              = undef,
+  Optional[Nginx::Time]               $queue_timeout          = undef,
+  Optional[String]                    $random                 = undef,
+  Optional[Stdlib::Unixpath]          $statefile              = undef,
+  Optional[Nginx::UpstreamSticky]     $sticky                 = undef,
+  Optional[Nginx::UpstreamZone]       $zone                   = undef,
+  Hash                                $cfg_append             = {},
+  Hash                                $cfg_prepend            = {},
```
bastelfreak: Can you enforce the datatypes within the hash please?

manifests/resource/upstream.pp
```diff
@@ -18,82 +33,121 @@
...
+  Boolean                             $ip_hash                = false,
+  Optional[Integer[1]]                $keepalive              = undef,
+  Optional[Integer[1]]                $keepalive_requests     = undef,
+  Optional[Nginx::Time]               $keepalive_timeout      = undef,
+  Boolean                             $least_conn             = false,
+  Optional[Nginx::UpstreamLeastTime]  $least_time             = undef,
+  Boolean                             $ntlm                   = false,
+  Optional[Integer]                   $queue_max              = undef,
+  Optional[Nginx::Time]               $queue_timeout          = undef,
+  Optional[String]                    $random                 = undef,
```
bastelfreak: For strings, please always use ```String[1]```, it will enforce a minimal string length of 1.

spec/acceptance/nginx_upstream_spec.rb
```diff
@@ -0,0 +1,113 @@
+require 'spec_helper_acceptance'
+
+describe 'nginx::resource::upstream define:' do
```
bastelfreak: acceptance tests \o/

spec/type_aliases/size.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe 'Nginx::Size' do
```
bastelfreak: please always add
```
it{ is_expected.to compile.with_all_deps }
```

manifests/resource/upstream.pp
SaschaDoering: I've added a datatype for this.

manifests/resource/upstream.pp
SaschaDoering: Should be fixed.

spec/type_aliases/size.rb
ekohl: Correct, you can leave that out.

bastelfreak: sorry, you're right.

bastelfreak:
looks good to me, thanks to the massive PR!  I will keep this open so we can get another review from somebody else.

SaschaDoering:
Ok, thanks for the review.

README.md
```diff
@@ -85,6 +94,38 @@ nginx::resource::mailhost { 'domain1.example':
...
+```
+
+From now on, the configuration must look like this:
+
+```puppet
+nginx::resource::upstream { 'puppet_rack_app':
+  members => {
+    'localhost:3000':
+      server => 'localhost'
+      port   => 3000
```
dhollinger: At first glance this can be a little confusing. My first question would be "Why not parse the `localhost:3000` into server and port?"

Could we potentially make it an array of hashes so that it looks more like:
``` puppet
nginx::resource::upstream { 'puppet_rack_app':
  members => [
    {
      server => 'localhost',
      port    => 3000,
    },
    {
      server => 'localhost',
      port    => 3001,
    },
    {
      server => 'localhost',
      port    => 3002,
    },
  ]
```
Does that work or does the parameter rely on each member to have a unique name?

manifests/resource/upstream.pp
```diff
@@ -18,82 +33,121 @@
...
-  $members_tag                              = undef,
-  Enum['present', 'absent'] $ensure         = 'present',
-  Optional[Hash] $upstream_cfg_append       = undef,
-  Optional[Hash] $upstream_cfg_prepend      = undef,
-  $upstream_fail_timeout                    = '10s',
-  $upstream_max_fails                       = undef,
-  Enum['http', 'stream'] $upstream_context  = 'http',
+  Enum['present', 'absent']           $ensure                 = 'present',
+  Enum['http', 'stream']              $context                = 'http',
+  Nginx::UpstreamMembers              $members                = {},
```
dhollinger: Same comment as above

README.md
SaschaDoering: Mainly cause Unix sockets are accepted too. We would first have to see if it is a valid Unix socket and only if not split at the colon. After that we need to check the parts if they are valid. We almost have to implement the used datatype in code. And at the end we have the problem to name the resources for the defined type which is necessary for exporting upstream members. That's why we need a unique name here.
Should I change the README.md file and use keys that do not consist of server and port? Maybe it's a good idea to use the keys as default for the comment.

baurmatt:
Can we please get this merged or an update what changes should be made?

manifests/resource/upstream.pp
```diff
@@ -105,8 +159,24 @@
   }
 
   concat::fragment { "${name}_upstream_footer":
-    target  => "${nginx::conf_dir}/${conf_dir_real}/${name}-upstream.conf",
+    target  => "${conf_dir}/${name}-upstream.conf",
     order   => '90',
-    content => template('nginx/upstream/upstream_footer.erb'),
+    content => epp('nginx/upstream/upstream_footer.epp', {
```
juniorsysadmin: My personal preference is for no parameters for `epp()`, where possible.

bastelfreak: I highly prefer them, because that allows scoped variables with proper datatypes.

PR #1234, opened 2018-07-25
Update nginx.conf.erb - only included files with the extension '.conf'

kristofvandam:
suffix the 'to include' files. I expect only *.conf files to be configuration files for nginx.

Usecase:

troubleshooting and 'disabling' the file quickly by adding .bck for example (vhost.conf.bck) so this vhost (or stream) is ignored.

### DIFF
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -217,7 +217,7 @@ http {
 
   include <%= @conf_dir %>/conf.d/*.conf;
 <% unless @confd_only -%>
-  include <%= @conf_dir %>/sites-enabled/*;
+  include <%= @conf_dir %>/sites-enabled/*.conf;
 <% end -%>
 }
 <% if @mail -%>
@@ -230,7 +230,7 @@ stream {
   <%-# conf.stream.d gets included either way if $stream is enabled -%>
   include <%= @conf_dir %>/conf.stream.d/*.conf;
 <% unless @confd_only -%>
-  include <%= @conf_dir %>/streams-enabled/*;
+  include <%= @conf_dir %>/streams-enabled/*.conf;
 <% end -%>
 }
 <% end -%>


LongLiveCHIEF:
Hey @kristofvandam, thanks for the PR!

As @ekohl mentioned, this is definitely a breaking change. 
As it stands, it would break way too many implementations, and it's a feature that's far too likely to have 3rd party dependencies for each user that's implemented it. Nginx is an enterprise level product, and unfortunately that means we have to be very cautious about introducing breaking changes.

 Look at the following common scenario:
- an enterprise uses a puppet profile to implement the nginx class, and to lay down conf files
- the names of these profiles can currently be more dynamic than the restriction you've added
- Implementors _may_ need to update their puppet profiles to implement this change
- Enterprise products like this typically have 3rd party tools to audit the state of machines, and they're looking for whatever the puppet profiles lay down, so now these have to be changed, or it will cause all kinds of errors with their audit tooling

This would have likely been a great feature for when the module was released, but at this point it's a _restrictive_ enhancement instead of a _progressive_ enhancement, and that means we have to reject it out of responsibility for our consumers.

In the end, I don't think the value outweighs the risk here. A string is a string to implementors, but modifying the tail of that string after the fact, or making them all change their existing strings is a disruption that will cause many of our customers months of grief.

LongLiveCHIEF:
@kristofvandam if you can figure out a way to write this as a progressive enhancement, and re-submit, I'd totally :thumbsup: it. You're reason for submitting it was :100:. There's nothing wrong functionally with what you submitted, it's just unfortunate that other circumstances impact the ability to accept this PR.

I would recommend maybe adding a `include_dir_extension` or something which defaults to `undef`, and then use a conditional in the template to add the desired behavior if the value of that isn't `undef`.

If you decide to go this route, make sure to add tests for this new param.

kristofvandam:
@LongLiveCHIEF , I totally get it, will see when I have time to write the extra param and tests.

feltra:
What about have a include_only param that by default follows the current '/*' and enables users to define its own if necessary ?

feltra:
@LongLiveCHIEF what if we iterate an array pre defined in params that could be overwritten ? 
  'sites_enabled_dirs'   => [ '/sites-enabled/*' ],
  'streams_enabled_dirs' => [ '/streams-enabled/*' ],

LongLiveCHIEF:
I'd vote to accept anything that isn't backwards incompatible.

PR #1238, opened 2018-08-09
start one worker process per core

bastelfreak:
Many people are confused by this default setting. The default of one
core is often a bottleneck. I would like to switch the default to the
amount of cores.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -136,7 +136,7 @@ class nginx (
   $types_hash_max_size                                       = '1024',
   Integer $worker_connections                                = 1024,
   Enum['on', 'off'] $ssl_prefer_server_ciphers               = 'on',
-  Variant[Integer, Enum['auto']] $worker_processes           = 1,
+  Variant[Integer, Enum['auto']] $worker_processes           = 'auto',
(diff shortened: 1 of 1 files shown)


ekohl:
Acceptance tests are very unhappy for unrelated reasons

alexjfisher:
IMO, this is a 'breaking change'. Even though they [recommend ](https://www.nginx.com/blog/tuning-nginx/)'auto', the upstream nginx default is still `1`. The change would at the very least need clearly documenting in the README.

Perhaps we should actually allow and default to `undef` and not write anything into nginx.conf (so that it 
 implicitly uses the default).  That way, we wouldn't have to make any changes if *they* decide to change it.

baurmatt:
For us, its totally valid to default to `auto` an mention this in the README as this is the recommended setting. But this is probably a larger discussion: Are Puppet module in general supposed to roll out:
1. The default of the software
2. The default of the OS package
3. Sane defaults known to be best practices

We always tend to 2. or 3. because 1. is for the majority of software not really usable in the real world.

bastelfreak:
I think switching to ```auto``` would be the best approach.

ekohl:
Btw, please do fix the tests before merging :)

PR #1243, opened 2018-09-03
Add mime.types file template and default values for it

### LINKED ISSUE #1240: Reliance on mime.types
Hey there,

excellent puppet module!

However, while experimenting with this module, I have stumbled upon a problem: the module blindly relies on the presence of a `mime.types` files (specifically in https://github.com/voxpupuli/puppet-nginx/blob/0ef72eb7af17a8255dae10ea1eb901b8894a8abf/templates/conf.d/nginx.conf.erb#L69 ). While this will work for most distributions, it does not work when using a custom config directory (`nginx::conf_dir`).

In my case, the configuration gets generated as follows (`nginx.conf`):
```
# MANAGED BY PUPPET
user www-data;
worker_processes 1;
worker_rlimit_nofile 1024;
...
http {

  include       /mnt/etc/nginx/mime.types;
  default_type  application/octet-stream;
  ...
  include /mnt/etc/nginx/conf.d/*.conf;
  include /mnt/etc/nginx/sites-enabled/*;
}
```

Since `/mnt/etc/nginx` was created by the module itself, it is initially empty and therefore lacks `mime.types`.

Upon startup, nginx then complains:
``` nginx -c /mnt/etc/nginx/nginx.conf 
nginx: [emerg] open() "/mnt/etc/nginx/mime.types" failed (2: No such file or directory) in /mnt/etc/nginx/nginx.conf:17
```

martialblog:
- The mine.types file can now be configured using a simple hash

When using this module to only configure nginx and then run in a Container, we ran into an issue when we tried to store the config in a different directory. The config directory can be adjusted using the conf_dir parameter, however the module does not provide a mime.types file.

This PR fixes this issue by adding a simple mime.types.erb that can be filled with a hash, and some default values for the file.

Fixes #1240


manifests/init.pp
```diff
@@ -147,6 +147,7 @@
   $package_source                                            = 'nginx',
   $package_flavor                                            = undef,
   $manage_repo                                               = $nginx::params::manage_repo,
+  $mime_types                                                = $nginx::params::mime_types,
```
bastelfreak: Please add a datatype for the new parameter. Probably ```Hash[String[1], String[1]]```

templates/conf.d/mime.types.erb
```diff
@@ -0,0 +1,6 @@
+# MANAGED BY PUPPET
+types {
+<% @mime_types.each do |key, value| -%>
+    <%= key %> <%= value %>;
+<% end -%>
+}
```
bastelfreak: please add a trailing newline

templates/conf.d/mime.types.erb
```diff
@@ -0,0 +1,6 @@
+# MANAGED BY PUPPET
```
bastelfreak: Please convert this to an [epp](https://puppet.com/docs/puppet/5.5/lang_template_epp.html) template.

martialblog:
@bastelfreak Hi, I'll take care of it

martialblog:
I addressed the issues, but the Beaker tests blew up O_o

templates/conf.d/mime.types.epp
```diff
@@ -0,0 +1,6 @@
+# MANAGED BY PUPPET
+types {
+<% $nginx::mime_types.each |$key, $value| { -%>
+    <%= $key %> <%= $value %>;
+<% } -%>
```
bastelfreak: remove the `-` I guess.

templates/conf.d/mime.types.epp
```diff
@@ -0,0 +1,6 @@
...
+
```
bastelfreak: Please do not an empty newline.

martialblog:
@bastelfreak Ok, thanks. I'm not too familiar with EPP Templates yet.

templates/conf.d/mime.types.epp
bastelfreak: To solve a bit of the confusion I might have created. A quote from our [review docs](https://voxpupuli.org/docs/#reviewing-a-module-pr):
> * Files should always terminate with a newline if possible, with an exception being file or template fragments like those used with concat. This is the [POSIX 
standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206), and some tools don't handle the lack of a terminating newline properly

martialblog:
@bastelfreak Hi, should be fine now. 

```
cat -A templates/conf.d/mime.types.epp                                                                              [11:14:57]
# MANAGED BY PUPPET$
types {$
<% $nginx::mime_types.each |$key, $value| { %>$
    <%= $key %> <%= $value %>;$
<% } %>$
}$
```

manifests/init.pp
```diff
@@ -147,6 +147,7 @@
   $package_source                                            = 'nginx',
   $package_flavor                                            = undef,
   $manage_repo                                               = $nginx::params::manage_repo,
+  Hash[String[1], String[1]] $mime_types                     = $nginx::params::mime_types,
```
bastelfreak: datatypes \o/

bastelfreak:
Thanks!

PR #1245, opened 2018-09-05
add initial REFERENCE.md

bastelfreak:


REFERENCE.md
```diff
@@ -0,0 +1,3823 @@
+# Reference
+<!-- DO NOT EDIT: This document was generated by Puppet Strings -->
+
+## Table of Contents
+
+**Classes**
+
+* [`nginx`](#nginx): Class: nginx  This module manages NGINX.  Parameters:  Actions:  Requires:  puppetlabs-stdlib - https://github.com/puppetlabs/puppetlabs-stdl
+* [`nginx::config`](#nginxconfig): Class: nginx::config  This module manages NGINX bootstrap and configuration  Parameters:  There are no default parameters for this class.  Ac
```
ekohl: This all appears to be limited to a certain width. IMHO we should fix that before including a REFERENCE.md. Looks like this is still the rdoc format instead of yardoc and I'm guessing puppet-strings chokes on it.

PR #1246, opened 2018-09-07
[WIP] Convert documentation to yardoc

ekohl:
This isn't perfect since the examples aren't using @example but it's a step in the right direction


manifests/package/redhat.pp
```diff
@@ -1,18 +1,4 @@
-# Class: nginx::package::redhat
-#
 # This module manages NGINX package installation on RedHat based systems
```
alexjfisher: 'module'?

manifests/init.pp
```diff
@@ -1,29 +1,136 @@
...
+# @param nginx_locations
+# @param nginx_locations_defaults
+# @param nginx_mailhosts
+# @param nginx_mailhosts_defaults
+# @param nginx_streamhosts
+# @param nginx_upstreams
+# @param nginx_servers
+# @param nginx_servers_defaults
+# @param purge_passenger_repo
+# @param add_listen_directive
```
alexjfisher: Are these just place holders for people to fill in?  ie if you left this out, wouldn't puppet-strings generate the same content from the parameter list below?

manifests/config.pp
```diff
@@ -1,18 +1,4 @@
...
-#
-# There are no default parameters for this class.
-#
-# Actions:
-#
-# Requires:
-#
-# Sample Usage:
-#
-# This class file is not called directly
```
alexjfisher: `@api private`?

manifests/config.pp
```diff
@@ -1,18 +1,4 @@
-# Class: nginx::config
-#
 # This module manages NGINX bootstrap and configuration
```
alexjfisher: `# @summary This module manages NGINX bootstrap and configuration`?

ekohl:
It looks like I missed some parts of the docs in the conversion.

bastelfreak:
@ekohl ping

PR #1249, opened 2018-09-28
Allow setting unquoted or custom flags on add_headers

itbm:
Allows any combination of add_headers quoting.
_This solves issues highlighted in #991 #992 #1020_ 

**Double Quotes:**
```
add_header => {
    'Access-Control-Allow-Origin'      => '*' , 
}
```
would produce
`  add_header              "Access-Control-Allow-Origin" "*";`

**Double Quotes and Unquoted**
```
add_header => {
    'Access-Control-Allow-Origin'      => {'*' => 'always'}, 
}
```
would produce
`  add_header              "Access-Control-Allow-Origin" "*" always;`

**Custom Quotes**
```
add_header => {
    'Access-Control-Allow-Origin'      => {'' => '\'*\' always'}, 
}
```
would produce
`  add_header              "Access-Control-Allow-Origin" '*' always;`

### DIFF
--- a/spec/defines/resource_location_spec.rb
+++ b/spec/defines/resource_location_spec.rb
@@ -404,17 +404,20 @@ describe 'nginx::resource::location' do
             let(:params) do
               default_params.merge(
                 'add_header' => {
-                  'X-Frame-Options' => 'SAMEORIGIN',
-                  'X-Powered-By'    => 'Puppet'
+                  'header 1' => 'test value 1',
+                  'header 2' => { 'test value 2' => 'tv2' },
+                  'header 3' => { '' => '\'test value 3\' tv3' }
                 }
               )
             end
 
-            it 'contains both add_header lines' do
+            it 'contains 3 add_header lines' do
               is_expected.to contain_concat__fragment('server1-500-' + Digest::MD5.hexdigest('location')).
-                with_content(%r{^\s+add_header\s+"X-Frame-Options"\s+"SAMEORIGIN";$})
+                with_content(%r{^\s+add_header\s+"header 1"\s+"test value 1";$})
               is_expected.to contain_concat__fragment('server1-500-' + Digest::MD5.hexdigest('location')).
-                with_content(%r{^\s+add_header\s+"X-Powered-By"\s+"Puppet";$})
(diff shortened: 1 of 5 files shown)


LongLiveCHIEF:
@itbm can you please update test specs to show these conditions? Otherwise, LGTM. Thanks!

alexjfisher:
nginx::location also has an `add_header` parameter. Does its template also need updating? https://github.com/voxpupuli/puppet-nginx/blob/master/templates/server/locations/headers.erb

itbm:
I have changed the sort.each and included the locations/headers.erb file in the others.

avillarreal83:
I wanted to share that when attempting use the add_header in a hiera role, the using the following syntax would add the header to Nginx but the header would not be seen by Nginx due to not specifying the _always_ setting in quotes:

**Initial Puppet settings** 
<code >
add_header:
       Strict-Transport-Security: 'max-age=31536000; includeSubDomains; always'
</code >

### Initial Nginx Results
<code >
add_header "Strict-Transport-Security" "max-age=31536000; includeSubDomains;" always";
</code >

When Nginx is restarted, the header is not recognized. In order to get Nginx configuration to recognize the "always" setting I had to add some extra double quotes to the puppet hiera role I had in place:

**Updated Puppet settings** 
<code >
    add_header:
       Strict-Transport-Security: 'max-age=31536000; includeSubDomains;" "always'
</code >
### Updated Nginx Results
<code >
add_header "Strict-Transport-Security" "max-age=31536000; includeSubDomains;" "always";
</code >

Wanted to share in case anyone else has runs into trying to use the _add_header_ option from the [server resource](https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/server.pp) and experiences this issue. Using the _add_header_ setting from this module was tricky, would be nice to make it easier to use the _always_ setting in future releases.

ajcollett:
Hi.
For all those Heira Users who see the above, here is a better way to do it, for the case above where you need to add a param after the string, and for long strings.
``` yaml
    add_header:
      Strict-Transport-Security: 
        'max-age=<your time>; includeSubDomains': always
      X-Frame-Options: deny
      X-Content-Type-Options: nosniff
      Content-Security-Policy: >-
        default-src 'self'; style-src 'self'; 
        style-src-elem; 
```

PR #1251, opened 2018-10-01
Propery handle ${client_body,proxy}_temp_path

ekohl:
The template already allowed setting it to false. This completes it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -26,7 +26,7 @@
 # }
 class nginx (
   ### START Nginx Configuration ###
-  $client_body_temp_path                                     = $nginx::params::client_body_temp_path,
+  Variant[Stdlib::Absolutepath, Boolean] $client_body_temp_path = $nginx::params::client_body_temp_path,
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -26,7 +26,7 @@
 # }
 class nginx (
   ### START Nginx Configuration ###
-  $client_body_temp_path                                     = $nginx::params::client_body_temp_path,
+  Variant[Stdlib::Absolutepath, Boolean] $client_body_temp_path = $nginx::params::client_body_temp_path,
```
ekohl: I wasn't sure about a boolean false here or an empty string. Thoughts?

alexjfisher: I think `Boolean` is better than empty string and it's probably not worth making it a breaking change by defaulting to `undef` and making it `Optional`.

ekohl: It does mean that true is allowed but not really handled - OTOH, it wasn't either in the past

PR #1254, opened 2018-10-01
Fix logging setup on Debian

ekohl:
This mitigates DSA-3701 / CVE-2016-1247 and matches the default setup on Debian.

The CVE mentioned Gentoo needs a similar treatment but I haven't looked into that yet.

https://www.debian.org/security/2016/dsa-3701
https://security-tracker.debian.org/tracker/CVE-2016-1247


alexjfisher:
Does this not fix the issue on RedHat systems too? (even if the RPMs never had this bug, did this module introduce it for RedHat users)?

ekohl:
I do not know. I just remembered this CVE when I saw this module was trying to change it.

spec/classes/nginx_spec.rb
```diff
@@ -360,9 +360,9 @@
             it do
               is_expected.to contain_file('/var/log/nginx').with(
                 ensure: 'directory',
-                owner: 'www-data',
+                owner: 'root',
                 group: 'adm',
-                mode: '0750'
+                mode: '0755'
```
dhollinger: This test is failing on RHEL-based tests.

ekohl: Looks like that's a valid failure because I copy-pasted an incorrect default

PR #1256, opened 2018-10-02
Introduce two new optional proxy parameters

### LINKED ISSUE #1176: Feature_request: Add proxy_max_temp_file_size and proxy_busy_buffers_size to parameter list

- Puppet:
- Ruby:
- Distribution:
- Module version:

ruriky:
We want to be able to manage the following (optional) configuration values:

[proxy_busy_buffers_size](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_busy_buffers_size)
[proxy_max_temp_file_size](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_max_temp_file_size)

With this PR user can modify the above values inside the location-block and also in the main nginx.conf.

Fixes #1176

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -93,6 +93,9 @@
 #   [*proxy_set_body*]        - If defined, sets the body passed to the backend.
 #   [*proxy_buffering*]       - If defined, sets the proxy_buffering to the passed
 #     value.
+#   [*proxy_max_temp_file_size*] - Sets the maximum size of the temporary buffer file.
+#   [*proxy_busy_buffers_size*] - Sets the total size of buffers that can be
+#     busy sending a response to the client while the response is not yet fully read.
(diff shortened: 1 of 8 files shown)


manifests/resource/location.pp
```diff
@@ -217,6 +220,8 @@
   Optional[String] $proxy_http_version                 = undef,
   Optional[String] $proxy_set_body                     = undef,
   Optional[Enum['on', 'off']] $proxy_buffering         = undef,
+  Optional[String] $proxy_max_temp_file_size           = undef,
```
bastelfreak: Can you please use `Optional[String[1]]` here please?

mmerfort: There is a size datatype which should be used in my opinion.

bastelfreak: good catch

bastelfreak:
I added the datatype and will merge it after travis turns green.

PR #1257, opened 2018-10-04
modulesync 2.1.0 and  allow puppet 6.x

bastelfreak:


.msync.yml
```diff
@@ -1 +1 @@
-modulesync_config_version: '2.0.0'
+modulesync_config_version: '2.1.0'
```
alexjfisher: Commit/PR title say modulesync 2.0.0

bastelfreak: title was wrong, I updated it

PR #1261, opened 2018-10-09
Archlinux: Set default log user to http

mauricemeyer:
This PR sets the default log user for Archlinux to `http`, the same user the daemon is running with.

The log user being set to `nginx` by default on Archlinux. That user does not exist (contrary to `http`), therefore Puppet wail fail to apply the ownership of the log directory.


ekohl:
Thanks!

PR #1266, opened 2018-10-17
Change gzip default to off and update tests

### LINKED ISSUE #1085: gzip is enabled by default
The module has  gzip is enabled by default, when someone configures to uses HTTPS without modifying this parameter they are susceptible to  www.breachattack.com.  

It would be more appropriate and safer for the default to be undef and take the default of the nginx module which is "off" when not configured. 

If someone wants to explicitly configure then it should be added.
This will avoid misconfiguration and opening users sites to the possible attack.

willrigling:
Closes #1085

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -77,7 +77,7 @@ class nginx (
   String $fastcgi_cache_max_size                             = '500m',
   Optional[String] $fastcgi_cache_path                       = undef,
   Optional[String] $fastcgi_cache_use_stale                  = undef,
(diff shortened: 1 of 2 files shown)


willrigling:
@alexfisher whoops thanks for the fix :)

alexjfisher:
Given that in upstream nginx, the default is 'off' and that there is a security risk in having it 'on', I'm happy with changing the default even though it's a breaking change.

@willrigling Do you want to amend your commit message with the spelling fixup too?

PR #1267, opened 2018-10-17
allow setting resolver at the http level

bryangwilliam:
Adding support for the resolver directive at the http level.
http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver

n/a

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -127,6 +127,7 @@ class nginx (
   Array $proxy_hide_header                                   = [],
   Array $proxy_pass_header                                   = [],
   Array $proxy_ignore_header                                 = [],
+  Optional[Nginx::Resolver] $resolver                        = undef,
   Enum['on', 'off'] $sendfile                                = 'on',
   Enum['on', 'off'] $server_tokens                           = 'on',
   Enum['on', 'off'] $spdy                                    = 'off',
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -111,6 +111,7 @@ class nginx::config {
   $proxy_set_header               = $nginx::proxy_set_header
   $proxy_hide_header              = $nginx::proxy_hide_header
   $proxy_pass_header              = $nginx::proxy_pass_header
+  $resolver                       = $nginx::resolver
   $sendfile                       = $nginx::sendfile
   $server_tokens                  = $nginx::server_tokens
   $spdy                           = $nginx::spdy
--- a/spec/classes/nginx_spec.rb
+++ b/spec/classes/nginx_spec.rb
@@ -574,10 +574,16 @@ describe 'nginx' do
                 match: %r{^\s+resolver 127.0.0.1;}
               },
               {
+                title: 'should set resolver with ipv6 address',
+                attr: 'resolver',
+                value: { 'addresses' => ['FF01::101'] },
+                match: %r{^\s+resolver \[FF01::101\];}
+              },
+              {
                 title: 'should set resolver with setting for ipv6',
                 attr: 'resolver',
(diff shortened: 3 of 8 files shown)


othalla:
I'm used to put this directive at server level but as Nginx doc says we can use it in http block, this PR seems fine.

manifests/init.pp
```diff
@@ -127,6 +127,7 @@
   Array $proxy_hide_header                                   = [],
   Array $proxy_pass_header                                   = [],
   Array $proxy_ignore_header                                 = [],
+  Array $resolver                                            = [],
```
ekohl: Should this be a `Array[String]` or can we even narrow it down to `Array[Stdlib::IP::Address]`?

bryangwilliam: I currently use strings to be able to add extra settings for "valid" and "ipv6" to the end of the directive, but that could (and probably should) be handled better. I can add a couple of extra params for controlling those instead which would allow us to put proper validation on the resolver array.

ekohl: We could start with a string and later make it `Variant[Stdlib::IP::Address, Struct[...]]`. I also realized we might need to escape the IPv6 addresses which you could do in a string yourself.

bryangwilliam: On second thought... the parameter should support address:port combinations if we are trying to match the nginx feature but stdlib doesn't seem to have a way to validate that combo. Any thoughts on how to handle that? I still think adding the extra params is better, but I don't believe doing so will allow us to narrow down the validation at all.

ekohl: You could use data types:

```puppet
type Nginx::Resolver = Variant[Array[Nginx::Address], Struct[
  ips               => Array[Nginx::Address],
  Optional['ipv6']  => Enum['on', 'off'],
  Optional['valid'] => Nginx::Time,
]]
type Nginx::Time = String // TODO: stricter validation
type Nginx::Address = Variant[
  Stdlib::IP::Address,
  Struct[
    ip   => Stdlib::IP::Address,
    port => Stdlib::Port,
  ],
]
```

We may want to allow simple arrays as well. The template to output this would be somewhat complicated though so you probably want to recast it in puppet code:
```puppet
if $resolver =~ Array {
  $_resolver = { 'ips' => $resolver }
} else {
  $_resolver = $resolver
}
```

bryangwilliam:
I added validation to the parameter as suggested using a struct and it works fine. I skipped IPv6 support for now since I don't know a good way to conditionally add square brackets to v6 addresses, but I'm ok with reworking it a little if you feel it is important to include and/or you have any suggestions for how to accomplish that.

bryangwilliam:
Latest commit includes support for IPv6, but the solution doesn't feel particularly elegant. I'd be interested in any thoughts on how it could be improved.

bastelfreak:
Converting them sounds pretty good, +1

bastelfreak:
Hi @bryangwilliam, are you still interested in this PR?

bastelfreak:
Hi. I'm going to close this PR due to inactivity. Please reopen it and rebase if you're still interested in it.

saz:
I'd like to work on this PR to get it integrated, but I find it pretty hard to do the conversion in a clean way, as the current implementation e.g. on server allows to add a IPv6 address.
Just splitting it on ':' won't work on such cases.

What if we allow passing in Array[String] which will stick to the old behavior and the variant implement in this PR?
This will avoid breaking things.

Is there already a type for IPv6 addresses?

ekohl:
There's `Stdlib::IP::Address::v6` and `Stdlib::IP::Address::v6::Nosubnet`

PR #1268, opened 2018-10-18
allow adding custom mime types while still using the module defaults

bryangwilliam:
Allows users to add custom mime types to mime.types without requiring them to manage the full list. It will append any custom types to the default list rather than completely overriding it.

n/a


othalla:
I'm used to put this directive at server level but as Nginx doc says we can use it in http block, seems fine ok.

othalla:
This is a bit weird, why not explicitely specify all Mime types?
It is a bit weird to have additionals values while setting the `$mime_types` variable.
I prefer to have a second var like `preserve_default_mime_types` of something like that.

bastelfreak:
I'm not sure if the merge approach is a good idea, this might confuse users. I agree with @othalla that this should be another variable or it should be handled in a profile.

bryangwilliam:
To clarify, are you saying that you would prefer passing in the customizations in the normal parameter and then have a second parameter that triggers a merge with the defaults if requested? I agree that seems like a better approach and have no problem switching, just want to make sure I understand before I make the changes.

othalla:
Yep that's what i mean :)

othalla:
This is better :) should be ok

PR #1275, opened 2018-11-09
Add autoindex to ssl_header too

bc-bjoern:
Hello,

the option autoindex was missing in server_ssl_header.erb. 

Thanks!

### DIFF
--- a/spec/defines/resource_server_spec.rb
+++ b/spec/defines/resource_server_spec.rb
@@ -437,28 +437,19 @@ describe 'nginx::resource::server' do
         end
 
         describe 'server_ssl_header template content' do
-         [    
-             {    
-               title: 'should set autoindex',
-               attr: 'autoindex',
-               value: 'on',
-               match: '  autoindex on;'
-             }    
-           ].each do |param|
-            context 'without a value for the nginx_version fact do' do
-               let :facts do
-                 facts[:nginx_version] ? facts.delete(:nginx_version) : facts
-               end  
-               let :params do
-                default_params.merge(
-                   ssl: true,
-                   ssl_key: 'dummy.key',
-                   ssl_cert: 'dummy.crt'
-                 )    
-               end  
- 
-               it { is_expected.to contain_concat__fragment("#{title}-ssl-header").with_content(%r{  ssl on;}) }
-            end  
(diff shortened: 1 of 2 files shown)


ekohl:
Could you add a test for this?

bc-bjoern:
May be...

bc-bjoern:
Okay, I can't. 
`   439         describe 'server_ssl_header template content' do
+  440           [
+  441             {
+  442               title: 'should set autoindex',
+  443               attr: 'autoindex',
+  444               value: 'on',
+  445               match: '  autoindex on;'
+  446             }
+  447           ]                                                                                                                                                                        
   448           context 'without a value for the nginx_version fact do' do
   449             let :facts do
   450               facts[:nginx_version] ? facts.delete(:nginx_version) : facts
   451             end
   452             let :params do
   453               default_params.merge(
   454                 ssl: true,
   455                 ssl_key: 'dummy.key',
   456                 ssl_cert: 'dummy.crt'
   457               )
   458             end
   459 
   460             it { is_expected.to contain_concat__fragment("#{title}-ssl-header").with_content(%r{  ssl on;}) }
   461           end
`

ekohl:
Feel free to keep the PR open and add commits to it. If you need help with writing a test we're happy to help.

bc-bjoern:
I do not understand this unfortunately. I guess I need to add it in spec/defines/resource_server_spec.rb in line 439 "describe server_ssl_header template content". But I do not understand how to do it correctly.

ekohl:
It would be very similar to:
https://github.com/voxpupuli/puppet-nginx/blob/ae2af805596703a08349b02d8e859d7125aaafa7/spec/defines/resource_server_spec.rb#L836-L841

That's where we test the SSL part of it. You can copy that one and modify it to autoindex instead of index_files.

bc-bjoern:
Sure, this part is clear more or less. 
```
           [    
             {    
               title: 'should set autoindex',
               attr: 'autoindex',
               value: 'on',
               match: '  autoindex on;'
             }    
           ] 
```
But I didn't managed it to include this into the server_ssl_header section.

bc-bjoern:
Travis says no

ekohl:
Thanks!

Issue #1274, opened 2018-11-08
Upstream members are never collected


linuxdaemon:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.10.10
- Ruby: 2.1.9p490 (2016-03-30 revision 54437) [x86_64-linux]
- Distribution: Ubuntu 16.04
- Module version: 0.15.0

### node1.example.com
```puppet
::nginx::resource::upstream { 'test_upstream': }

# Some server or location which proxies to `http://test_upstream`
```

### node2.example.com
```puppet
@@::nginx::resource::upstream::member {$fqdn:
  upstream => 'test_upstream',
  server => $::ipaddress,
  port => 80
}
```

## What are you seeing

Deployment fails on `node2` due to the upstream having no members specified.

## What behaviour did you expect instead

The upstream should collect all exported members when `members` is not set explicitly.

## Output log
```
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for bnc.snoonet.org
Info: Applying configuration version '1541692371'
Notice: /Stage[main]/Snoonet::Web::Sites::Anope/Nginx::Resource::Upstream[anope_servers]/Concat[/etc/nginx/conf.d/anope_servers-upstream.conf]/File[/etc/nginx/conf.d/anope_servers-upstream.conf]/ensure: defined content as '{md5}a20d79a0d0491d83667ee63e43648fad'
Info: Concat[/etc/nginx/conf.d/anope_servers-upstream.conf]: Scheduling refresh of Class[Nginx::Service]
Notice: /Stage[main]/Snoonet::Web::Sites::Anope/Nginx::Resource::Server[anope]/File[anope.conf symlink]/ensure: created
Info: /Stage[main]/Snoonet::Web::Sites::Anope/Nginx::Resource::Server[anope]/File[anope.conf symlink]: Scheduling refresh of Class[Nginx::Service]
Info: Class[Nginx::Service]: Scheduling refresh of Service[nginx]
Error: /Stage[main]/Nginx::Service/Service[nginx]: Failed to call refresh: Systemd restart for nginx failed!
journalctl log for nginx:
-- Logs begin at Sat 2018-11-03 06:40:01 MDT, end at Thu 2018-11-08 08:52:56 MST. --
Nov 08 08:52:55 bnc.snoonet.org systemd[1]: Stopping A high performance web server and a reverse proxy server...
[...]

PR #1276, opened 2018-11-12
Fix the condition for upstream members

SaschaDoering:
The condition which decides if we have upstream members configurred or
if we import from PuppetDB was wrong. Even if no upstream members are
defined the condition was true. As a result, you couldn't import upstream
members no matter what you configured. Sorry for that

Fixes GH-1274

### DIFF
--- a/manifests/resource/upstream.pp
+++ b/manifests/resource/upstream.pp
@@ -134,7 +134,7 @@ define nginx::resource::upstream (
     }),
   }
 
-  if $members != undef {
+  if ! empty($members) {
     $members.each |$member,$values| {
       $member_values = merge($member_defaults,$values,{'upstream' => $name,'context' => $context})
 


SaschaDoering:
Hi,

there seem to be problems installing Nginx, which is why some tests fail. I don't know how this should be related to the change I made. Also the master branch failed too, so I think this is another issue. So in my opinion, this problem is fixed. Or have I overlooked something here?

ekohl:
I think it'd be better if we made members an `Optional[...]` and default to undef. That way you have explicit control over if you want to specify members or use a collector. An alternative is an additional parameter to specify the behavior but that feels redundant.

SaschaDoering:
First, thanks for your reply. I think both solutions are valid. In both cases, I have the decision whether to define or collect the upstream members. You choose to collect in which you don't define a member, but I think that's independent of the datatype we're using. In this respect, I don't see any logical difference between the two solutions, or did I get something wrong?

ekohl:
No, you got it right: essentially there must be a choice for the module user between members and collection. Currently you can't.

I think that changing the parameter to `Optional[Nginx::UpstreamMembers] $members = undef,` restores the behavior from before f0bf83a1abac6c2ee7fe7257a37514319a96f0fa.

Could you also add a test that ensures this is possible so we don't accidentally break this again? Currently the tests are broken on a new GPG key on the passenger repository though.

SaschaDoering:
The first version of the commit https://github.com/voxpupuli/puppet-nginx/commit/f0bf83a1abac6c2ee7fe7257a37514319a96f0fa contained a hash with `undef` as default. This was changed in cause of this comment https://github.com/voxpupuli/puppet-nginx/pull/1233#discussion_r213726829. 
In addition, this contradicts the documentary on https://voxpupuli.org/docs/:
> Are there new params with datatype Hash or Array? If possible, they should default to empty Hash/Array instead of undef. You can also enforce the datastructure like Array[String].

But I agree, there should be a test for that. I'll try to create one

ekohl:
Note the `If possible`, but here additional functionality is overloaded in it. (cc @bastelfreak)

Dan33l:
I tested locally `master` branch with `centos7`. And i was able to reproduce failing run at same step with same messages : 
```
  Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install passenger' returned 1: Importing GPG key 0xD59097AB:
   Userid     : "packagecloud ops (production key) <ops@packagecloud.io>"
   Fingerprint: 418a 7f2f b0e1 e6e7 eabf 6fe8 c2e7 3424 d590 97ab
   From       : https://packagecloud.io/gpg.key
  
  
   One of the configured repositories failed (passenger repo),
   and yum doesn't have enough cached data to continue. At this point the only
   safe thing yum can do is fail.
```

ekohl:
I think we'll need the equivalent of https://github.com/puppetlabs/puppetlabs-apache/pull/1848

SaschaDoering:
Unfortunately, I didn't get to create a test for it. Because I would have to import exported resources, which, according to my information booth, is not possible. Or does anyone know a way to implement that?

SaschaDoering:
So, all tests are OK again. If nobody knows a way to implement a test with imported resources from somewhere I think I'm finished here.

Dan33l:
> I think we'll need the equivalent of [puppetlabs/puppetlabs-apache#1848](https://github.com/puppetlabs/puppetlabs-apache/pull/1848)

For easily reading history of the PR , the job was done with PR #1277

ekohl:
Thanks!

khaefeli:
Thanks for the fix. Could you please release it also to forge?
I lost much time because I only checked the open issues / PR and then started debugging my code, because I thought I'm on the latest 0.15.0 ;)

PR #1283, opened 2018-12-10
remove port from upstream member when service is defined

### LINKED ISSUE #1282: service upstream may not have port
The `service`  parameter of the `nginx::resource::upstream::member` enables port discovery via DNS SRV records. When it is used, a server port must not be specified.  Currently the module adds port 80 by default resulting in nginx failing to start.

- Puppet:all
- Ruby:all
- Distribution:all
- Module version: v0.15.0

```Puppet
nginx::resource::upstream { 'backend':
  zone    => $zone,
  ensure  => present,
  members => {
    'backend' => {
      'server'  => "upstream.endpoints.cluster.local",
      'service' => 'http',
      'resolve'  =>  true,
    }
  }
}
```
## What are you seeing
nginx startup fails because the upstream .conf file specifies the default port 80
```
upstream backend {
  server upstream.endpoints.cluster.local:80 service=http resolve;
}
```
The resulting upstream .conf file should not define a port
```
upstream backend {
  server upstream.endpoints.cluster.local service=http resolve;
}
```

## Output log
```
nginx: [emerg] service upstream may not have port
```

hbog:
The `service`  parameter of the upstream member enables port discovery
via DNS SRV records. When it is used, a server port must not be
specified or nginx will fail with the following error:
```
nginx: [emerg] service upstream may not have port
```

The PR removes the port from the server directive when the `service`parameter is set and updates the corresponding unit test.

Fixes #1282

### DIFF
--- a/manifests/resource/upstream/member.pp
+++ b/manifests/resource/upstream/member.pp
@@ -75,7 +75,7 @@ define nginx::resource::upstream::member (
     default  => "${nginx::config::conf_dir}/conf.d",
   }
 
-  $_server = ($server =~ Pattern[/^unix:\/([^\/\0]+\/*)*$/]) ? {
+  $_server = ($server =~ Pattern[/^unix:\/([^\/\0]+\/*)*$/] or $service != undef) ? {
     true  => $server,
     false => "${server}:${port}",
   }
--- a/spec/defines/resource_upstream_spec.rb
+++ b/spec/defines/resource_upstream_spec.rb
@@ -334,7 +334,7 @@ describe 'nginx::resource::upstream' do
               },
               {
                 value: { member1: { service: 'member1.backend' } },
-                match: 'member1:80 service=member1.backend;'
+                match: 'member1 service=member1.backend;'
               },
               {
                 value: { member1: { slow_start: '20s' } },


bastelfreak:
Hi @hbog, thanks for the change. Are you able to provide a tiny acceptance test for this? So we know for  sure that the generated config works. You can find some examples in https://github.com/voxpupuli/puppet-nginx/tree/master/spec/acceptance. Let us know if you need some help. We're reachable in the #voxpupuli IRC channel on freenode and on https://slack.puppet.com

hbog:
The `service` parameter for the `server` directive that I use is only available in the commercial offering [NGINX-Plus](https://www.nginx.com/products/nginx/).
Is there a license to add NGINX-plus to the acceptance tests ?
I think I could include dnsmasq in the test box to provide responses to the DNS SRV queries.

bastelfreak:
I would already be happy when we know that the generated config works and the daemon starts. IMO there is no need to test the actual SRV records. As mentioned in the other PR, I didn't know that this requires nginx-plus, so I assume simple acceptance testing won't be possible. Can you please update the docs with this new parameter? I think we can merge this without the acceptance tests.

bastelfreak:
Ping @hbog :)

TheMeier:
Closing due to lack of activity. Feel free to re-open.

PR #1290, opened 2018-12-18
Update smartos support

joelgarboden:
Update SmartOS support, which is in the Solaris family.

```
facter -p os
{"name"=>"SmartOS", "family"=>"Solaris", "release"=>{"minor"=>"11", "full"=>"5.11"}}
```

SmartOS is in the Solaris family, so was never reaching the `$facts['os']['family']` default case, but instead landing in Solaris.

In addition, `/usr` is read-only on SmartOS, but not Solaris.

Tested on a Base-64 17.4.0-LTS zone, initially setup by 0.9.0 version of this module.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -151,9 +151,21 @@ class nginx::params {
       }
     }
     'Solaris': {
-      $_module_os_overrides = {
-        'daemon_user'  => 'webservd',
-        'package_name' => undef,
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @joelgarboden, thanks for this PR! Any chance you could provide facts from smartos to https://github.com/camptocamp/facterdb ?

PR #1296, opened 2018-12-29
Automatically require SSL cert files in the server

### LINKED ISSUE #1295: Is there a way to only create an entry if the cert exists?
When using Dehydrated and Puppet Nginx, the Nginx config is updated before the cert files are there and Nginx crashes. 

Not so much expected to see, but wondering if it's possible to check for the required files first, and only if they are there would we created add the Nginx configs.

Thank you for working on this package!

ekohl:
If you define a file to create the SSL key and/or cert then these should
be in place before nginx uses them. This prevents writing out a
configuration that will not start.

Fixes #1295

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -445,7 +445,8 @@ define nginx::resource::server (
   if $ssl {
     # Access and error logs are named differently in ssl template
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
@ekohl I've no idea why this is broken :(

PR #1309, opened 2019-02-01
Fix `$nginx_upstream_defaults` type

saz:
In my understanding, $nginx_upstream_defaults should be of type UpstreamDefaults, not UpstreamMemberDefaults

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Nginx]: parameter 'nginx_upstream_defaults' unrecognized key 'member_defaults' at /etc/puppetlabs/code/environments/production/private_modules/profile/manifests/nginx.pp:53:3 on node host.example.com
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -180,7 +180,7 @@ class nginx (
   Hash $nginx_streamhosts                                 = {},
   Hash $nginx_streamhosts_defaults                        = {},
   Hash $nginx_upstreams                                   = {},
-  Nginx::UpstreamMemberDefaults $nginx_upstreams_defaults = {},
+  Nginx::UpstreamDefaults $nginx_upstreams_defaults       = {},
   Boolean $purge_passenger_repo                           = true,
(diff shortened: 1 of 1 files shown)


bastelfreak:
Could you add some tests, so we know that the issue is fixed?

alexjfisher:
@bastelfreak By simple inspection, I think we can be pretty sure this is correct though.

bastelfreak:
ah, looks like we both updated it @alexjfisher :D

alexjfisher:
@saz Thanks for finding this!  Sorry for the silly number of force pushes to your PR branch.  First I didn't know both @bastelfreak and me were simultaneously working on getting this merged.  I also hadn't noticed the parameter got renamed in https://github.com/voxpupuli/puppet-nginx/pull/1080! :)

Should be correct and merged now.  Thanks again.

PR #1314, opened 2019-02-12
Fixed varibale name and code style

aleksmark:
https://puppet.com/docs/puppet/5.4/lang_reserved.html#variables

> If the first character is an underscore, that variable should only be accessed from its own local scope; using qualified variable names where any namespace segment begins with _ is deprecated.


manifests/resource/upstream/member.pp
```diff
@@ -85,7 +85,7 @@
     target  => "${conf_dir}/${upstream}-upstream.conf",
     order   => 40,
     content => epp('nginx/upstream/upstream_member.epp', {
-      _server        => $_server,
+      real_server    => $real_server,
```
ekohl: On the left hand side it can just be `server`. Right hand side can still be `$_server`

PR #1318, opened 2019-02-26
add static gzip support

compermisos:
Add config to use "gzip_static" config (https://nginx.org/en/docs/http/ngx_http_gzip_static_module.html)

This PR not close open issues

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -72,6 +72,7 @@ class nginx::config {
   $gzip_proxied                   = $nginx::gzip_proxied
   $gzip_types                     = $nginx::gzip_types
   $gzip_vary                      = $nginx::gzip_vary
+  $gzip_static                      = $nginx::gzip_static
   $http_raw_prepend               = $nginx::http_raw_prepend
   $http_raw_append                = $nginx::http_raw_append
   $http_cfg_prepend               = $nginx::http_cfg_prepend
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -72,7 +72,7 @@ class nginx::config {
   $gzip_proxied                   = $nginx::gzip_proxied
   $gzip_types                     = $nginx::gzip_types
   $gzip_vary                      = $nginx::gzip_vary
-  $gzip_static                      = $nginx::gzip_static
+  $gzip_static                    = $nginx::gzip_static
   $http_raw_prepend               = $nginx::http_raw_prepend
   $http_raw_append                = $nginx::http_raw_append
   $http_cfg_prepend               = $nginx::http_cfg_prepend
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -86,7 +86,7 @@ class nginx (
   $gzip_proxied                                              = 'off',
   $gzip_types                                                = undef,
   Enum['on', 'off'] $gzip_vary                               = 'off',
-  Enum['on', 'off'] $gzip_static                             = 'off',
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -86,6 +86,7 @@
   $gzip_proxied                                              = 'off',
   $gzip_types                                                = undef,
   Enum['on', 'off'] $gzip_vary                               = 'off',
+  Enum['on', 'off'] $gzip_static                             = 'off',
```
ekohl: The manual suggests you can also use `always` as a valid value

manifests/config.pp
```diff
@@ -72,6 +72,7 @@
   $gzip_proxied                   = $nginx::gzip_proxied
   $gzip_types                     = $nginx::gzip_types
   $gzip_vary                      = $nginx::gzip_vary
+  $gzip_static                      = $nginx::gzip_static
```
ekohl: Could you fix the alignment?

manifests/init.pp
compermisos: added always

manifests/config.pp
compermisos: correct identation

templates/conf.d/nginx.conf.erb
```diff
@@ -110,6 +110,12 @@ http {
   etag <%= @etag %>;
 
 <% end -%>
+<% if @gzip_static == 'on' -%>
```
ekohl: The alternative is to use:
```erb
<% if @gzip_static != 'off' -%>
   gzip_static       <%= @gzip_static %>;
<% end -%>
```

alexjfisher: Or should it be `Optional` as the [ngx_http_gzip_static module](https://nginx.org/en/docs/http/ngx_http_gzip_static_module.html) might not be available?

> This module is not built by default, it should be enabled with the --with-http_gzip_static_module configuration parameter.

([`ngx_http_gzip_module`](http://nginx.org/en/docs/http/ngx_http_gzip_module.html) is built by default and provides the options `gzip_vary`, `gzip_types` and `gzip_proxied`)

compermisos: > Or should it be `Optional` as the [ngx_http_gzip_static module](https://nginx.org/en/docs/http/ngx_http_gzip_static_module.html) might not be available?
> 
> > This module is not built by default, it should be enabled with the --with-http_gzip_static_module configuration parameter.
> 
> ([`ngx_http_gzip_module`](http://nginx.org/en/docs/http/ngx_http_gzip_module.html) is built by default and provides the options `gzip_vary`, `gzip_types` and `gzip_proxied`)

The default value is off, the static module is different to normal gzip module, the normal module compress "on the fly", and static serve pre-compressed files

compermisos: > The alternative is to use:
> 
> ```
> <% if @gzip_static != 'off' -%>
>    gzip_static       <%= @gzip_static %>;
> <% end -%>
> ```

change made

ekohl:
Have you had time to look at writing a test?

dhoppe:
@compermisos It would be great, if you could provide some tests for the new parameter.

compermisos:
you can guide me to make some tests?

igalic:
you can find us in #voxpupuli on feenode or the outlet community slack.

in general, i would start by looking how the surrounding parameters are tested

saz:
As far as I can see, this PR isn't required anymore due to the merge of #1386

PR #1321, opened 2019-03-07
Allow setting a custom path for mime.types

jacksgt:
In certain environment the user may wish to use mime.types file from
a path other than $conf_dir.

For example when using `conf_dir => '/mnt/gluster/nginx'` and mounting the path into a container at `/etc/nginx`.
The module needs to know the 'real' configuration path because it has to write the files there, but at the same time it can't use this path to substitute the absolute path of the mime.types file, because it is at a different location in the container (here: `/mnt/gluster/nginx/mime.types` vs. `/etc/nginx/mime.types`)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -96,6 +96,7 @@ class nginx (
   $keepalive_requests                                        = '100',
   $log_format                                                = {},
   Boolean $mail                                              = false,
+  Variant[String, Boolean] $mime_types_path                  = 'mime.types',
   Boolean $stream                                            = false,
   String $multi_accept                                       = 'off',
   Integer $names_hash_bucket_size                            = 64,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -96,6 +96,7 @@
   $keepalive_requests                                        = '100',
   $log_format                                                = {},
   Boolean $mail                                              = false,
+  Variant[String, Boolean] $mime_types_path                  = 'mime.types',
```
ekohl:
```suggestion
  Variant[String[1], Boolean] $mime_types_path               = 'mime.types',
```

jacksgt: I have added a check to the nginx.conf.erb template

ekohl:
I'll let Travis continue testing, but it looks good to me. If anyone else sees it before I get back to it: just merge it :)

jacksgt:
Ping.

ekohl:
Gitlab has a feature 'merge when tests pass'. I miss that here because as you can see, I tend to forget.

PR #1325, opened 2019-04-18
Added support for new feature: port ranges

00ssh:
Updated listen_port data type from Integer to Variant[String, Integer] in 'nginx::resource::server' and 'nginx::resource::streamhost' as newer versions of Nginx support port range at listen port.
Please check the release notes from 26th of March: 
- http://nginx.org/en/CHANGES
- https://www.nginx.com/blog/nginx-plus-r18-released#port-ranges
```
Changes with nginx 1.15.10                                       26 Mar 2019

    *) Change: when using a hostname in the "listen" directive nginx now
       creates listening sockets for all addresses the hostname resolves to
       (previously, only the first address was used).

    *) Feature: port ranges in the "listen" directive.
...
```

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -5,7 +5,7 @@
 # Parameters:
 #   [*ensure*]                     - Enables or disables the specified server (present|absent)
 #   [*listen_ip*]                  - Default IP Address for NGINX to listen with this server on. Defaults to all interfaces (*)
-#   [*listen_port*]                - Default IP Port for NGINX to listen with this server on. Defaults to TCP 80
+#   [*listen_port*]                - Default IP Port for NGINX to listen with this server on. Defaults to TCP 80. It can be a port or a port range (eg. '8081-8085').
 #   [*listen_options*]             - Extra options for listen directive like 'default_server' to catchall. Undef by default.
 #   [*listen_unix_socket_enable*]  - BOOL value to enable/disable UNIX socket listening support (false|true).
 #   [*listen_unix_socket*]         - Default unix socket for NGINX to listen with this server on. Defaults to UNIX /var/run/nginx.sock
@@ -147,7 +147,7 @@
 define nginx::resource::server (
   Enum['absent', 'present'] $ensure                                              = 'present',
   Variant[Array, String] $listen_ip                                              = '*',
-  Integer $listen_port                                                           = 80,
+  Variant[Integer, String] $listen_port                                          = 80,
   Optional[String] $listen_options                                               = undef,
   Boolean $listen_unix_socket_enable                                             = false,
   Variant[Array[Stdlib::Absolutepath], Stdlib::Absolutepath] $listen_unix_socket = '/var/run/nginx.sock',
@@ -157,7 +157,7 @@ define nginx::resource::server (
   Array $location_deny                                                           = [],
   Boolean $ipv6_enable                                                           = false,
   Variant[Array, String] $ipv6_listen_ip                                         = '::',
-  Integer $ipv6_listen_port                                                      = 80,
+  Variant[Integer, String] $ipv6_listen_port                                     = 80,
   String $ipv6_listen_options                                                    = 'default ipv6only=on',
   Hash $add_header                                                               = {},
   Boolean $ssl                                                                   = false,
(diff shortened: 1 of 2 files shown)


00ssh:
Hi,
Can someone point to me, please, what is wrong with my travis tests? The error seems to be:
`An error occurred while loading ./spec/acceptance/class_spec.rb.
Failure/Error: install_module_dependencies_on(hosts)
Beaker::Host::CommandFailure:
  Host 'debian9-64-1' exited with 1 running:
   puppet module install puppetlabs-concat -v 5.3.0
  Last 10 lines of output were:
  	Notice: Preparing to install into /etc/puppetlabs/code/environments/production/modules ...
  	Notice: Downloading from https://forgeapi.puppet.com ...
  	Error: Could not install module 'puppetlabs-concat' (???)
  	  No version of 'puppetlabs-concat' can satisfy all dependencies
  	    Use `puppet module install --ignore-dependencies` to install only this module`

00ssh:
It might have been a forge issue. Now all checks have passed. Can you please take a look at might changes and let me know if it can be merged?

00ssh:
@ekohl @dhoppe can you please give me feedback on this change? we are eager to have this implemented. Thanks!

dhoppe:
I wonder if we should check the version of Nginx somewhere before configuring port ranges. We already did that for some templates, based on the fact ```nginx_version```.

manifests/resource/server.pp
```diff
@@ -316,9 +320,39 @@
...
+      $ipv6_port = $ipv6_listen_port_range
+    }
+    elsif ($ipv6_listen_port_range != undef) and ($port_range_support == false) {
+      fail('nginx: this version of nginx does not support port ranges (must be >= 1.15.10)')
+    }
+    else {
+      $ipv6_port = $ipv6_listen_port
+    }
+  }
+
```
dhoppe: @00ssh I think your code is to complicated and does not need to throw an error message. Maybe something like this could work:
```
  # If port range is defined, ignore any other $listen_port defined
  if versioncmp(fact('nginx_version'), '1.15.10') < 0 {
    $port = $listen_port

    if $ipv6_enable == true {
      $ipv6_port = $ipv6_listen_port
    }
  } else {
    $port = pick_default($listen_port_range, $listen_port)

    if $ipv6_enable == true {
      $ipv6_port = pick_default($ipv6_listen_port_range, $ipv6_listen_port)
    }
  }
```
https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/REFERENCE.md#pick_default

00ssh: @dhoppe Hi, Thanks for your suggestions. I've just tested with pick_default and it works just fine. However, I still have to figure out why the travis pipeline is failing (I don't have an environment connected to the internet right now). It looks like "fact('nginx_version')" doesn't return anything.

dhoppe: @00ssh Maybe, for that reason I used ```$facts['nginx_version']``` at the example from above.

00ssh: @dhoppe Thanks. I've updated accordingly. Waiting for tests to pass this time!

spec/defines/resource_server_spec.rb
```diff
@@ -125,6 +125,18 @@
               value: 45,
               match: %r{\s+listen\s+\[::\]:45 default ipv6only=on;}
             },
+            {
+              title: 'should set the IPv6 listen port',
```
dhoppe: You might want to change the title to 'should set the IPv6 listen port range'.

00ssh: Updated. Thanks for pointing it out!

dhoppe:
@ekohl I would like to merge this pull request. Any objections?

manifests/resource/server.pp
```diff
@@ -316,9 +320,24 @@
     }
   }
 
+  # If port range is defined, ignore any other $listen_port defined
+  if versioncmp($facts['nginx_version'], '1.15.10') < 0 {
+    $port = $listen_port
+
+    if $ipv6_enable == true {
```
bastelfreak: You don't need to update the code, but the following would work as well and is shorter:
```suggestion
    if $ipv6_enable {
```

00ssh:
Hi guys! Are there any other pending issues with this pull request? We would like to start using it from Forge as soon as possible.

Thanks!

00ssh:
@ekohl any updates on my pull request, please?

spec/defines/resource_server_spec.rb
```diff
@@ -446,13 +458,6 @@
               )
             end
 
-            context 'without a value for the nginx_version fact do' do
```
ekohl: Why do you delete this test?

PR #1326, opened 2019-04-29
Add rewrite_non_www_to_www option

simmerz:
This adds the opposing configuration to `rewrite_www_to_non_www`. Many site owners prefer (and this is not to get into the discussion of who is right or wrong) site urls to begin `www.` but also want a rewrite from non-`www.` for SEO purposes.

This PR adds `rewrite_non_www_to_www`, defaulting to false as per the existing param.

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -279,6 +279,10 @@ define nginx::resource::server (
     fail('You must include the nginx base class before using any defined resources')
   }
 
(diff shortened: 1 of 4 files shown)


simmerz:
That's a great idea. I'll update the PR

simmerz:
I'll update the PR. Great idea.

simmerz:
> Perhaps we should fail to compile if both `rewrite_` options are true?

That's pushed @ekohl - thanks for the feedback!

PR #1327, opened 2019-05-02
Enable SSL when requested

xaque208:
Without this change, a true value on nignx::resource::server::ssl does not
result in the 'ssl on' directive necessary to enable SSL in some cases, like
IPv6.  Here we include a fix to adhere to the user request to enable SSL for the
given nginx server.

### DIFF
--- a/templates/server/server_ssl_settings.erb
+++ b/templates/server/server_ssl_settings.erb
@@ -1,4 +1,4 @@
-<% if @add_listen_directive -%>
+<% if @ssl -%>
   ssl on;
 <% end -%>
 <% if @ssl_cert -%>


xaque208:
I think maybe this is just me not knowing the state of nginx.

xaque208:
This PR includes the `ssl on` in my config, but after upgrading 1.14.2_13,2 -> 1.16.0,2 I now see this warning in my `nginx -t`.

```
nginx: [warn] the "ssl" directive is deprecated, use the "listen ... ssl" directive instead in /usr/local/etc/nginx/sites-enabled/zleslie.info.conf:20
```

I guess its still a step in the right direction, but I need to read the nginx docs for a different PR to fix the listen lines, but probably someone will get to it before me.

bastelfreak:
Hi @xaque208, thanks for the PR. Can you please rebase this PR and also check https://github.com/voxpupuli/puppet-nginx/pull/1330 ? That might address your issue here as well.

dhoppe:
I will close this issue, because it has been solved by #1330.

PR #1330, opened 2019-05-08
Replace `add_listen_directive` with `nginx_version`

alexjfisher:
Nginx 1.16 has recently been released.  Instead of updating the README
to recommend users set the `add_listen_directive` parameter, this
parameter has been replaced by an `nginx_version` parameter that
defaults to the fact (if available).  If the fact isn't available, it
defaults to a very conservative `1.6.0` (the version that ships in Debian
8).  Future enhancements could make the non-fact default be OS specific.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -24,6 +24,14 @@
 # node default {
 #   include nginx
 # }
+#
+# @param nginx_version
+#   The version of nginx installed (or being installed).
+#   Unfortunately, different versions of nginx may need configuring
+#   differently.  The default is derived from the version of nginx
+#   already installed.  If the fact is unavailable, it defaults to '1.6.0'.
+#   You may need to set this manually to get a working and idempotent
+#   configuration.
 class nginx (
   ### START Nginx Configuration ###
   Variant[Stdlib::Absolutepath, Boolean] $client_body_temp_path = $nginx::params::client_body_temp_path,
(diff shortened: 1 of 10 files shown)


manifests/init.pp
```diff
@@ -24,6 +24,14 @@
 # node default {
 #   include nginx
 # }
+#
+# @param nginx_version
+#   The version of nginx installed (or being installed).
```
alexjfisher: language shameless borrowed from one of theforeman/puppet parameters ;)

manifests/init.pp
```diff
@@ -24,6 +24,14 @@
...
+#   Unfortunately, different versions of nginx may need configuring
+#   differently.  The default is derived from the version of nginx
+#   already installed.  If the fact is unavailable, it defaults to '1.6.0'.
```
bastelfreak: `1.6.0`? Did you mean `1.16.0`?

manifests/init.pp
```diff
@@ -185,7 +193,7 @@
   Hash $nginx_upstreams                                   = {},
   Nginx::UpstreamDefaults $nginx_upstreams_defaults       = {},
   Boolean $purge_passenger_repo                           = true,
-  Boolean $add_listen_directive                           = $nginx::params::add_listen_directive,
+  String[1] $nginx_version                                = pick(fact('nginx_version'), '1.6.0'),
```
bastelfreak: `1.6.0`? Did you mean `1.16.0`?

manifests/init.pp
alexjfisher: No, that wasn't a typo.  See commit message or https://github.com/voxpupuli/puppet-nginx/pull/1330#issue-276960603  I wanted to pick a number < 1.15.0 and beyond that it didn't really matter what.  Debian 8 is a supported OS that ships with 1.6.x, so that seemed as good as any other to go with.

Maybe in a separate PR you might want to implement some of
> Future enhancements could make the non-fact default be OS specific.

Issue #1086, opened 2017-05-10
Bump puppetlabs/apt dependency


den-is:
```
Warning: Module 'puppetlabs-apt' (v4.0.0) fails to meet some dependencies:
  'puppet-nginx' (v0.6.0) requires 'puppetlabs-apt' (>= 2.1.0 <3.0.0)
```
According to the Puppet Style Guide https://docs.puppet.com/puppet/latest/style_guide.html#dependencies , epel and apt are soft dependencies and should be treated as such.

cfoehrdes:
We are also experiencing a lot of conflicts due to puppet-nginx referencing hard to this old apt module

wyardley:
How about #1106 as a fix for this?

cristianbalan:
Any development here please?

cristianbalan:
Sorry, what's the conclusion to this? I face the same problem.
> Warning: Module 'puppetlabs-apt' (v3.0.0) fails to meet some dependencies:
>   'puppet-nginx' (v0.6.0) requires 'puppetlabs-apt' (>= 2.1.0 <3.0.0)
Upgrading the module to 0.7.0 would still be requiring the old module but would not throw the warning?

wyardley:
@cristianbalan take a look at the changes in:
https://github.com/voxpupuli/puppet-nginx/pull/1106/files
now that it's a soft dependency, the module is not explicitly required; you're just responsible for making sure that an appropriate version of the apt module is available.

cristianbalan:
Looks perfect @wyardley! Any plan to push that to Forge at some point in the next days?
Thanks.

PR #1332, opened 2019-05-28
Update metadata.json to add apt dependency

davidalexander83:
apt module is not included within puppet agent by default, and causes catalog compile failure on Ubuntu/Debian.


alexjfisher:
@davidalexander83 Thanks for the PR.  Not having `puppetlabs/apt` in the metadata is actually a conscious decision though,
See https://github.com/voxpupuli/puppet-nginx/issues/1086

As such, I'm going to close this.  Please don't be put off creating more PRs in the future though! :)

PR #1333, opened 2019-05-31
fixing some documentation for setting up UDP streams

martinrw:
#### Pull Request (PR) description
The hiera example in the readme file about configuring a udp stream had a few errors. 
- The ports need to have no apostrophes as they are numbers
- Some colons were missing
- The appropriate "include" sections don't get added to nginx.conf unless "nginx::stream" is set

#### This Pull Request (PR) fixes the following issues
n/a

PR #1334, opened 2019-06-03
Allow puppetlabs/concat 6.x

dhoppe:
This wil fix #1333.


bastelfreak:
marked as bug because of https://github.com/voxpupuli/puppet-openvpn/pull/340

PR #1336, opened 2019-06-14
Add nginx config for stream log

ssodhi:


ardrigh:
This should be closed and refer to updated PR #1401

PR #1341, opened 2019-08-08
Install libnginx-mod-http-passenger on newer systems

### LINKED ISSUE #1340: libnginx-mod-http-passenger should be installed instead of passenger

- Puppet: Any
- Ruby: Any
- Distribution: Debian 9 and Ubuntu 18.04+
- Module version: Any

add the following to hiera on a debian 9 or Ubuntu 18.04 and apply then nginx module
```
nginx::package_source: passenger
nginx::package_name: nginx-extras
nginx::http_cfg_append:
  'passenger_root': '/usr/lib/ruby/vendor_ruby/phusion_passenger/locations.ini'
  'passenger_ruby': '/usr/bin/passenger_free_ruby
```
## What are you seeing

Nginx throws an error regarding passenger_root
```
nginx: [emerg] unknown directive "passenger_root" in /etc/nginx/nginx.conf:64
```

A valid nginx configuration thats able to understand passenger.

The passenger install guide says to use libnginx-mod-http-passenger and not passenger for Debian 9 and Ubuntu 18.04. libnginx-mod-http-passenger will pull passenger as a dependency.

https://www.phusionpassenger.com/docs/tutorials/deploy_to_production/installations/oss/ownserver/ruby/nginx/

Jonasee:
On Debian 9 & Ubuntu 18.04 libnginx-mod-http-passenger needs to be
installed instead of passenger according to the installation guide by
passenger. libnginx-mod-http-passenger has a dependency on passenger, so
it will be installed. Without libnginx-mod-http-passenger nginx will not
be able to handle passenger.

Installation guide:
https://www.phusionpassenger.com/docs/tutorials/deploy_to_production/installations/oss/ownserver/ruby/nginx/

Fixes #1340

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -165,7 +165,7 @@ class nginx (
   Boolean $mime_types_preserve_defaults                      = false,
   Optional[String] $repo_release                             = undef,
   $passenger_package_ensure                                  = 'present',
-  $passenger_package_name                                    = $nginx::params::passenger_package_name,
+  String $passenger_package_name                             = $nginx::params::passenger_package_name,
   Optional[Stdlib::HTTPUrl] $repo_source                     = undef,
   ### END Package Configuration ###
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -63,6 +63,7 @@ class nginx (
   Boolean $super_user                                        = $nginx::params::super_user,
   $temp_dir                                                  = $nginx::params::temp_dir,
   Boolean $server_purge                                      = false,
+  Boolean $include_modules_enabled                           = $nginx::params::include_modules_enabled,
 
   # Primary Templates
   $conf_template                                             = 'nginx/conf.d/nginx.conf.erb',
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -165,6 +165,7 @@ class nginx (
   Boolean $mime_types_preserve_defaults                      = false,
   Optional[String] $repo_release                             = undef,
   $passenger_package_ensure                                  = 'present',
+  $passenger_package_name                                    = $nginx::params::passenger_package_name,
   Optional[Stdlib::HTTPUrl] $repo_source                     = undef,
   ### END Package Configuration ###
 
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -71,9 +71,19 @@ class nginx::package::debian {
           key      => {'id' => '16378A33A6EF16762922526E561F9B9CAC40B2F7'},
         }
 
-        package { 'passenger':
-          ensure  => $passenger_package_ensure,
-          require => Exec['apt_update'],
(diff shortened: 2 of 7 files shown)


manifests/package/debian.pp
```diff
@@ -71,9 +71,19 @@
...
         }
 
-        package { 'passenger':
-          ensure  => $passenger_package_ensure,
-          require => Exec['apt_update'],
+        # From Ubuntu 18.04/Debian 9 libnginx-mod-http-passenger is needed
+        # libnginx-mod-http-passenger will install passenger as a dependency
+        if ( ($::operatingsystem == 'Ubuntu') and ($facts['os']['release']['major'] >='18')) or
+           ( ($::operatingsystem == 'Debian') and ($facts['os']['release']['major'] >='9')) {
+              package { 'libnginx-mod-http-passenger':
```
ekohl: Since this block is very similar to the previous block (besides the package name), could you use a variable?

Jonasee:
So something like what I just added with the passenger package name in params.pp?

manifests/package/debian.pp
Jonasee: So something like what I just added with the passenger package name in params.pp?

manifests/package/redhat.pp
```diff
@@ -45,7 +46,7 @@
         }
 
         if $purge_passenger_repo {
-          yumrepo { 'passenger':
+          yumrepo { $passenger_package_name:
```
ekohl: This is a yumrepo, not a package

manifests/package/redhat.pp
```diff
@@ -63,7 +64,7 @@
...
```
ekohl: Same here

manifests/package/redhat.pp
Jonasee: fixed

manifests/package/redhat.pp
Jonasee: fixed

manifests/init.pp
```diff
@@ -165,6 +165,7 @@
   Boolean $mime_types_preserve_defaults                      = false,
   Optional[String] $repo_release                             = undef,
   $passenger_package_ensure                                  = 'present',
+  $passenger_package_name                                    = $nginx::params::passenger_package_name,
```
ekohl:
```suggestion
  String $passenger_package_name                              = $nginx::params::passenger_package_name,
```

Jonasee:
I looked in to that error and it seems that nginx-extras package on Ubuntu 18.04 and Debain 9 adds the /etc/nginx/modules-enabled/ and /etc/nginx/modules-available folders to load dynamic modules. The configuration files are include by default in the nginx configuration but not in the one generated by puppet.

It could be solved by adding loading the passenger module using the dynamic_modules parameter for those versions in the tests or by including /etc/nginx/modules-enabled/*.conf by default. But neither options seems like the correct solution.

PR #1346, opened 2019-09-18
Add ssl_password_file directive to support encrypted ssl keys

joernott:
This pull request adds support for the ssl_password_file directive of nginx. This directive is needed, if the ssl key file is password encrypted (should be standard nowadays).

n/a (I didn't create an issue first but tried to not only raise a request but also deliver the solution)

### DIFF
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -188,7 +188,7 @@ define nginx::resource::server (
   Optional[String] $ssl_session_ticket_key                                       = undef,
   Optional[String] $ssl_trusted_cert                                             = undef,
   Optional[Integer] $ssl_verify_depth                                            = undef,
-  Optional[String] $ssl_password_file                                            = undef,
+  Optional[Stdlib::Absolutepath] $ssl_password_file                              = undef,
   Enum['on', 'off'] $spdy                                                        = $nginx::spdy,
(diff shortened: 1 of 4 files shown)


joernott:
Note: I designed this change to be non-breaking

manifests/resource/server.pp
```diff
@@ -187,6 +188,7 @@
   Optional[String] $ssl_session_ticket_key                                       = undef,
   Optional[String] $ssl_trusted_cert                                             = undef,
   Optional[Integer] $ssl_verify_depth                                            = undef,
+  Optional[String] $ssl_password_file                                            = undef,
```
bastelfreak: Can you please update the type?
```suggestion
  Optional[Stdlib::Absolutepath] $ssl_password_file                              = undef,
```

bastelfreak:
Hi @joernott, thanks for the PR. Can you please add tests to it?

joernott:
I did not find where you get your certificates from. Or is a simple test like "I added the directive and it shows up in the config file" sufficient?

bastelfreak:
It's okay to add an rspec test that verifies the content of the file.

bastelfreak:
Thanks for the awesome acceptance test!

PR #1354, opened 2019-10-31
Add VirtuozzoLinux support

bastelfreak:

### DIFF
--- a/manifests/package/redhat.pp
+++ b/manifests/package/redhat.pp
@@ -27,8 +27,9 @@ class nginx::package::redhat {
   #clone and provide the Red Hat-specific package. This comes into play when not
   #on RHEL or CentOS and $manage_repo is set manually to 'true'.
   $_os = $facts['os']['name'] ? {
-    'centos' => 'centos',
-    default  => 'rhel'
+    'centos'         => 'centos',
+    'VirtuozzoLinux' => 'centos',
+    default          => 'rhel'
   }
 
   if $manage_repo {
@@ -70,7 +71,7 @@ class nginx::package::redhat {
         }
       }
       'passenger': {
-        if ($facts['os']['name'] in ['RedHat', 'CentOS']) and ($facts['os']['release']['major'] in ['6', '7']) {
+        if ($facts['os']['name'] in ['RedHat', 'CentOS', 'VirtuozzoLinux']) and ($facts['os']['release']['major'] in ['6', '7']) {
           yumrepo { 'passenger':
             baseurl       => "https://oss-binaries.phusionpassenger.com/yum/passenger/el/${facts['os']['release']['major']}/\$basearch",
             descr         => 'passenger repo',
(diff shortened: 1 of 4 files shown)


ghoneycutt:
Does `facter operatingsystem` return `VirtuozzoLinux` ? I thought it was just a RH derivative.

bastelfreak:
Since facter 3.14 it does.

manifests/package/redhat.pp
```diff
@@ -70,7 +71,7 @@
         }
       }
       'passenger': {
-        if ($facts['os']['name'] in ['RedHat', 'CentOS']) and ($facts['os']['release']['major'] in ['6', '7']) {
+        if ($facts['os']['name'] in ['RedHat', 'CentOS', 'VirtuozzoLinux']) and ($facts['os']['release']['major'] in ['6', '7']) {
```
runejuhl: This looks similar to https://github.com/voxpupuli/puppet-nginx/pull/1354/files#diff-2f383a197b1a428561b364755701d60eR146 ; should this line not also have `Oracle` in there?

If that's the case, maybe it'd be an idea to put that snippet into a function, so you only every have to change it in one place.

spec/classes/nginx_spec.rb
```diff
@@ -83,7 +84,7 @@
 
             it do
               is_expected.to contain_yumrepo('nginx-release').with(
-                'baseurl' => "https://nginx.org/packages/mainline/#{facts[:operatingsystem] == 'CentOS' ? 'centos' : 'rhel'}/#{facts[:operatingsystemmajrelease]}/$basearch/"
+                'baseurl' => "https://nginx.org/packages/mainline/#{%w[CentOS VirtuozzoLinux].include?(facts[:operatingsystem]) ? 'centos' : 'rhel'}/#{facts[:operatingsystemmajrelease]}/$basearch/"
```
runejuhl: Same as above -- should these perhaps have included `Oracle` as well? Might it be better to simply reverse the test, is only RedHat proper uses `rhel` and the rest should use `centos`?

```ruby
'baseurl' => "https://nginx.org/packages/mainline/#{facts[:operatingsystem] == 'RedHat' ? 'rhel' : 'centos'}/#{facts[:operatingsystemmajrelease]}/$basearch/"
```

spec/classes/nginx_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 describe 'nginx' do
   on_supported_os.each do |os, facts|
+    next unless facts[:operatingsystem] == 'VirtuozzoLinux'
```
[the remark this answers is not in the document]
bastelfreak: crap. fixed it

manifests/package/redhat.pp
bastelfreak: Sadly I don't have an oracle box to test. I don't know which repo is corret. Since it's an RHEL clone `rhel` might actually be correct? IMO we should keep it as is until somebody with an Oracle box tests it / reports back.

runejuhl: SGTM!

spec/classes/nginx_spec.rb
runejuhl: See https://github.com/voxpupuli/puppet-nginx/pull/1354/files#r347156132

PR #1355, opened 2019-10-31
Mock facter version based on puppet version & unit tests: print puppet/facter version

bastelfreak:
Getting the correct facter version is tricky. We use facterdb as a source to mock facts see https://github.com/camptocamp/facterdb. People might provide a specific facter version.  In that case we use it. Otherwise we need to match the correct facter version to the used puppet version. As of 2019-10-31, puppet 5 ships facter 3.11 and puppet 6 ships facter 3.14 (see https://puppet.com/docs/puppet/5.5/about_agent.html).

### DIFF
--- a/spec/defines/resource_geo_spec.rb
+++ b/spec/defines/resource_geo_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe 'nginx::resource::geo' do
   on_supported_os.each do |os, facts|
-    context "on #{os}" do
+    context "on #{os} with Facter #{facts[:facterversion]} and Puppet #{facts[:puppetversion]}" do
       let(:facts) do
         facts
       end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -42,6 +42,18 @@ if Dir.exist?(File.expand_path('../../lib', __FILE__))
 end
 
 RSpec.configure do |c|
+  # getting the correct facter version is tricky. We use facterdb as a source to mock facts
+  # see https://github.com/camptocamp/facterdb
+  # people might provide a specific facter version. In that case we use it.
+  # Otherwise we need to match the correct facter version to the used puppet version.
+  # as of 2019-10-31, puppet 5 ships facter 3.11 and puppet 6 ships facter 3.14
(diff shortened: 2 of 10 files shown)


spec/spec_helper.rb
```diff
@@ -42,6 +42,18 @@
 end
 
 RSpec.configure do |c|
+  # getting the correct facter version is tricky. We use facterdb as a source to mock facts
```
ekohl: Are you going to submit this to modulesync as well?

bastelfreak: Of course. I just need to test it here.

spec/defines/resource_geo_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe 'nginx::resource::geo' do
   on_supported_os.each do |os, facts|
-    context "on #{os}" do
+    context "on #{os} with Facter #{facts[:facterversion]} and Puppet #{facts[:puppetversion]} with Facter #{facts[:facterversion]} and Puppet #{facts[:puppetversion]}" do
```
DavidS: double paste

bastelfreak: thanks, fixed it.

spec/spec_helper.rb
```diff
@@ -42,6 +42,18 @@
...
+  # see https://github.com/camptocamp/facterdb
+  # people might provide a specific facter version. In that case we use it.
+  # Otherwise we need to match the correct facter version to the used puppet version.
+  # as of 2019-10-31, puppet 5 ships facter 3.11 and puppet 6 ships facter 3.14
+  # https://puppet.com/docs/puppet/5.5/about_agent.html
+  c.default_facter_version = if ENV['FACTERDB_FACTS_VERSION']
+                               ENV['FACTERDB_FACTS_VERSION']
+                             else
+                               Gem::Dependency.new('', ENV['PUPPET_VERSION']).match?('', '5') ? '3.11.0' : '3.14.0'
```
alexjfisher: In theory, `3.11` and `3.14` (without the `.0`) would work now.  But that's not important.

PR #1363, opened 2019-12-09
Enforce Usage Of Managed Repository

fzoske:
This change enforce the usage of the managed repository to work around the AppStream of CentOS and RedHat, which will take precedence over none module based repositories.

### DIFF
--- a/manifests/package/redhat.pp
+++ b/manifests/package/redhat.pp
@@ -31,6 +31,11 @@ class nginx::package::redhat {
     'VirtuozzoLinux' => 'centos',
     default          => 'rhel'
   }
+  
+  $_package_install_options ? {
+     $manage_repo => '--repo=nginx-release',
+     default      => '',
+  }
 
   if $manage_repo {
     case $package_source {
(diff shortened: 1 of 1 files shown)


manifests/package/redhat.pp
```diff
@@ -31,6 +31,11 @@
     'VirtuozzoLinux' => 'centos',
     default          => 'rhel'
   }
+  
+  $_package_install_options ? {
+     $manage_repo => '--repo=nginx-release',
+     default      => '',
```
ekohl: This probably needs to be `undef`

manifests/package/redhat.pp
```diff
@@ -31,6 +31,11 @@
     'VirtuozzoLinux' => 'centos',
     default          => 'rhel'
   }
+  
+  $_package_install_options ? {
```
ekohl: There's something missing here.

bastelfreak:
This basically adds RedHat 8 support, right? If so, can you please add CentOS 8 / RedHat 8 to the metadata.json? This depends on https://github.com/camptocamp/facterdb/pull/137

PR #1371, opened 2020-01-25
release 1.1.0

bastelfreak:


bastelfreak:
@ghoneycutt we've a convention to only merge a release PR if the same person runs the `travis_release` rake task. Otherwise the release process often gets lost. Usually reviewers only approve a release PR and the person that created it merges it and runs the rake task. Please follow that pattern as well.

PR #1375, opened 2020-02-20
Add apt to dependencies in metadata.json

### LINKED ISSUE #1374: puppetlabs/apt missing from metadata.json dependencies
https://github.com/voxpupuli/puppet-nginx/blob/be0a031a0090181a9c9c2c9b5eb3eb817449e239/manifests/package/debian.pp#L35 pulls in puppetlabs/apt but it is not listed in https://github.com/voxpupuli/puppet-nginx/blob/be0a031a0090181a9c9c2c9b5eb3eb817449e239/metadata.json#L10-L19

genebean:

Prior to this commit, apt was required in the fixutures but not in metadata.json. Version 7.2.0 was chosen as the lower bounds of the apt module version because that is when Debian 10 support was added.

Fixes #1374

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -9,6 +9,10 @@
   "issues_url": "https://github.com/voxpupuli/puppet-nginx/issues",
   "dependencies": [
     {
+      "name": "puppetlabs/apt",
+      "version_requirement": ">= 7.2.0 < 8.0.0"
+    },
+    {
       "name": "puppetlabs/concat",
       "version_requirement": ">= 4.1.0 < 7.0.0"
     },


ekohl:
https://github.com/voxpupuli/puppet-nginx#requirements states this as a soft dependency. Not all platforms use apt so this has been a conscious decision.

genebean:
Personally, I’ve never seen another module that didn’t list this module when it was needed. I think omitting it is a mistake unless using nginx on Debian doesn’t require it by default (i.e. it’s included being excluded by the default value of a conditional expression)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ekohl:
https://puppet.com/docs/puppet/latest/style_guide.html#dependencies states:
> Hard dependencies must be declared explicitly in your module’s metadata.json file.
> Soft dependencies should be called out in the README.md, and must not be enforced as a hard requirement in your metadata.json. A soft dependency is a dependency that is only required in a specific set of use cases. For an example, see the rabbitmq module.

genebean:
Learned something new today, thanks @ekohl

PR #1376, opened 2020-02-20
Manage libnginx-mod-http-passenger on modern Debian / Ubuntu

### LINKED ISSUE #1340: libnginx-mod-http-passenger should be installed instead of passenger

smortex:
This is a follow-up to #1341 (closed without being merged) which was an attempt to fix #1340. I took the commits in this legacy PR, rebased them on top of master (fixing a bunch of conflicts) and am now fixing the CI.

Fixes #1340

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -166,6 +166,7 @@ class nginx (
   Boolean $mime_types_preserve_defaults                      = false,
   Optional[String] $repo_release                             = undef,
   $passenger_package_ensure                                  = 'present',
+  $passenger_package_name                                    = $nginx::params::passenger_package_name,
   Optional[Stdlib::HTTPUrl] $repo_source                     = undef,
   ### END Package Configuration ###
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -166,7 +166,7 @@ class nginx (
   Boolean $mime_types_preserve_defaults                      = false,
   Optional[String] $repo_release                             = undef,
   $passenger_package_ensure                                  = 'present',
-  $passenger_package_name                                    = $nginx::params::passenger_package_name,
+  String $passenger_package_name                             = $nginx::params::passenger_package_name,
   Optional[Stdlib::HTTPUrl] $repo_source                     = undef,
   ### END Package Configuration ###
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -25,6 +25,10 @@
 #   include nginx
 # }
 #
+# @param include_modules_enabled
+#   When set, nginx will include module configurations files installed in the
+#   /etc/nginx/modules-enabled directory.
+#
 # @param nginx_version
 #   The version of nginx installed (or being installed).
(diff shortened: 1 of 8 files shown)


manifests/init.pp
```diff
@@ -64,6 +64,7 @@
   Boolean $super_user                                        = $nginx::params::super_user,
   $temp_dir                                                  = $nginx::params::temp_dir,
   Boolean $server_purge                                      = false,
+  Boolean $include_modules_enabled                           = $nginx::params::include_modules_enabled,
```
bastelfreak: can you please add documentation for the new parameters?

manifests/init.pp
```diff
@@ -166,6 +171,7 @@
   Boolean $mime_types_preserve_defaults                      = false,
   Optional[String] $repo_release                             = undef,
   $passenger_package_ensure                                  = 'present',
+  String $passenger_package_name                             = $nginx::params::passenger_package_name,
```
bastelfreak: Same here. Please document the parameter and enforce a minimal string length of 1:
```suggestion
  String[1] $passenger_package_name                          = $nginx::params::passenger_package_name,
```

smortex: Sorry for the delay, I just pushed a fix!

bastelfreak:
I'm not sure if this is a backwards-incompatible release or not. What about those newer Debian versions if they currently try to use the module? Doesn't it work at all? What happens if the module installed the passanger package in the past and now install the new package? Does that work at all?

smortex:
Back from holidays, sorry for the delay!

The `libnginx-mod-http-passenger` package depends on the `passenger` package which was previously including everything.  The new package installs the actual nginx shlib and the adequate configuration file.

Per se, the previous behavior is working from the Puppet point of view, but the resulting configuration is not working (because the nginx module is not installed nor configured).

My guess is that this would only require a MINOR version bump: the change is backward compatible but adds support for systems previously unsupported.

Thanks!

PR #1379, opened 2020-04-04
modulesync 2.12.0

bastelfreak:
modulesync 2.12.0

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '2.10.1'
+modulesync_config_version: '2.12.0'
--- a/Gemfile
+++ b/Gemfile
@@ -11,27 +11,9 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'puppetlabs_spec_helper', '>= 2.14.0',                        :require => false
-  gem 'rspec-puppet-facts', '>= 1.9.5',                             :require => false
-  gem 'rspec-puppet-utils',                                         :require => false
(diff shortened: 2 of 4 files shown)


ekohl:
Restarted the timed out build.

bastelfreak:
I don't quite understand why the Puppet 6 tests are exloding and take over 50 minutes :(

.github/workflows/ruby.yaml
```diff
@@ -0,0 +1,47 @@
...
+        bundle install --jobs 4 --retry 3
+    - name: execute puppet-lint
+      run: bundle exec rake lint
+    - name: Check syntax of Ruby files and call :syntax and :metadata_lint
+      run: bundle exec rake validate
+    - name: Fails if symlinks are present in directory
+      run: bundle exec rake check:symlinks
+    - name: Fails if .pp files present in tests folder
+      run: bundle exec rake check:test_file
+    - name: Check syntax of Ruby files and call :syntax and :metadata_lint
```
ekohl, line 40: These names appear inconsistent with their actual test

bastelfreak:
I wanted to get proper timings for each rake task. Because of that I explicitly listed every check seperately, instead of using the release_checks. In the future we can probably rely on that again.

.github/workflows/ruby.yaml
```diff
@@ -0,0 +1,47 @@
+name: Ruby
```
ekohl, line 1: I'd be more descriptive here

ekohl:
So I removed the Github actions Puppet 6 change, but now that fails on Travis again because it takes too long.

PR #1382, opened 2020-04-22
Add ability to set ssl-settings globally - fixes #670

### LINKED ISSUE #670: SSL Parameter should be configurable outside of vhosts
It may be common practice, but there is no point in configuring the ssl parameters in every vhosts.
There should be a global configuration that sets secure defaults and only if a vhosts really needs a custom config, should the vhost parameters be used.
This would make all vhosts more readable, mass updates more stable and analysis of the configuration by admins much easier.

This seems even more important since the default value is a very verbose way of naming the currently mostly used ciphers.

What does @jfryman think about this? If you can confirm that this would be helpful, i could work on a PR to implement this. But if this is not desired I will stick with a fork.

Another alternative would to be to handle an undef or empty value of ciphers and handling that case by not adding a configuration line.

TuningYourCode:
This PR enables to set ssl_* settings globally in nginx.conf to prevent setting them on all vhosts/servers manually.
This is also useful if you do not maintain all nginx vhosts/servers by puppet but still be able to have the same ssl_* settings (like ssl_ciphers, ssl_protocols etc.) on all of them.

Fixes #670


manifests/resource/server.pp
```diff
@@ -171,32 +171,32 @@
   Optional[Variant[String, Boolean]] $ssl_cert                                   = undef,
   Optional[String] $ssl_client_cert                                              = undef,
   String $ssl_verify_client                                                      = 'on',
-  Optional[String] $ssl_dhparam                                                  = $nginx::ssl_dhparam,
+  Optional[String] $ssl_dhparam                                                  = undef,
```
TuningYourCode, line 174: Is this change ok, as if `$nginx::ssl_dhparam` is defined it will now get written to `nginx.conf` or is this considered a backward-incompatible change even if nginx should be working as before this change?

dhoppe: I do not think this change is backwards-incompatible, because the value was ```undef```before and can be overwritten any time. However, you are using a different data type in ```init.pp``` and ```server.pp```. 

Unfortunately I can not have a look at the failing Travis CI tests, because the website is under maintenance.

ekohl: I don't mind a breaking change for this since it makes sense IMHO.

ekohl:
A test worked timed out, restarted the build now.

TuningYourCode:
Sadly it timed out again. Did i introduce to many new test cases? But wouldn't explain why puppet 5.x ran only like 30 mins and on puppet 6.x it takes more than 60 minutes.

manifests/resource/server.pp
TuningYourCode: @dhoppe i made the settings in init.pp more strict as these are new parameters. I didn't want to break anything with my changes if somebody is already using `nginx::resource::server`.

There are also `$ssl_stapling` and `$ssl_stapling_verify` which are `Enum['on', 'off']` on init.pp and `Boolean` in server.pp
Unifying them will be definitely a breaking-change and should probably be done in a separate ticket?

If there is any new data type to strict just tell me and i will change it :) I would like to see this change as soon as possible released as we would like to move some ssl-settings vom vhost/servers to the global config in our infrastructure.

bastelfreak:
@TuningYourCode I rebased it and it should be fine now

PR #1384, opened 2020-05-04
Drop Debian 8 support

bastelfreak:
It's currently in extended support time and even that ends in about a
month.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
   gem 'rspec-puppet-facts',          :require => false, :git => 'https://github.com/mcanevet/rspec-puppet-facts', :ref => '9541292d4fc35db3be1badace673c1108154b571'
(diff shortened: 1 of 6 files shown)


manifests/params.pp
```diff
@@ -121,8 +121,7 @@
           'passenger_package_name'  => 'libnginx-mod-http-passenger',
           'include_modules_enabled' => true,
         }
-      } elsif ($facts['os']['name'] == 'ubuntu' and $facts['os']['distro']['codename'] in ['lucid', 'precise', 'trusty', 'xenial'])
-      or ($facts['os']['name'] == 'debian' and $facts['os']['release']['major'] in ['6', '7', '8']) {
+      } elsif ($facts['os']['name'] == 'ubuntu' and $facts['os']['distro']['codename'] in ['lucid', 'precise', 'trusty', 'xenial']) {
```
ekohl, line 124: Isn't lucid AFAIK only Xenial is supported here. However, I'd argue we should just drop this and never manage the repo by default. Just trust the OS versions and leave it up to users.

bastelfreak: I had the same in mind, but I guess it should be handled in a different PR

Gemfile
```diff
@@ -11,7 +11,7 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
```
ekohl, line 14: This is modulesynced.

manifests/params.pp
ekohl: Perhaps that should be done first to keep this PR smaller (and the other really focused on just dropping that).

Gemfile
bastelfreak: I know, I wanted a hotfix to get master green again. I will have a look at modulesync_config later :)

PR #1385, opened 2020-05-05
Drop Ubuntu 10/12/14 code

bastelfreak:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -109,8 +109,17 @@ class nginx::params {
       }
     }
     'Debian': {
-      if ($facts['os']['name'] == 'ubuntu' and $facts['os']['distro']['codename'] in ['bionic'])
-      or ($facts['os']['name'] == 'debian' and $facts['os']['release']['major'] in ['9', '10']) {
+      if ($facts['os']['name'] == 'ubuntu' and $facts['os']['distro']['codename'] == 'xenial') {
+        $_module_os_overrides = {
+          'manage_repo' => true,
+          'daemon_user' => 'www-data',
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -130,14 +130,6 @@
           'log_mode'    => '0755',
           'run_dir'     => '/run/nginx',
         }
-      } else {
```
ekohl, line 133: Why drop this? Doesn't Ubuntu 18.04 hit this? I also think it'd be a better and safer thing to have an `else` clause. If anything, Xenial should perhaps not manage the repo at all.

bastelfreak: the syntax isn't perfect here, above this block is one explicitly for Debian 9/10 and Ubuntu 18

ekohl: Still, having a safe default feels better to me. I'd also be in favor of dropping `manage_repo => true` and moving to not managing it by default. Then we can make the else block the future proof one so Debian 11/Ubuntu 20.04 are more likely to just work.

bastelfreak: I'm not sure if we should change any default here. However I'm fine with providing a default block that catches newer ubuntu/debian version to be future-proof. My personal preference is also to always use the upstream repositories and not the packages from Ubuntu.

ekohl:
> I feel that the `manage_repo` behavior should be documented in the changelog somehow.

Can we open an issue for this and let this PR close it? Then it's documented.

bastelfreak:
closes https://github.com/voxpupuli/puppet-nginx/issues/1441

PR #1396, opened 2020-07-17
Add debug_connection events option

juokelis:
Addes a possibility to use [debug_connection](http://nginx.org/en/docs/ngx_core_module.html#debug_connection) under events. 

n/a

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,6 +40,10 @@
 #   already installed.  If the fact is unavailable, it defaults to '1.6.0'.
 #   You may need to set this manually to get a working and idempotent
 #   configuration.
+#
+# @param debug_connections
+#   Configures nginx `debug_connection` lines in the `events` section of the nginx config.
+#   See http://nginx.org/en/docs/ngx_core_module.html#debug_connection
 class nginx (
   ### START Nginx Configuration ###
   Variant[Stdlib::Absolutepath, Boolean] $client_body_temp_path = $nginx::params::client_body_temp_path,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -88,6 +88,7 @@ class nginx (
   $lingering_timeout                                         = '5s',
   Optional[Enum['on', 'off']] $etag                          = undef,
   Optional[String] $events_use                               = undef,
+  Array[Nginx::DebugConnection] $debug_connections           = [],
   String $fastcgi_cache_inactive                             = '20m',
   Optional[String] $fastcgi_cache_key                        = undef,
   String $fastcgi_cache_keys_zone                            = 'd3:100m',
--- a/spec/type_aliases/debugconnection_spec.rb
+++ b/spec/type_aliases/debugconnection_spec.rb
@@ -1,0 +1,10 @@
+require 'spec_helper'
+
+describe 'Nginx::DebugConnection' do
+  it { is_expected.to allow_value('127.0.0.1') }
+  it { is_expected.to allow_value('localhost') }
+  it { is_expected.to allow_value('192.0.2.0/24') }
+  it { is_expected.to allow_value('::1') }
+  it { is_expected.to allow_value('2001:0db8::/32') }
+  it { is_expected.to allow_value('unix:') }
+end
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -40,10 +40,13 @@ events {
   <%- end -%>
   worker_connections <%= @worker_connections -%>;
   <%- if @multi_accept == 'on' -%>
-    multi_accept on;
+  multi_accept on;
   <%- end -%>
   <%- if @events_use -%>
-    use <%= @events_use %>;
+  use <%= @events_use %>;
+  <%- end -%>
+  <%- @debug_connections.each do |address| -%>
+  debug_connection <%= address %>;
   <%- end -%>
 }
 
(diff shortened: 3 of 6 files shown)


templates/conf.d/nginx.conf.erb
```diff
@@ -40,11 +40,16 @@ events {
...
   <%- if @multi_accept == 'on' -%>
-    multi_accept on;
+  multi_accept on;
   <%- end -%>
   <%- if @events_use -%>
-    use <%= @events_use %>;
+  use <%= @events_use %>;
   <%- end -%>
+  <% if @debug_connection && Array(@debug_connection).size > 0 -%>
+    <%- Array(@debug_connection).each do |address| -%>
```
ekohl, line 49: The data type is enforced so no need to cast to `Array()`. It's also cleaner to use empty though it's entirely redundant since it'll just loop 0 times
```suggestion
  <% if @debug_connection -%>
    <%- @debug_connection.each do |address| -%>
```

manifests/init.pp
```diff
@@ -88,6 +88,7 @@
   $lingering_timeout                                         = '5s',
   Optional[Enum['on', 'off']] $etag                          = undef,
   Optional[String] $events_use                               = undef,
+  Optional[Array[String]] $debug_connection                  = undef,
```
ekohl, line 91: How are these specified? Can you use `Stdlib::Host` instead of `String? If not, `String[1]` is probably better than allowing an empty string.

juokelis: Thought to keep Array or Sting, but let's stick with array and cleaner code.
Done.

templates/conf.d/nginx.conf.erb
juokelis: Thought to keep Array or Sting, but let's stick with array and cleaner code.
Done.

templates/conf.d/nginx.conf.erb
```diff
@@ -53,8 +53,8 @@ events {
 }
 
 http {
-<% if @http_raw_prepend && Array(@http_raw_prepend).size > 0 -%>
-    <%- Array(@http_raw_prepend).each do |line| -%>
+<% if @http_raw_prepend -%>
```
ekohl, line 56: I think this is not the line you wanted to change. raw prepend can be both a string and an array so here it is needed.

juokelis: Right. Thanks. Fixed.

juokelis:
> Thanks! Test failure is unrelated. Do you mind squashing the commits or shall I squash them on merge?

Done. Thank you!

ekohl:
Travis is still running. I expect it'll still pass but let's wait for it to complete.

juokelis:
As I can see there's an issue with passenger signature, not with updated code. Or am I wrong? I can see in Travis logs:

>  Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install passenger' returned 1: Importing GPG key 0xA9F2A55D:
   Userid     : "https://packagecloud.io/phusion/passenger (https://packagecloud.io/docs#gpg_signing) <support@packagecloud.io>"
   Fingerprint: fd6f dce0 2b2a 3a80 e2c6 e719 f4e6 c567 a9f2 a55d
   From       : https://packagecloud.io/phusion/passenger/gpgkey

alexjfisher:
Closing and reopening PR to restart build.

manifests/init.pp
```diff
@@ -88,6 +88,7 @@
   $lingering_timeout                                         = '5s',
   Optional[Enum['on', 'off']] $etag                          = undef,
   Optional[String] $events_use                               = undef,
+  Optional[Array[Nginx::DebugConnection]] $debug_connection  = undef,
```
alexjfisher, line 91: Since it takes an array, maybe `$debug_connections` (plural) would be a better name.
(or maybe `$debug_connection_addresses`).

Ether way, instead of making it `Optional`, I think it'd be nicer to just default it to an empty array `[]`.

templates/conf.d/nginx.conf.erb
```diff
@@ -40,11 +40,16 @@ events {
...
   worker_connections <%= @worker_connections -%>;
   <%- if @multi_accept == 'on' -%>
-    multi_accept on;
+  multi_accept on;
   <%- end -%>
   <%- if @events_use -%>
-    use <%= @events_use %>;
+  use <%= @events_use %>;
   <%- end -%>
+  <% if @debug_connection -%>
```
alexjfisher, line 48: If we simplify the data type to only allow Arrays, (and not `undef`), the template can be simplified too.

alexjfisher (changes requested):

templates/conf.d/nginx.conf.erb
juokelis: done :)

manifests/init.pp
juokelis: right, thanks!

alexjfisher:
@juokelis Thanks!  I've added an extra commit with some tests and will merge when the build goes green.

spec/type_aliases/debugconnection_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe 'Nginx::DebugConnection' do
+  it { is_expected.to allow_value('127.0.0.1') }
+  it { is_expected.to allow_value('localhost') }
+  it { is_expected.to allow_value('192.0.2.0/24') }
+  it { is_expected.to allow_value('::1') }
+  it { is_expected.to allow_value('2001:0db8::/32') }
```
alexjfisher, line 8: It was this line that required adding `Stdlib::IP::Address` to the type alias.

alexjfisher (approved):

PR #1397, opened 2020-07-17
Minor refactor of mailhost.pp

alexjfisher:
* Remove what looks like copy-n-pasted lines (copied from resource/server.pp ?)
* Use `Stdlib::Port` instead of `Integer`

### DIFF
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -1,7 +1,5 @@
 # define: nginx::resource::mailhost
 #
-# This definition creates a virtual host
-#
 # Parameters:
 #   [*ensure*]                    - Enables or disables the specified mailhost (present|absent)
 #   [*listen_ip*]                 - Default IP Address for NGINX to listen with this server on. Defaults to all interfaces (*)
@@ -77,13 +75,13 @@
 #  }
 #
 define nginx::resource::mailhost (
-  Integer $listen_port,
+  Stdlib::Port $listen_port,
   Enum['absent', 'present'] $ensure              = 'present',
   Variant[Array[String], String] $listen_ip      = '*',
   Optional[String] $listen_options               = undef,
   Boolean $ipv6_enable                           = false,
   Variant[Array[String], String] $ipv6_listen_ip = '::',
-  Integer $ipv6_listen_port                      = 80,
+  Stdlib::Port $ipv6_listen_port                 = 80,
   String $ipv6_listen_options                    = 'default ipv6only=on',
   Boolean $ssl                                   = false,
   Optional[String] $ssl_cert                     = undef,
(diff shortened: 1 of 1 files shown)


manifests/resource/mailhost.pp
```diff
@@ -1,7 +1,5 @@
 # define: nginx::resource::mailhost
 #
-# This definition creates a virtual host
```
alexjfisher, line 3: No it doesn't...

manifests/resource/mailhost.pp
```diff
@@ -122,22 +120,10 @@
   String $proxy_pass_error_message               = 'off',
   Array $server_name                             = [$name]
 ) {
-
```
alexjfisher, line 125: I think the new whitespace lint plugin complains about this line

manifests/resource/mailhost.pp
```diff
@@ -122,22 +120,10 @@
...
   if ! defined(Class['nginx']) {
     fail('You must include the nginx base class before using any defined resources')
   }
 
-  $root_group = $nginx::root_group
```
alexjfisher, line 130: No need to create a local variable here.

manifests/resource/mailhost.pp
```diff
@@ -122,22 +120,10 @@
...
-
-  File {
```
alexjfisher, line 132: There are no `File` resources in this manifest (and resource defaults like this are discouraged anyway)

manifests/resource/mailhost.pp
```diff
@@ -122,22 +120,10 @@
...
-    owner => 'root',
-    group => $root_group,
-    mode  => $nginx::global_mode,
-  }
-
-  $config_dir  = "${nginx::conf_dir}/conf.mail.d"
```
alexjfisher, line 138: Moved these lines to after the other validation stuff.

manifests/resource/mailhost.pp
```diff
@@ -77,13 +75,13 @@
 #  }
 #
 define nginx::resource::mailhost (
-  Integer $listen_port,
+  Stdlib::Port $listen_port,
```
ekohl, line 78: Since these are ports, you can drop the `+ 0` later on in the code. That used to be needed when it could have been a String.

manifests/resource/mailhost.pp
```diff
@@ -77,13 +75,13 @@
...
   Enum['absent', 'present'] $ensure              = 'present',
   Variant[Array[String], String] $listen_ip      = '*',
   Optional[String] $listen_options               = undef,
   Boolean $ipv6_enable                           = false,
   Variant[Array[String], String] $ipv6_listen_ip = '::',
```
ekohl, line 83: I don't know if this should also be tightened. Not sure about the nginx spec for listening.

manifests/resource/mailhost.pp
alexjfisher: Done!

manifests/resource/mailhost.pp
```diff
@@ -169,7 +158,7 @@
   }
 
   # Create SSL File Stubs if SSL is enabled
-  if ($ssl) {
+  if $ssl {
```
ekohl, line 161: If you're removing these braces, do we also remove them around line 143?

PR #1399, opened 2020-07-26
nginx isnt allowed to write into tmp directories, due to permissions …

attachmentgenie:
on centos 7 (and perhaps other os-es) the run dir parent directory isnt open enough to allow the nginx process to write into the underlying directories.


vegaaz:
I experienced the same problem.
+1

genebean (approved):

PR #1400, opened 2020-07-27
Fix permission denied while reading upstream Error

vegaaz:
To fix 'permission denied while reading upstream' Erros, the run_dir needs to be accessable by the nginx worker.
Explantation:
Typically this is a problem of saving the buffered data from the proxy server and sending it back. When the upstream server response returns large number of bytes then nginx keeps the part of data at the disk and start sending the first received bytes to browser. So for that it uses a certain directory to maintain the data at the configured directory, which is "/run/nginx" in my case. So you just need to give access to that directory to worker user of nginx.

I think the easiest way is to set the owner to the $daemon_user.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -197,6 +197,7 @@ class nginx::config {
 
   file {$run_dir:
     ensure => directory,
+    owner  => $daemon_user
   }
 
   if $nginx::manage_snippets_dir {


vegaaz:
While talking to colleague we discovered this could be a security issue.
In my case the permissions for /run/nginx are 750 ; setting it to 755 will be more secure.

PR #1398, opened 2020-07-25
modulesync 3.0.0 & puppet-lint updates

bastelfreak:
modulesync 3.0.0

PR #1407, opened 2020-08-27
Adding resolver entries to streamhost.erb

smkillen:
When declaring nginx::resource::streamhost, there is the ability to add a "resolver" option, but this is not actually honoured in the configuration, a section to test the existence of that Variable and define it in the configuration erb file is needed.

```
nginx::resource::streamhost {'test':
  ...
  resolver => ['255.255.255.255'],
  ...
}
```
should create a file in /etc/nginx/conf.stream.d/XXX-test.conf

which contains
```
server {
  ...
  resolver 255.255.255.255;
  ...
}
```
but currently does not, this code ensures this entry is present.

When declaring the parameter resolver as per
```
#   [*resolver*]            - Array: Configures name servers used to resolve
#     names of upstream servers into addresses.
```
for a streamhost, nothing is actually configured.

### DIFF
--- a/spec/defines/resource_stream_spec.rb
+++ b/spec/defines/resource_stream_spec.rb
@@ -102,6 +102,12 @@ describe 'nginx::resource::streamhost' do
               match: %r{\s+listen\s+\[::\]:80 spdy;}
             },
             {
+              title: 'should set resolver(s)',
+              attr: 'resolver',
+              value: ['203.0.113.1', '203.0.113.2'],
+              match: %r{\s+resolver\s+203.0.113.1 203.0.113.2;}
+            },
+            {
               title: 'should contain raw_prepend directives',
               attr: 'raw_prepend',
               value: [
--- a/templates/streamhost/streamhost.erb
+++ b/templates/streamhost/streamhost.erb
@@ -18,6 +18,9 @@ server {
   listen [<%= @ipv6_listen_ip %>]:<%= @ipv6_listen_port %> <% if @ipv6_listen_options %><%= @ipv6_listen_options %><% end %>;
   <%- end -%>
 <%- end -%>
+<%- unless @resolver.empty? -%>
+  resolver              <% @resolver.each do |res| %> <%= res %><% end %>;
+<%- end -%>
 
   <% Array(@raw_prepend).each do |line| -%>
     <%= line %>


igalic (changes requested):
can you please

- explain in your pull request or commit messages, what that means, and what impact it has
- add spec tests… should the code you're modifying be already tested in some way…

smkillen:
I have updated the PR description with details as to what this fixes.

igalic (approved):

igalic:
we should first merge https://github.com/voxpupuli/puppet-nginx/pull/1398

igalic:
or alternatively, this one https://github.com/voxpupuli/puppet-nginx/pull/1408

templates/streamhost/streamhost.erb
```diff
@@ -18,6 +18,9 @@ server {
   listen [<%= @ipv6_listen_ip %>]:<%= @ipv6_listen_port %> <% if @ipv6_listen_options %><%= @ipv6_listen_options %><% end %>;
   <%- end -%>
 <%- end -%>
+<%- if Array(@resolver).count > 0 -%>
```
alexjfisher, line 21: The puppet code already enforces that resolver is an array.  No need to cast it.

```suggestion
<%- unless @resolver.empty -%>
```

alexjfisher: (There's no type checking on raw_prepend/raw_append and the docs say it takes strings or arrays of strings, so the casting is needed on those lines)

alexjfisher (changes requested):
Thanks for fixing this. Could you remove the redundant casting though?

alexjfisher:
Would you be able to add a test to spec/defines/resource_stream_spec.rb?

templates/streamhost/streamhost.erb
```diff
@@ -18,7 +18,7 @@ server {
   listen [<%= @ipv6_listen_ip %>]:<%= @ipv6_listen_port %> <% if @ipv6_listen_options %><%= @ipv6_listen_options %><% end %>;
   <%- end -%>
 <%- end -%>
-<%- if Array(@resolver).count > 0 -%>
+<%- unless @resolver.empty -%>
   resolver              <% Array(@resolver).each do |res| %> <%= res %><% end %>;
```
alexjfisher, line 22: casting is redundant here too.

templates/streamhost/streamhost.erb
```diff
@@ -18,6 +18,9 @@ server {
   listen [<%= @ipv6_listen_ip %>]:<%= @ipv6_listen_port %> <% if @ipv6_listen_options %><%= @ipv6_listen_options %><% end %>;
   <%- end -%>
 <%- end -%>
+<%- unless @resolver.empty -%>
```
alexjfisher, line 21: Sorry!  There was a slight typo in what I told you.

```suggestion
<%- unless @resolver.empty? -%>
```

alexjfisher: and that's why the tests are failing (and also why tests are great to have!) :)

spec/defines/resource_stream_spec.rb
```diff
@@ -101,6 +101,12 @@
               value: 'spdy',
               match: %r{\s+listen\s+\[::\]:80 spdy;}
             },
+            {
+              title: 'should set resolver(s)',
+              attr: 'resolver',
+              value: ['8.8.8.8', '8.8.4.4'],
+              match: %r{\s+resolver\s+8.8.8.8 8.8.4.4;}
```
ekohl, line 108: I'm always a fan of using [RFC5735](https://tools.ietf.org/html/rfc5735) IPs. 
Name | Range
-- | --
TEST-NET-1 | 192.0.2.0/24
TEST-NET-2 | 198.51.100.0/24
TEST-NET-3 | 203.0.113.0/24

igalic: i can't believe no one's tried to abolish and sell these yet

alexjfisher (approved):

alexjfisher:
@smkillen Thanks!

smkillen:
No problem, thanks for the input and for accepting ;-)

PR #1408, opened 2020-08-27
Update passenger rpm gpg key

alexjfisher:
without this change, we see the following error since november 2019:

```
passenger/7/x86_64/
Retrieving key from https://packagecloud.io/phusion/passenger/gpgkey
passenger/7/x86_64/
https://oss-binaries.phusionpassenger.com/yum/passenger/el/7/x86_64/repodata/repomd.xml: [Errno -1] repomd.xml signature could not be verified for passenger
Trying other mirror.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,10 +11,11 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
+  gem 'voxpupuli-test', '< 2.0.0',   :require => false
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
-  gem 'rspec-puppet-facts',          :require => false, :git => 'https://github.com/mcanevet/rspec-puppet-facts', :ref => '9541292d4fc35db3be1badace673c1108154b571'
+  gem 'rspec-puppet-facts', '2.0.0', :require => false
+  gem 'facterdb', '1.4.0',           :require => false
 end
 
 group :development do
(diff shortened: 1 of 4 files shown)


alexjfisher:
It's not just the gpg key.

https://oss-binaries.phusionpassenger.com/yum/passenger/el/6/x86_64/ contains nginx packages, but https://oss-binaries.phusionpassenger.com/yum/passenger/el/7/x86_64/ doesn't.

alexjfisher:
https://blog.phusion.nl/2020/05/29/passenger-6-0-5/

"Passenger 6.0.5 also changes how the Nginx packages on CentOS 7 work. We now provide a dynamic module instead of a full Nginx install, which integrates with the Nginx package available from the EPEL repo."

Gemfile
```diff
@@ -11,10 +11,11 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
+  gem 'voxpupuli-test', '< 2.0.0',   :require => false
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
-  gem 'rspec-puppet-facts',          :require => false, :git => 'https://github.com/mcanevet/rspec-puppet-facts', :ref => '9541292d4fc35db3be1badace673c1108154b571'
+  gem 'rspec-puppet-facts', '2.0.0', :require => false
+  gem 'facterdb', '1.4.0',           :require => false
```
alexjfisher, line 18: The next modulesync will overwrite this file, but these are the versions we'll end up with.  I needed to get them in here for the build not to hit the 50 minute travis timeout.

igalic (approved):

PR #1410, opened 2020-08-29
Refactor to remove some `File` defaults

alexjfisher:
Resource defaults outside of site.pp are not recommended.
https://puppet.com/docs/puppet/6.18/lang_defaults.html

In these cases, there was only a single file resource in each manifest.

### DIFF
--- a/manifests/resource/map.pp
+++ b/manifests/resource/map.pp
@@ -95,16 +95,12 @@ define nginx::resource::map (
     default  => 'file',
   }
 
-  File {
-    owner => 'root',
-    group => $root_group,
-    mode  => $nginx::global_mode,
-  }
-
   file { "${conf_dir}/${name}-map.conf":
     ensure  => $ensure_real,
+    owner   => 'root',
+    group   => $root_group,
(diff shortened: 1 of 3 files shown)


manifests/resource/map.pp
```diff
@@ -95,16 +95,12 @@
...
-  }
-
   file { "${conf_dir}/${name}-map.conf":
     ensure  => $ensure_real,
+    owner   => 'root',
+    group   => $root_group,
+    mode    => $nginx::global_mode,
     content => template('nginx/conf.d/map.erb'),
     notify  => Class['nginx::service'],
-    require => File[$conf_dir],
```
alexjfisher, line 108: This will be autorequired.

bastelfreak (approved):

igalic:
> Resource defaults outside of site.pp are not recommended.
> https://puppet.com/docs/puppet/6.18/lang_defaults.html

from that document:


> Therefore, they not be set outside of site.pp. Use per-resource default attributes when possible.

That document sounds like it needs a fix

PR #1411, opened 2020-08-30
Rename type alias tests to end `_spec.rb`

alexjfisher:
Not sure how no one noticed this before, but these tests weren't being
run.


spec/type_aliases/upstreamzone_spec.rb
```diff
@@ -1,7 +1,6 @@
 require 'spec_helper'
 
 describe 'Nginx::UpstreamZone' do
-  it { is_expected.to compile.with_all_deps }
```
alexjfisher, line 4: This wasn't valid here and failed. (Which is how I know for sure that these weren't being run)

ekohl (approved):

PR #1413, opened 2020-08-30
Add stricter parameter types

ekohl:
This includes a rebase of https://github.com/voxpupuli/puppet-nginx/pull/1189. That adds a type Nginx::Duration which looks similar to Nginx::Time (added in f0bf83a1abac6c2ee7fe7257a37514319a96f0fa). We likely want to unify those, but right now I'm not sharp enough to see the exact differences.


manifests/init.pp
```diff
@@ -185,26 +185,26 @@
...
   ### END Package Configuration ###
 
   ### START Service Configuation ###
-  $service_ensure                                            = running,
-  $service_enable                                            = true,
-  $service_flags                                             = undef,
-  $service_restart                                           = undef,
-  $service_name                                              = 'nginx',
-  $service_manage                                            = true,
+  Enum['running', 'absent', 'stopped', 'undef'] $service_ensure = running,
```
igalic, line 202: what's undef doing here?

manifests/service.pp
```diff
@@ -14,12 +14,12 @@
...
 # This class file is not called directly
 class nginx::service (
-  $service_restart = $nginx::service_restart,
-  $service_ensure  = $nginx::service_ensure,
-  $service_enable  = $nginx::service_enable,
-  $service_name    = $nginx::service_name,
-  $service_flags   = $nginx::service_flags,
-  $service_manage  = $nginx::service_manage,
+  Optional[String[1]] $service_restart                          = $nginx::service_restart,
+  Enum['running', 'absent', 'stopped', 'undef'] $service_ensure = $nginx::service_ensure,
```
igalic, line 18: what's undef doing here?

igalic (commented):
👀

manifests/init.pp
ekohl: Copied from the previous PR. Haven't looked too closely at the content yet.

ekohl:
https://github.com/voxpupuli/puppet-nginx/pull/1414 should probably be merged first.

manifests/service.pp
ekohl: In https://github.com/voxpupuli/puppet-nginx/pull/1414 I chose `Stdlib::Ensure::Service`

manifests/resource/mailhost.pp
```diff
@@ -76,49 +76,49 @@
...
-  Optional[String] $smtp_auth                    = undef,
-  Optional[Array] $smtp_capabilities             = undef,
-  Optional[Variant[Array, String]] $raw_prepend  = undef,
-  Optional[Variant[Array, String]] $raw_append   = undef,
-  Optional[Hash] $mailhost_cfg_prepend           = undef,
-  Optional[Hash] $mailhost_cfg_append            = undef,
-  String $proxy_pass_error_message               = 'off',
-  Array $server_name                             = [$name]
+  Enum['absent', 'present'] $ensure                                = 'present',
+  Variant[Enum['*'], Array[Stdlib::Ipv4], Stdlib::Ipv4] $listen_ip = '*',
```
ekohl, line 80: These should be `Stdlib::IP::Address::V4::Nosubnet`. Same below for IPv4

bastelfreak (approved):

ekohl:
Converting to draft since I'm using Travis instead of local tests. Feel free to point out errors or mistakes, but no need to approve until it's green.

types/duration.pp
```diff
@@ -0,0 +1 @@
+type Nginx::Duration = Pattern[/^((\d+y)?\s?(\d+M)?\s?(\d+w)?\s?(\d+d)?\s?(\d+h)?\s?(\d+m)?\s?(\d+s)?\s?(\d+ms)?\s?|\d+)$/]
```
igalic, line 1: why is there a trailing `\s?`?

igalic (commented):
🙋🏻‍♀️

types/duration.pp
ekohl: diffing this regex with the stricter pattern that's `Nginx::Time` is still on my todo list.

ekohl: I added a commit that merges it into Nginx::Time. That also copies the Nginx docs. While it now allows a lot of additional whitespace, it's probably easier to read this way. The various suffixes can be combined and I can't see how the regex can be really improved.

ekohl:
Note to self: disable `Style/BracesAroundHashParameters` per https://github.com/voxpupuli/modulesync_config/commit/01141b273bab9f314cfcb66840f5365e31fd593c

TheMeier:
@ekohl I think this is obsolete now.

PR #1414, opened 2020-08-31
Simplify service class

ekohl:
It's valid to pass in undef values. That makes the OpenBSD check redundant. The same goes for the restart command. That removes pretty much all logic so there is no more benefit to having class parameters. That's why it now uses all the variables directly from the nginx class. The has{status,restart} parameters default to true since Puppet 2.7 so they can be removed.

Replaces https://github.com/voxpupuli/puppet-nginx/pull/1209

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -199,7 +199,7 @@ class nginx (
   ### END Package Configuration ###
 
   ### START Service Configuation ###
-  $service_ensure                                            = running,
+  Stdlib::Ensure::Service $service_ensure                    = 'running',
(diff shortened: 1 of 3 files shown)


bastelfreak (approved):

igalic (approved):

bastelfreak (approved):

ekohl:
And now with a fix to make rubocop happy.

bastelfreak (approved):

ekohl:
Finally green! That took too many tries for such a simple PR :P

PR #1419, opened 2020-10-28
implement config check on config change before service restart

crazymind1337:
This implements a switch to enable the nginx config check via `nginx -t` on all config file changes. If the config check is enabled (disabled by default) and fails, the service refresh of nginx will be unscheduled b puppet.

No issue created, started directly with the feature.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,6 +1,5 @@
-# Class: nginx
-#
-# This module manages NGINX.
+# @summary
+#   This module manages NGINX.
 #
 # Parameters:
 #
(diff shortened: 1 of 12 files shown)


bastelfreak:
hey @crazymind1337, thanks for the PR! can you add an acceptance test that sets this and also fix the puppet-lint warning?

manifests/init.pp
```diff
@@ -205,6 +205,7 @@
   $service_restart                                           = undef,
   $service_name                                              = 'nginx',
   $service_manage                                            = true,
+  Boolean $service_config_check                              = false,
```
bastelfreak, line 208: Please add puppet-strings documentation for this new parameter :)

crazymind1337: Done.

bastelfreak (approved):
looks good to me. We can merge this when the tests pass

crazymind1337:
Tests passed.

PR #1431, opened 2021-01-12
(maint) bump apt to 7.7.0 for performance tests

gimmyxd:
Facter 4 changed how the code from custom facts is evaluated,
this introuced a couple of changes needed for apt module,
more details here: https://github.com/puppetlabs/puppetlabs-apt/commit/6786d506e373c08df8ad0b00c5a9d7275ba0ed1a

Bump apt module to 7.7.0 to resolve performance issues when
running unit tests on Linux

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,8 @@
 fixtures:
   repositories:
-    apt: "https://github.com/puppetlabs/puppetlabs-apt.git"
+    apt:
+      repo: "https://github.com/puppetlabs/puppetlabs-apt.git"
+      ref: "v7.7.0"
     concat: "https://github.com/puppetlabs/puppetlabs-concat.git"
     stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
     yumrepo_core:


.fixtures.yml
```diff
@@ -1,6 +1,8 @@
 fixtures:
   repositories:
-    apt: "https://github.com/puppetlabs/puppetlabs-apt.git"
+    apt:
+      repo: "https://github.com/puppetlabs/puppetlabs-apt.git"
+      ref: "v7.7.0"
```
ekohl, line 5: Shouldn't the commit be in git as well?

gimmyxd: It is, i just added the latest released version to see how it performs.

ekohl: But how are you going to determine a baseline for this repository? Since it was merged in April, all CI should have run with it for months already. Are you going to open a PR pinning it to puppetlabs-apt v7.4.2?

gimmyxd: I was looking at the wrong logs, this was already running with the latest module ref. So this change is pointless 😅

ekohl:
Closing it then.

PR #1432, opened 2021-01-13
Replace Travis build status in README

saz:
Github Actions is used by this module.
This way, the real build status is shown


saz:
As this is only changing the README, the failing tests are unrelated.

ekohl (approved):
CentOS 6 is failing because the CentOS mirrors no longer have the 6 bits so that's expected. We are slowly dropping it because of that. As you say, it's completely unrelated to this change.

This is actually something we should do in bulk in all our repositories.

PR #1435, opened 2021-01-20
Add two more lingering_ config options

sbeaulie:
Adding two more lingering_ config options
lingering_close
lingering_time
See https://nginx.org/en/docs/http/ngx_http_core_module.html#lingering_timeout

n/a

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -93,6 +93,8 @@ class nginx (
   $client_body_timeout                                       = '60s',
   $send_timeout                                              = '60s',
   $lingering_timeout                                         = '5s',
+  Optional[Enum['on','off','always']] $lingering_close       = undef,
+  Optional[String[1]] $lingering_time                        = undef,
   Optional[Enum['on', 'off']] $etag                          = undef,
   Optional[String] $events_use                               = undef,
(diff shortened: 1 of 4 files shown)


genebean:
The two failures for CentOS 6 seem to be due to the CentOS 6 base repo going away (http://mirror.centos.org/centos-6/6/readme)

genebean:
Once CI goes green on https://github.com/voxpupuli/puppet-nginx/pull/1436 this PR can be rebased and testing here _should_ also be good.

genebean (approved):

manifests/init.pp
```diff
@@ -93,6 +93,8 @@
   $client_body_timeout                                       = '60s',
   $send_timeout                                              = '60s',
   $lingering_timeout                                         = '5s',
+  Optional[String] $lingering_close                          = undef,
```
bastelfreak, line 96: can we enforce the string a bit more? maybe at least `Optional[String[1]]`?

sbeaulie: done.
for lingering_close list the 3 valid options as enum
for lingering_time enforce a non empty string

genebean: Good catch @bastelfreak - I totally missed that.

genebean (approved):

PR #1436, opened 2021-01-20
Drop EL6 & Puppet 5 from and add Puppet 7 to metadata

genebean:

This PR drops Red Hat 6 and CentOS 6 from the metadata as those OS's are end of life and their yum repositories have been archived. It also raises the upper bounds of acceptable Puppet versions to include Puppet 7 and raises the lower bounds to Puppet 6.1 since Puppet 5 is also EoL.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -20,7 +20,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

attachmentgenie:
while you are at it could you drop puppet 5 as well?

bastelfreak (approved):

PR #1437, opened 2021-01-22
Use $service_name in nginx config check command

saz:
Config check isn't working, if you're running e.g. openresty as the service name is `openresty`, not `nginx`.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -7,7 +7,7 @@ class nginx::service {
 
   if $nginx::service_config_check {
     exec { 'nginx_config_check':
-      command     => 'nginx -t',
+      command     => "${nginx::service_name} -t",
       refreshonly => true,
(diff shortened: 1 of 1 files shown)


manifests/service.pp
```diff
@@ -7,7 +7,7 @@
 
   if $nginx::service_config_check {
     exec { 'nginx_config_check':
-      command     => 'nginx -t',
+      command     => "${nginx::service_name} -t",
```
ekohl, line 10: The command doesn't have to be the same as the service, right? For example, it may be possible that someone has `nginx.service@myinstance` if the run multiple services using systemd. Perhaps it deserves its own parameter?

saz: You're right. It's just matching for me. A separate parameter will be the best way to handle that.

PR #1438, opened 2021-01-22
makes service_config_check exec command configurable

saz:
Right now, the service_config_check exec command is using `nginx -t` and there's no way to change it.
If you're running e.g. openresty, there's no `nginx` executable and the check will fail.

This PR adds a paramter `$nginx::service_config_check_command` with a default value of `nginx -t`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -31,6 +31,9 @@
 # @param service_config_check
 #  whether to en- or disable the config check via nginx -t on config changes
 #
+# @param service_config_check_command
+#  Command to execute to validate the generated configuration.
+#
 class nginx (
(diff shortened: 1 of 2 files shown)


genebean (approved):

saz:
@ekohl Mind taking a look at this? As you've commented on my last PR related to that topic.

ekohl (approved):
I'd appreciate a puppet-strings parameter documentation. While most parameters are missing, let's try to add them for any new ones. Code wise :+1:

saz:
@ekohl Good point, added.

ekohl (approved):
Thanks!

PR #1445, opened 2021-03-05
Fixes an error on ubuntu >= 18.04 when using nginx apt repository

theq86:
Package manager would show: N: Skipping acquire of configured file 'nginx/binary-i386/Packages' as repository 'http://nginx.org/packages/ubuntu bionic InRelease' doesn't support architecture 'i386'
Adding correct architecture fixes the error

On Ubuntu 18.04 and later 20.04 I always got following message on 'apt update':
`N: Skipping acquire of configured file 'nginx/binary-i386/Packages' as repository 'http://nginx.org/packages/ubuntu bionic InRelease' doesn't support architecture 'i386'`

My PR changes the apt:source resource for amd64 ubuntu instances, to include the architecture of the distribution in the sources file, as well, eliminating the above error and allowing nginx to be updated correctly.


kenyon:
Would be great if you could enhance the tests in https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb to cover the Ubuntu case.

bastelfreak:
can you maybe share the puppet code you use to call the nginx moduke or provide an acceptance test that triggers this? I have a few ubuntu 16.04 machines and they didnt shiw the error in the past. On monday I should be able to access then agaun and check the config.

bastelfreak:
So we enable the upstream repos per default on debian based platforms: 
https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/params.pp#L111-L121
and the current acceptance tests on Debian/Ubuntu already pass.

theq86:
> can you maybe share the puppet code you use to call the nginx moduke or provide an acceptance test that triggers this?

Since I'm not really able to write acceptance tests I will provide you with the profile I wrote to use the nginx module

```
# Class for the nginx profile
class profile::nginx (
  $client_max_body_size = '16m'
) {
  include profile::base

  $dhparams_path = '/etc/ssl/dhparams.pem'

  file { $dhparams_path:
    ensure => 'present',
    mode   => '0644',
    owner  => 'root',
    group  => 'root',
    source => 'puppet:///modules/base/ssl/dhparams.pem',
    before => Class['nginx'],
  }

  file { '/var/nginx' :
    ensure => 'directory',
    group  => 'root',
    mode   => '0644',
  }

  class { 'nginx':
    server_tokens         => 'off',
    http2                 => 'on',
    ssl_protocols         => 'TLSv1.2 TLSv1.3',
    ssl_ciphers           => 'EECDH+ECDSA+AESGCM:EECDH+aRSA+AESGCM:EECDH+ECDSA+SHA512:EECDH+ECDSA+SHA384:EECDH+ECDSA+SHA256:ECDH+AESGCM:ECDH+AES256:DH+AESGCM:DH+AES256:RSA+AESGCM:!aNULL:!eNULL:!LOW:!RC4:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS',
    ssl_dhparam           => $dhparams_path,
    server_purge          => true,
    gzip                  => 'on',
    client_body_temp_path => '/var/nginx/client_body_temp',
    client_max_body_size  => $client_max_body_size,
    proxy_temp_path       => '/var/nginx/proxy_temp',
    require               => File['/var/nginx'],
  }

  file { "${nginx::conf_dir}/conf.d/default.conf":
    ensure => 'absent'
  }
}
```

However, as you already stated, for debian/ubuntu the upstream repos are used by default.

Problem is on 18.04 and 20.04, that apt fails with the error from the first post. It can be solved by including the architecture into the apt sources file. On 16.04 as I know this does not occur.

theq86:
I will close this pull request because it does not remove the actual problem and is unneccessary on other machines than mine.

Why did the message from above occur?

Remember, it was:
```
N: Skipping acquire of configured file 'nginx/binary-i386/Packages' as repository 'http://nginx.org/packages/ubuntu bionic InRelease' doesn't support architecture 'i386'
```

Apparently, the nginx repository does not provide i386 packages and this is what the message tells us about. But why? I am not on an i386 machine. Having no i386 packages is neither nginx vendor's problem, nor a problem of this puppet module.

So I went to my apt configuration - more precisely, my dpkg configuration - and guess what:

`dpkg --print-architecture` ==> _amd64_

But dpkg can have multiarch enabled. This is disabled by default, but somehow the ubuntu image that my server host installs on new machines has additional architectures installed:

`dpkg --print-foreign-architectures` ==> _i386_

Wow, what has that to do there? I don't know. But I needed to check twice before getting rid of it (maybe it is just there for reason)

`dpkg -l | grep i386` ==> 

Nothing? Great, so I don't know why i386 is enabled, but we do not need it and we probably never will.

So, what's left ?

```
dpkg --remove-architecture i386
```

This will remove the unwanted architecture. And now dpkg/apt will not look for such packages and apt will no more complain about  that.

PR #1446, opened 2021-03-22
allow configuration of multiple ssl certificates and keys

### LINKED ISSUE #1434: Support hybrid RSA and ECDSA configuration

- Puppet: Any
- Ruby: Any
- Distribution: Any
- Module version: Any

Attempt to specify more than one ssl_certificate_key and ssl_certificate/chain/fullchain fails.

## What are you seeing
Attempting to specify more than one ssl_certificate_key and ssl_certificate/chain/fullchain fails.

These parameters should be Boolean|String|Array rather than the current Boolean|String, and create one ssl_* configuration line for each entry in the array.

https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_certificate

saz:
Allow configuration of multiple ssl_certificate and ssl_certificate_key lines

Fixes #1434


manifests/resource/server.pp
```diff
@@ -592,8 +594,29 @@
   if $ssl {
     # Access and error logs are named differently in ssl template
 
-    File <| title == $ssl_cert or path == $ssl_cert or title == $ssl_key or path == $ssl_key |>
-    -> concat::fragment { "${name_sanitized}-ssl-header":
+    if $ssl_key.is_a(Array) {
```
kenyon, line 597: Should we prefer the built-in `=~` operator these days over the stdlib `is_a` function?

manifests/resource/server.pp
```diff
@@ -592,8 +594,29 @@
...
+      $ssl_key.each | $key | {
+        File <| title == $key or path == $key |> {
+          before => Concat::Fragment["${name_sanitized}-ssl-header"],
+        }
+      }
+    } else {
+      File <| title == $ssl_key or path == $ssl_key |> {
+        before => Concat::Fragment["${name_sanitized}-ssl-header"],
+      }
+    }
```
kenyon, line 607:
```suggestion
    $ssl_key.flatten.each |$key| {
      File <| title == $key or path == $key |> {
        before => Concat::Fragment["${name_sanitized}-ssl-header"],
      }
    }
```

Hmm, would something like this work to reduce code duplication?

saz: $ssl_key might be an Array, String or Boolean. I'm not sure, if this code is working like that.

saz: I'm not able to find a better way, as there is no way, I'm aware of, to use a variable for the collectors. E.g. `File <| $filter |>` isn't working. `File <| title = $array_var |>` isn't working either.

I'm open for suggestions here, as I don't like this part either.

saz: I think I've found a better way.

manifests/resource/server.pp
```diff
@@ -592,8 +594,31 @@
   if $ssl {
     # Access and error logs are named differently in ssl template
 
-    File <| title == $ssl_cert or path == $ssl_cert or title == $ssl_key or path == $ssl_key |>
-    -> concat::fragment { "${name_sanitized}-ssl-header":
+    if $ssl_key {
+      if $ssl_key =~ Array {
+        $ssl_key_real = $ssl_key
+      } else {
+        $ssl_key_real = [$ssl_key]
```
kenyon, line 601: OK but this is doing the same as what [`flatten()`](https://puppet.com/docs/puppet/7.5/function.html#flatten) does like I suggested, but in a more complicated way. :shrug:

saz: Sorry for wasting your time. While trying your suggestion, I checked also for the Boolean type, which isn't required at all.
I've updated the commit again.

kenyon (approved):

PR #1451, opened 2021-04-20
Add variables to overwrite fastcgi.conf and uwsgi_params template

TuningYourCode:
This PR introduces variables to overwrite the fastcgi.conf and uwsgi_params templates.

Relates to #862

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,7 +69,9 @@ class nginx (
   Boolean $include_modules_enabled                           = $nginx::params::include_modules_enabled,
 
   # Primary Templates
-  $conf_template                                             = 'nginx/conf.d/nginx.conf.erb',
+  String[1] $conf_template                                   = 'nginx/conf.d/nginx.conf.erb',
+  String[1] $fastcgi_conf_template                           = 'nginx/server/fastcgi.conf.erb',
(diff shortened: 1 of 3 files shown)


TuningYourCode:
Could anybody help with supplying a test? The already existing `$conf_template` seems to also have no test for the template overwrite only checks for template variables - which we do not have in these templates (yet).

manifests/init.pp
```diff
@@ -70,6 +70,8 @@
 
   # Primary Templates
   $conf_template                                             = 'nginx/conf.d/nginx.conf.erb',
+  $fastcgi_conf_template                                     = 'nginx/server/fastcgi.conf.erb',
+  $uwsgi_params_template                                     = 'nginx/server/uwsgi_params.erb',
```
bastelfreak, line 74: please add a datatype for both parameters:
```suggestion
  Stdlib::Absolutepath $fastcgi_conf_template                = 'nginx/server/fastcgi.conf.erb',
  Stdlib::Absolutepath $uwsgi_params_template            = 'nginx/server/uwsgi_params.erb',
```

bastelfreak: ah lol, my bad. I think `String[1]` is fine here.

PR #1456, opened 2021-05-26
default ipv6_listen_port to listen_port

bastelfreak:
in 99% of the deployments both ports are identical. to make this easier
for the users we can just tell puppet that the default value for
$ipv6_listen_port is the value of $listen_port. That makes it easier for
people that want their dual stack application to listen on a nonstandard
port. They only need to set it once.


kenyon (approved):

kenyon:
I'm approving the idea, of course need passing tests ;)

kenyon (approved):

PR #1459, opened 2021-07-20
Add support for AppStream package installation

Henrik-Hansson:
Add support to install nginx through AppStream on EL8.


manifests/init.pp
```diff
@@ -196,6 +196,7 @@
   $passenger_package_ensure                                  = 'present',
   String[1] $passenger_package_name                          = $nginx::params::passenger_package_name,
   Optional[Stdlib::HTTPUrl] $repo_source                     = undef,
+  Optional[Float] $dnfmodule                                 = undef,
```
smortex, line 199: I would not use a float to represent a package version… You know:

```puppet
0.1 * 0.2 #=> 0.020000000000000004
```

manifests/package/redhat.pp
```diff
@@ -94,6 +95,14 @@
     }
   }
 
+  if $dnfmodule and ($facts['os']['name'] in ['RedHat', 'CentOS', 'VirtuozzoLinux']) and ($facts['os']['release']['major'] in ['8']) {
```
smortex, line 98: Prefer `fact('os.name')` ([from stdlib](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/REFERENCE.md#fact)) or `$facts.get('os.name')` ([from Puppet core](https://puppet.com/docs/puppet/7/function.html#get)) over `$facts['os']['name']`.

Also, regarding version comparison, unless you are sure version 9, 10 and so on will not be concerned, better rely on [`versioncmp`](https://puppet.com/docs/puppet/7/function.html#versioncmp), e.g. `versioncmp(fact('os.release.full'), '8.0') >= 0`

manifests/init.pp
Henrik-Hansson: I have changed from Float to String.

manifests/package/redhat.pp
Henrik-Hansson: I changed to fact('os.name') and as EL 9, 10 etc. is not released I took your suggestion with versioncmp.

anders-larsson:
It seems this pull request needs additional work. Currently it attempts to install the "package" every time Puppet is executed.

manifests/package/redhat.pp
```diff
@@ -94,6 +95,14 @@
     }
   }
 
+  if $dnfmodule and fact('os.name') in ['RedHat', 'CentOS', 'VirtuozzoLinux'] and versioncmp(fact('os.release.full'), '8.0') >= 0 {
```
ekohl, line 98:
```suggestion
  if $dnfmodule and fact('os.name') in ['RedHat', 'CentOS', 'VirtuozzoLinux'] and versioncmp(fact('os.release.major'), '8') >= 0 {
```

ekohl (commented):
Maybe it helps, but in our Candlepin module we use `enable_only => true`: https://github.com/theforeman/puppet-candlepin/blob/27a0f4d4de5d2180309db05a508041d997139739/manifests/install.pp#L14-L21

I'm not sure if that can also switch modules and haven't looked at the details if I'm honest.

manifests/package/redhat.pp
```diff
@@ -94,6 +95,15 @@
     }
   }
 
+  if $dnfmodule and fact('os.name') in ['RedHat', 'CentOS', 'VirtuozzoLinux'] and versioncmp(fact('os.release.major'), '8') >= 0 {
```
ekohl, line 98: Would it make sense to use `os.family` instead?

ekohl (approved):
Looks good to me. Let me know if you want to address my comment. Without it it's already a good change IMHO, but it can help other RHEL clones.

Henrik-Hansson:
I agree that os.family seems better. I will rebase and update with os.family.

ekohl (approved):
Thanks!

ekohl:
I think https://github.com/voxpupuli/puppet-nginx/pull/1501 should fix the acceptance tests, except for EL7. But that shouldn't hold up this PR. Unit tests should suffice for that (but those are still running).

Henrik-Hansson:
Hello, what can I do to get this merged?
The CI validation fails on something not related to this PR.
BR,
Henrik

ekohl:
I've opened https://github.com/voxpupuli/puppet-nginx/pull/1519 to fix that.

ekohl:
Please rebase.

Henrik-Hansson:
Hi, sorry for delay.
Most of the tests fails on passenger, don't think it's related to my change.
I have fixed the Package error I introduced with "Package['nginx']", with "'Package[nginx]'".

manifests/init.pp
```diff
@@ -196,6 +196,7 @@
   $passenger_package_ensure                                  = 'present',
   String[1] $passenger_package_name                          = $nginx::params::passenger_package_name,
   Optional[Stdlib::HTTPUrl] $repo_source                     = undef,
+  Optional[String[1]] $dnfmodule                             = undef,
```
kenyon, line 199: need documentation added for this new parameter

kenyon (changes requested):

ardrigh:
@kenyon is it possible for you to do a fresh review for this to get merged?

It appears the documentation has been added, and this is much needed improvement for RHEL servers.

kenyon (approved):

kenyon:
#1573 removes EOL Ubuntu 18.04.

ardrigh:
@Henrik-Hansson if you are still keen to update this pull request, the version check could use a tweak.

RHEL9 started with 1.20, a version check for this should fix the tests.

It might be easier to assume a test for nginx 1.22 that is available in both RHEL8.8 and RHEL9.2 appstreams?

PR #1439, opened 2021-01-22
Add nginx config for stream log

genebean:
#### Pull Request (PR) description

This adds the ability to configure logging for stream configuration in Nginx, like getting IP addresses from access logging.

This PR includes the changes in #1401. That PR was built on #1336. This PR simply pulls in the changes atop the current version of master.

closes #1401
closes #1336

PR #1461, opened 2021-07-30
Add stream log support

ardrigh:
Following on work from previous PRs to add stream log changes to the module.

Updated changes to fix file conflicts and newer changes in v4.1.0
 
The configuration in Hiera:

```yaml
nginx::stream_access_log: '/var/log/nginx/nginx-stream-access.log'
nginx::stream_custom_format_log: 'basic'
nginx::stream_log_format:
  basic: '$remote_addr [$time_local] $protocol $status $bytes_sent $bytes_received $session_time $server_addr:$server_port $upstream_addr'
```

Adds the following lines to nginx.conf:

```shell
+  log_format basic '$remote_addr [$time_local] $protocol $status $bytes_sent $bytes_received $session_time $server_addr:$server_port $upstream_addr';
 
+  access_log /var/log/nginx/nginx-stream-access.log basic;
```

If `stream_custom_format_log` is not set, nginx will use the built-in `combined` log format for the stream access log.

#### This Pull Request (PR) fixes the following issue

Closes: https://github.com/voxpupuli/puppet-nginx/pull/1439

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -66,6 +66,7 @@ class nginx (
   Variant[String, Array[String]] $http_access_log            = "${log_dir}/${nginx::params::http_access_log_file}",
   Optional[String] $http_format_log                          = undef,
   Variant[String, Array[String]] $stream_access_log          = "${log_dir}/stream-access.log",
+  Optional[String] $stream_custom_format_log                 = undef,
   Variant[String, Array[String]] $nginx_error_log            = "${log_dir}/${nginx::params::nginx_error_log_file}",
   Nginx::ErrorLogSeverity $nginx_error_log_severity          = 'error',
   $pid                                                       = $nginx::params::pid,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -65,6 +65,7 @@ class nginx (
   Stdlib::Filemode $log_mode                                 = $nginx::params::log_mode,
   Variant[String, Array[String]] $http_access_log            = "${log_dir}/${nginx::params::http_access_log_file}",
   Optional[String] $http_format_log                          = undef,
+  Variant[String, Array[String]] $stream_access_log          = "${log_dir}/stream-access.log",
   Variant[String, Array[String]] $nginx_error_log            = "${log_dir}/${nginx::params::nginx_error_log_file}",
   Nginx::ErrorLogSeverity $nginx_error_log_severity          = 'error',
   $pid                                                       = $nginx::params::pid,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -124,6 +124,7 @@ class nginx (
   $keepalive_timeout                                         = '65s',
   $keepalive_requests                                        = '100',
   Hash[String[1], Nginx::LogFormat] $log_format              = {},
+  Hash[String[1], Nginx::LogFormat] $stream_log_format       = {},
   Boolean $mail                                              = false,
   Variant[String, Boolean] $mime_types_path                  = 'mime.types',
   Boolean $stream                                            = false,
--- a/spec/classes/nginx_spec.rb
+++ b/spec/classes/nginx_spec.rb
@@ -1541,10 +1541,36 @@ describe 'nginx' do
               )
             end
 
-            it do
-              is_expected.to contain_file('/etc/nginx/nginx.conf').with_content(
-                %r{access_log\s/var/log/nginx/stream-access.log;}
-              )
+            context 'when stream_log_format is defined' do
+              let(:params) do
+                super().merge({ stream_log_format: { 'stream_format' => 'STREAM_FORMAT' } })
+              end
+
+              it do
+                is_expected.to contain_file('/etc/nginx/nginx.conf').with_content(
+                  %r{log_format stream_format 'STREAM_FORMAT';}
+                )
+              end
+            end
+
+            context 'when stream_custom_format_log is default' do
+              it do
+                is_expected.to contain_file('/etc/nginx/nginx.conf').with_content(
+                  %r{access_log /var/log/nginx/stream-access.log;}
+                )
+              end
+            end
+
+            context 'when stream_custom_format_log is non-default' do
+              let(:params) do
+                super().merge({ stream_custom_format_log: 'stream_format' })
+              end
+
+              it do
+                is_expected.to contain_file('/etc/nginx/nginx.conf').with_content(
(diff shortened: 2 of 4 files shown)


manifests/init.pp
```diff
@@ -58,6 +58,8 @@
   Stdlib::Filemode $log_mode                                 = $nginx::params::log_mode,
   Variant[String, Array[String]] $http_access_log            = "${log_dir}/${nginx::params::http_access_log_file}",
   Optional[String] $http_format_log                          = undef,
+  Variant[String, Array[String]] $stream_access_log          = "${log_dir}/${nginx::params::stream_access_log_file}",
```
bastelfreak, line 61: Can you please make the type more strict?
```suggestion
  Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $stream_access_log = "${log_dir}/${nginx::params::stream_access_log_file}",
```

bastelfreak: And Instead of hardcoding the data in the params.pp, we can directly set it in the init.pp
```suggestion
  Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $stream_access_log = "${log_dir}/stream-access.log",
```

manifests/init.pp
```diff
@@ -117,6 +119,7 @@
   $keepalive_timeout                                         = '65s',
   $keepalive_requests                                        = '100',
   $log_format                                                = {},
+  $stream_log_format                                         = {},
```
bastelfreak, line 122: Please enforce the datatype here as well. I guess something like this should work:
```suggestion
  Hash[String[1], String[1]] $stream_log_format              = {},
```

templates/conf.d/nginx.conf.erb
```diff
@@ -326,5 +326,19 @@ stream {
 <% unless @confd_only -%>
   include <%= @conf_dir %>/streams-enabled/*;
 <% end -%>
+
+<% if @stream_log_format -%>
```
bastelfreak, line 330: can you test if the if statement is required? The variable always exists, by default it's just empty. I think we can directly start with the .sort_by call

ekohl: What could make sense is
```ruby
<% end -%>
<% unless @stream_log_format.empty? -%>

<% @stream_log_format.sort_by{|k,v| k}.each do |key,value| -%> 
```

With this you can prevent the empty lines in the generated config. Because right now it generates (roughly):
```
  include .../streams-enabled/*;


  access_log ...
```

ardrigh:
I am working on making updates to fix this PR. I need to check it works with what we have deployed.

I think the tests are failing because they are in the wrong location in the test file. I think the test lines might be better in a context for `stream` specifically.

Would it work adding a "dynamic module" test similar to geoip module, checking the `stream` content lines for the `access_log` and `log_format` values?

geoip test example:
https://github.com/voxpupuli/puppet-nginx/blob/fffe7a3463e005944799f55f8782e9ba798375a8/spec/classes/nginx_spec.rb#L1162

spec/classes/nginx_spec.rb
```diff
@@ -566,6 +566,28 @@
                 value: {},
                 notmatch: %r{log_format}
               },
+              {
+                title: 'should set stream log formats',
+                attr: 'stream_log_format',
+                value: {
+                  'format1' => 'FORMAT1',
+                  'format2' => 'FORMAT2'
```
ekohl, line 574:
```suggestion
                  'format2' => 'FORMAT2',
```

templates/conf.d/nginx.conf.erb
```diff
@@ -330,5 +330,19 @@ stream {
 <% unless @confd_only -%>
   include <%= @conf_dir %>/streams-enabled/*;
 <% end -%>
+
+<% unless @stream_log_format.empty? -%>
```
ekohl, line 334: This usually renders poorly, but I'd do it the other way around: first the `if` or `unless`, then the empty line. That way you don't end up with a lot of empty lines if the block doesn't render.
```suggestion
<% unless @stream_log_format.empty? -%>

```

templates/conf.d/nginx.conf.erb
```diff
@@ -330,5 +330,19 @@ stream {
...
+<% @stream_log_format.sort_by{|k,v| k}.each do |key,value| -%>
```
ekohl, line 335: Pretty sure this is the default `sort` behavior, but not 100% sure.
```suggestion
<% @stream_log_format.sort.each do |key,value| -%>
```

templates/conf.d/nginx.conf.erb
```diff
@@ -330,5 +330,19 @@ stream {
...
+<% end -%>
+<% end -%>
+
+<% if @stream_access_log.is_a?(Array) -%>
+  <%- @stream_access_log.each do |log_item| -%>
+  access_log <%= log_item %><% if @stream_custom_format_log %> <%= @stream_custom_format_log%><% end %>;
+  <%- end -%>
+<% else -%>
+  access_log <%= @stream_access_log %><% if @stream_custom_format_log %> <%= @stream_custom_format_log%><% end %>;
+<% end -%>
```
ekohl, line 346: I think you can use the `Array()` method to ensure it's an array:
```suggestion
<%- Array(@stream_access_log).each do |log_item| -%>
  access_log <%= log_item %><% if @stream_custom_format_log %> <%= @stream_custom_format_log%><% end %>;
<% end -%>
```

ardrigh:
Updated with suggested changes

ekohl:
From a quick look this looks good, but the tests fail. The CentOS 7 tests failing looks unrelated to me, but the unit tests do.

ardrigh:
> I am working on making updates to fix this PR. I need to check it works with what we have deployed.
> 
> I think the tests are failing because they are in the wrong location in the test file. I think the test lines might be better in a context for `stream` specifically.

Still the same issue with the unit tests. Which isn't surprising since I did a quick copy & paste to get started.

I don't see anything that tests

> <% if @stream -%>

And without that being set, it's not expected the lines get written to the config file, and the unit tests correctly fail.

These tests might not cleanly fit within the existing template iterator.

ardrigh:
@ekohl I have added a basic set of tests to cover the main use case for the stream access log.

bastelfreak (approved):

ekohl (approved):
Thanks!

PR #1465, opened 2021-08-25
Allow stdlib 8.0.0

smortex:


ekohl (approved):
I'm going to merge this since I can't imagine how the changes could break CentOS 7.

PR #1466, opened 2021-08-26
modulesync 4.2.0

bastelfreak:
modulesync 4.2.0

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,7 @@
 # editorconfig.org
 
-# MANAGED BY MODULESYNC
+# Managed by modulesync - DO NOT EDIT
+# https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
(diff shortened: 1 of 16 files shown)


manifests/init.pp
```diff
@@ -225,7 +225,7 @@
   Hash $nginx_upstreams                                   = {},
   Nginx::UpstreamDefaults $nginx_upstreams_defaults       = {},
   Boolean $purge_passenger_repo                           = true,
-  String[1] $nginx_version                                = pick(fact('nginx_version'), '1.6.0'),
+  String[1] $nginx_version                                = pick(fact('nginx_version'), '1.20.0'),
```
ekohl, line 228: This looks like an unrelated change. Is there something so this shows up in the changelog?

bastelfreak: removed it

PR #1468, opened 2021-09-16
Set additional custom uwsgi_params in nginx::resource::server

danyel2014:


danyel2014:
Hello,

without the $uwsgi_param on nginx::resource::server which is then passed to nginx::resource::location , setting a custom uwsgi param for location / is not possible. This pull request fix this issue.

Dan

PR #1470, opened 2021-09-30
Add support for log_format escape parameter

smortex:

The [escape parameter] was added to nginx 1.11.8.  This optional
parameter allow 3 values: 'default', 'json' and 'none'.  Setting this
parameter is currently tricky do to the way the configuration snippet is
generated (note the unbalanced quotes in the middle of the log string):

```puppet
class { 'nginx':
  # ...
  log_format => {
    json_combined => "escape=json' '{...}",
    #                            ^ ^
    #                            | `-- start quote of the log string
    #                            `---- end quote of the escape format
  }
}
```

Adjust the data type of the log_format parameter to match the "legacy"
way of only passing a Hash consisting of a name (String) matching a
format string (String), but also accept a Tuple for the escape parameter
(Enum) followed by the format string, allowing a less cluttered Puppet
manifest:

```puppet
class { 'nginx':
  # ...
  log_format => {
    json_combined => [
      'escape=json',
      '{...}',
    ],
  },
}
```

[escape parameter]: https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format

n/a

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -116,7 +116,7 @@ class nginx (
   Enum['on', 'off'] $http_tcp_nopush                         = 'off',
   $keepalive_timeout                                         = '65s',
   $keepalive_requests                                        = '100',
-  $log_format                                                = {},
+  Hash[String[1],Nginx::Log_format] $log_format              = {},
   Boolean $mail                                              = false,
   Variant[String, Boolean] $mime_types_path                  = 'mime.types',
   Boolean $stream                                            = false,
--- a/spec/classes/nginx_spec.rb
+++ b/spec/classes/nginx_spec.rb
@@ -562,11 +562,13 @@ describe 'nginx' do
                 attr: 'log_format',
                 value: {
                   'format1' => 'FORMAT1',
-                  'format2' => 'FORMAT2'
+                  'format2' => 'FORMAT2',
+                  'format3' => ['escape=json', '{"auth": "$remote_user"}']
                 },
                 match: [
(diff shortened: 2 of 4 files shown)


smortex:
Maybe it would be better to use a `Variant[String,Struct[...]]` to accommodate possible future parameters?

```puppet
class { 'nginx':
  # ...
  log_format => {
    json_combined => {
      escape => 'json',
      format => '{...}',
    },
  },
}
```

Opinions from people using this module?

ghoneycutt:
This seems a little complex, suggest adding an example to the README.

smortex:
> This seems a little complex, suggest adding an example to the README.

The data type is also lacking documentation.  Maybe we can add examples at this level, I'll try in the next few days when I continue working on this at $WORK.

In the meantime, to you have an opinion regarding a Tuple (`['escape=json', '{...}']`) vs. a Struct (`{ escape => 'json', format => '{...}' }`) for the new way or providing a log format?  I was first thinking about a Tuple because it allows a (little) less [cumbersome] template, but the extra validation when using a Struct would be good to have. Beyond this `log_format` example, I was thinking about [access_log] which is currently implemented using multiple parameters and where passing multiple arguments could be error-prone.  I suspect there are many similar patterns in the module.

[cumbersome]: https://github.com/voxpupuli/puppet-nginx/blob/master/templates/conf.d/nginx.conf.erb#L84-L90
[access_log]: https://nginx.org/en/docs/http/ngx_http_log_module.html#access_log

PR #1471, opened 2021-10-01
add support to access_log, error_log log_not_found per location

ceonizm:
Hello,
This PR adds the support of the following directives at location level
  
- access_log
- error_log
- log_not_found 

it allows to be more precise in the logging strategy. 

Hope it helps.

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -319,9 +319,9 @@ define nginx::resource::location (
   Hash $add_header                                                 = {},
   Optional[Enum['on', 'off', 'always']] $gzip_static               = undef,
   Optional[Enum['on', 'off']] $reset_timedout_connection           = undef,
-  Optional[Variant[Array, String]] $access_log                     = undef,
-  Optional[Variant[Array, String]] $error_log                      = undef,
-  Optional[String] $format_log                                     = $nginx::http_format_log,
+  Optional[Variant[Array[String[1], 1], String[1]]] $access_log    = undef,
+  Optional[Variant[Array[String[1], 1], String[1]]] $error_log     = undef,
+  Optional[String[1]] $format_log                                  = $nginx::http_format_log,
   Optional[Enum['on', 'off']] $log_not_found                       = undef,
 ) {
   if !defined(Class['nginx']) {
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -174,6 +174,21 @@
 # @param reset_timedout_connection
 #   Enables or disables resetting timed out connections and connections closed
 #   with the non-standard code 444.
+# @param format_log
+#   Log_format to use with the defined access_log
+# @param access_log
+#   Where to write access log (log format can be set with $format_log). This
+#   can be either a string or an array; in the latter case, multiple lines will
+#   be created. Additionally, unlike the earlier behavior, setting it to
+#   'absent' in the server context will remove this directive entirely from the
+#   server stanza, rather than setting a default. Can also be disabled for this
+#   server with the string 'off'.
+# @param error_log
+#   Where to write error log. May add additional options like error level to
+#   the end. May set to 'absent', in which case it will be omitted in this
+#   server stanza (and default to nginx.conf setting)
+# @param log_not_found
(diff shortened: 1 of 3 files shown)


ceonizm:
Hello Maintainers :)
 
By looking at theses test results it seems theses kind of errors are certainly not due to my code changes

>   Notice: /Stage[main]/Apt::Update/Exec[apt_update]/returns: E: Failed to fetch https://oss-binaries.phusionpassenger.com/apt/passenger/dists/focal/main/binary-amd64/Packages.gz  File has unexpected size (8692 != 7637). Mirror sync in progress? [IP: 109.107.35.58 443]

 and all others failures are dependent of theses not downloaded packages.
So I guess with all the LetsEncrypt certificates's mess yesterday perhaps it was not the good time to post PR :)

is it possible to re-run the test suite when network issues will be fixed ? 

Have a nice day

manifests/resource/location.pp
```diff
@@ -297,15 +312,19 @@
...
   Boolean $mp4                                                     = false,
   Boolean $flv                                                     = false,
   Optional[String] $expires                                        = undef,
   Hash $add_header                                                 = {},
   Optional[Enum['on', 'off', 'always']] $gzip_static               = undef,
   Optional[Enum['on', 'off']] $reset_timedout_connection           = undef,
+  Optional[Variant[Array, String]] $access_log                     = undef,
+  Optional[Variant[Array, String]] $error_log                      = undef,
+  Optional[String] $format_log                                     = $nginx::http_format_log,
+  Optional[Enum['on', 'off']] $log_not_found                       = undef,
```
smortex, line 325: A lot of data types are currently missing/inconsistent in the module, let's make the new ones perfect :wink: I think all these string must be non-empty, and arrays should only contain non-empty strings?

```suggestion
  Optional[Variant[Array[String[1]], String[1]]] $access_log       = undef,
  Optional[Variant[Array[String[1]], String[1]]] $error_log        = undef,
  Optional[String[1]] $format_log                                  = $nginx::http_format_log,
  Optional[Enum['on', 'off']] $log_not_found                       = undef,
```

smortex (changes requested):
Hey!  I restarted the CI and the transient errors are gone.  The remaining errors on CentOS are handled in #1473 and you can ignore these!

Just a minor suggestion if it looks right to you.

Thanks!

manifests/resource/location.pp
ceonizm: I get the idea and I think it is definitely a good one ;) 
However I just wonder if for syntax consistency sake doing it another PR that would process all data type in the module in one pass wouldn't be a wiser move wouldn't it ?
As you wish :)

smortex: I would suggest you just adjust this for now.  You can adjust the code on your machine, `git commit --amend` to integrate these changes in your commit, and `git push -f` to update the PR.

If beyond these changes you want to help improving data types in the rest of the code, you are very welcome to  do so in another Pull Requests :smiley:.  Fixing everything would be a pain, and is likely to get in the way of other contributors work in progress, so I would suggest multiple successive PRs which focus on some aspects rather than a huge one with tries to fix everything.

ceonizm: done ;)

smortex (approved):
Looks good to me!  I would like to have another review before merging because I am not a "serious" user of this module :smiley:

ghoneycutt:
Adding an acceptance test would make more casual users of this module feel better about pressing the merge button :)

ceonizm:
> Adding an acceptance test would make more casual users of this module feel better about pressing the merge button :)

Hello @ghoneycutt 
Does the tests I've already added aren't sufficent ? 
https://github.com/ceonizm/puppet-nginx/commit/cbd614edd87d5a10a75251cc635919f8460038af#diff-254ab018dcc706b1c8f6aa90e170616c3a13d5fad27a90ed1c830ea4efe83b37

I however agree that there may still a case not tested when access_log is used jointly with log_format because I didn't knew how to write a such test with two params. (for now as you can see I've just mimiced the way existing tests were done by adding my case to the existing array).

I'd love to know what to add :)

PR #1473, opened 2021-10-01
Fix CI on CentOS 7

smortex:
These commits fix CI on CentOS 7 :tada: 

Management of the mail module can be improved in multiple way, but for now I focused on fixing CI so I just ensure that what is not managed by the module and is required is properly setup in the regression test suite.

The configuration of the mail module in CI now match the configuration I have in place for passenger which the module does not handle on CentOS.  We can probably improve this area in another PR to have the module able to have a similar behavior on CentOS and on Debian.


bastelfreak (approved):

ekohl:
This now appears to break things :(

PR #1478, opened 2021-10-27
(1372) Drop run_dir and make client_body_temp_path/proxy_temp_path optional

### LINKED ISSUE #1372: Issue with run files for nginx start with system start

- Puppet: 6.12.0
- Ruby:
- Distribution: Ubuntu 18.04.4
- Module version: 1.1.0

## What are you seeing
The latest upgrade to this module may have broken the nginx service startup process when rebooting server.  After a server reboot the nginx status reads as
`nginx.service - nginx - high performance web server
   Loaded: loaded (/lib/systemd/system/nginx.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2020-02-05 15:17:34 AEDT; 7min ago
     Docs: http://nginx.org/en/docs/
  Process: 664 ExecStart=/usr/sbin/nginx -c /etc/nginx/nginx.conf (code=exited, status=1/FAILURE)`

`Feb 05 15:17:33 jamfp02 systemd[1]: Starting nginx - high performance web server...
Feb 05 15:17:34 jamfp02 nginx[664]: nginx: [emerg] mkdir() "/run/nginx/client_body_temp" failed (2: No such file or directory)
Feb 05 15:17:34 jamfp02 systemd[1]: nginx.service: Control process exited, code=exited status=1
Feb 05 15:17:34 jamfp02 systemd[1]: nginx.service: Failed with result 'exit-code'.
Feb 05 15:17:34 jamfp02 systemd[1]: Failed to start nginx - high performance web server.`

When puppet runs it fixes things and this is the output.
`Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Caching catalog for jamfp02.svi.edu.au
Info: Applying configuration version '1580876853'
Info: /Stage[main]/Ntp::Service/Service[ntp]: Unscheduling refresh on Service[ntp]
Notice: /Stage[main]/Nginx::Config/File[/run/nginx]/ensure: created (corrective)
Notice: /Stage[main]/Nginx::Config/File[/run/nginx/client_body_temp]/ensure: created (corrective)
Notice: /Stage[main]/Nginx::Config/File[/run/nginx/proxy_temp]/ensure: created (corrective)
Info: Class[Nginx::Config]: Scheduling refresh of Class[Nginx::Service]
Info: Class[Nginx::Service]: Scheduling refresh of Service[nginx]
Notice: /Stage[main]/Nginx::Service/Service[nginx]/ensure: ensure changed 'stopped' to 'running' (corrective)
Info: /Stage[main]/Nginx::Service/Service[nginx]: Unscheduling refresh on Service[nginx]`

I expected the nginx process to start at startup before a puppet run corrects things.

b4ldr:
The [general description of `/run/`](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) is fairly vague and not in the original spec however my high level reading is that it relates more to small files which are used for service orchestration and not data (temporary or otherwise) which is specific to the programs operations for the later /var/lib which states the following seems more appropriate:

> Furthermore, this hierarchy holds state information pertaining to an application ....  State information is data that programs modify while they run, and that pertains to one specific host.

https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/Linux-Filesystem-Hierarchy.html#var

And as we see from the [debian packages](https://salsa.debian.org/nginx-team/nginx/-/blob/master/debian/rules#L61-63) this is the route they have gone (and likely don't consider this a bug).  although it is worth noting they only override this only for the full package, the light package keeps the default which ends up being `--prefix` and thus `/usr/share/nginx/client_temp/`.

Of course there is a performance argument to make that says theses files should be in a ramdisk, however its also possible that this dir could become very large and putting it in a ramdisk could cause swapping or other system issues as such i think this decision should be left to the operator with sane defaults provided by the package maintain

As run_dir is only ever used for setting `proxy_temp_path` and `client_body_temp_path` i suggest that we completely drop run_dir and make `client_body_temp_path` and `proxy_temp_path` optional, thus ensuring the configuration uses the distro configured defaults.  if we don't want to use the package defaults then we should ensure we ensure things work correctly annd in this case i think using the systemd dropin would make the most senses

Fixes #1372

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -26,7 +26,6 @@ class nginx::config {
   $pid                            = $nginx::pid
   $proxy_temp_path                = $nginx::proxy_temp_path
   $root_group                     = $nginx::root_group
-  $run_dir                        = $nginx::run_dir
   $sites_available_owner          = $nginx::sites_available_owner
   $sites_available_group          = $nginx::sites_available_group
   $sites_available_mode           = $nginx::sites_available_mode
@@ -189,11 +188,6 @@ class nginx::config {
     }
   }
 
-  file { $run_dir:
-    ensure => directory,
-    mode   => '0644',
-  }
-
   if $nginx::manage_snippets_dir {
     file { $nginx::snippets_dir:
       ensure => directory,
@@ -217,6 +211,28 @@ class nginx::config {
 
   if $proxy_temp_path {
     file { $proxy_temp_path:
+      ensure => directory,
+      owner  => $daemon_user,
+      mode   => '0700',
+    }
+  }
+
+  if $fastcgi_cache_path {
+    file { $fastcgi_cache_path:
+      ensure => directory,
+      owner  => $daemon_user,
+      mode   => '0700',
+    }
+  }
+
+  if $proxy_cache_path =~ Hash {
+    file { $proxy_cache_path.keys():
+      ensure => directory,
+      owner  => $daemon_user,
+      mode   => '0700',
+    }
+  } elsif $proxy_cache_path =~ String {
+    file { $proxy_cache_path:
       ensure => directory,
       owner  => $daemon_user,
       mode   => '0700',
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,7 +40,7 @@
 #
 class nginx (
   ### START Nginx Configuration ###
-  Variant[Stdlib::Absolutepath, Boolean] $client_body_temp_path = $nginx::params::client_body_temp_path,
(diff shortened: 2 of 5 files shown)


kenyon (approved):
Let's fix this thing finally! Thanks for the work @b4ldr.

b4ldr:
Hi all Is there anything outstanding on this one?  didn't want to self merge as im not a user of this module, thx

ashish1099:
I think it LGTM

PR #1480, opened 2021-10-31
modulesync 5.2.0

bastelfreak:
modulesync 5.0.0

### DIFF
--- a/spec/acceptance/nginx_mail_spec.rb
+++ b/spec/acceptance/nginx_mail_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'spec_helper_acceptance'
 
 describe 'nginx::resource::mailhost define:' do
@@ -78,9 +80,10 @@ describe 'nginx::resource::mailhost define:' do
 
       apply_manifest(pp, catch_failures: true)
     end
+
     describe file('/etc/nginx/conf.mail.d/domain1.example.conf') do
       it 'does\'t contain `ssl` on `listen` line' do
-        is_expected.to contain 'listen                *:465;'
+        expect(subject).to contain 'listen                *:465;'
       end
     end
   end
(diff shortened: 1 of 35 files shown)


spec/defines/resource_upstream_spec.rb
```diff
@@ -36,6 +38,7 @@
         ]
       end
 
+      # rubocop:disable RSpec/MultipleMemoizedHelpers
```
root-expert, line 41: Maybe that will make sense to disable it globally?

smortex: I was thinking about it but I have mixed feelings: on one side it is a smell that really should be avoided, but at the same time in acceptance tests it is something ugly for which I can't find a clean solution.

Disabling these checks here is the easy way to move on quickly without blocking everything, not something I expect to see generalized.

spec/classes/nginx_spec.rb
```diff
@@ -36,26 +38,29 @@
         it { is_expected.to contain_nginx__resource__streamhost('streamhost1').with_proxy('streamproxy') }
       end
 
-      context 'nginx::package' do
+      context 'nginx::package' do # rubocop:disable RSpec/EmptyExampleGroup
```
ekohl, line 41: Once https://github.com/rubocop/rubocop-rspec/pull/1249 is released, this should no longer be needed

spec/acceptance/nginx_mail_spec.rb
```diff
@@ -78,9 +83,10 @@ class { 'nginx':
 
       apply_manifest(pp, catch_failures: true)
     end
+
     describe file('/etc/nginx/conf.mail.d/domain1.example.conf') do
       it 'does\'t contain `ssl` on `listen` line' do
-        is_expected.to contain 'listen                *:465;'
+        expect(subject).to contain 'listen                *:465;'
```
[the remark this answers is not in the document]
bastelfreak: we applied this change before we updated voxpupuli-test. I'm currently trying to revert it locally.

PR #1482, opened 2021-11-02
Update mime types with upstream

adriankirchner:
Update default mime types with current upstream at https://github.com/nginx/nginx/blob/3334585539168947650a37d74dd32973ab451d70/conf/mime.types


smortex (approved):

ekohl:
Just a general thought: in RH-based distros there's `/etc/mime.types` provided by `mailcap` and on Debian-based there's `/etc/mime.types` provided by `media-types`. Can't we just use those rather than having this massive data entry?

kenyon:
@ekohl I would rather just leave the file alone and use what’s installed by the distribution package, e.g., in nginx-common on Debian.

adriankirchner:
My guess is that you can't source `/etc/mime.types` to nginx because of different syntax.

> I would rather just leave the file alone and use what’s installed by the distribution package, e.g., in nginx-common on Debian.

That would be fine, but as a user of puppet-nginx I want to be able to add more mime types if necessary

PR #1483, opened 2021-11-09
proxy set header X-Forwarded-Host

nod0n:
Set the X-Forwarded-Host header by default when used as a proxy.

from [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host):
> The X-Forwarded-Host (XFH) header is a de-facto standard header for identifying the original host requested by the client in the Host HTTP request header.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -145,6 +145,7 @@ class nginx (
     'Host $host',
     'X-Real-IP $remote_addr',
     'X-Forwarded-For $proxy_add_x_forwarded_for',
+    'X-Forwarded-Host $host',
     'X-Forwarded-Proto $scheme',
     'Proxy ""',
   ],


root-expert (approved):
Seems reasonable to me

bastelfreak (commented):
I'm still not sure if this is a breaking change / if this adds any security issues for people that miss it. I would like to see a few more comments. @voxpupuli/collaborators and ideas?

vchepkov:
imho, no security implication. Original Host header already contains this information, this additional header helps generate proper response by servers behind reverse proxies.

TuningYourCode:
Well, it's already sending the host as `Host`-header and there are also `X-Real-Ip` and `X-Forwarded-For` which could be critical too.
I guess if somebody cares about the fields as he has an untrusted upstream system he most likely already overwrites our default values.
I also think that almost nobody would proxy any requests to an untrusted upstream system anyway. For me no major release needed.

TuningYourCode (approved):

baurmatt (approved):
Sounds reasonable :)

PR #1484, opened 2021-11-09
Allow specifying error_log severity for servers

### LINKED ISSUE #1416: Cannot specify error_log severity for server definitions without overriding the auto-generated, undocumented error_log directive

- Puppet: 6.x
- Ruby: 2.6.x
- Distribution: FreeBSD
- Module version: 2.0.0

Any nginx server definition

## What are you seeing
error_log has an undocumented, magically constructed default value, and there is no way to add severity flag on a per-server level without losing the automatically named error_log.

1. Documented auto-naming of server-level error_log
2. Ability to specify per-server error_log severity
3. Examples of how to specify error_log achieving the same as what the template does

Note that for 3) above, it is already understood that in explicit server definitions, error_log can be given easily. The problem is that our server definitions are template-based and auto-generated (in the thousands), so there is no way for us to specify the error_log explicitly for each. In the hiera template used for the server definitions we don't have access to the generated title (right?) and nginx does not support using server variables in error_log statements.

ltning:
Fixes #1416

Note - same change may be needed in other template files.

### DIFF
--- a/spec/defines/resource_server_spec.rb
+++ b/spec/defines/resource_server_spec.rb
@@ -345,6 +345,12 @@ describe 'nginx::resource::server' do
               ]
             },
             {
+              title: 'should set error_log severity level',
+              attr: 'nginx_error_log_severity',
+              value: 'warn',
+              match: '  error_log             /path/to/error.log warn;'
+            },
+            {
               title: 'should not include error_log in server when set to absent',
               attr: 'error_log',
               value: 'absent',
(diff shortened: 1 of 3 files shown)


kenyon:
Could you add a test for this too? Seems like there's a similar one already here: https://github.com/voxpupuli/puppet-nginx/blob/master/spec/classes/nginx_spec.rb#L495

ltning:
I *think* I did it right; also added the same fix for non-TLS server entries. Since I'm not terribly familiar with this another set of eyes would be useful. Also uncertain why there's a failed test here..

ltning:
Right, now I think I see why it's failing the tests. My original patch used the top level error_log_severity; it did not introduce this as a setting for servers. Hence the tests are testing something that isn't actually there.
What would you like me to do about this? I can try to add the parameters to the server defines, hoping I get that right..

kenyon:
Yeah I think adding the severity as a class parameter to https://github.com/voxpupuli/puppet-nginx/blob/master/manifests/resource/server.pp is the way to go, similar to the parameters already there which are taking their value from the base class.

Enrice:
Let me direct you to #1594

Enrice:
obsolete now because of merged #1594. I think this can be closed.

PR #1485, opened 2021-11-11
Use rspec stubbed facts

ekohl:
This is to test https://github.com/voxpupuli/voxpupuli-test/pull/62 and see the impact on runtime.


ekohl:
Looks like this can shave off about 10 to 20 minutes off the test suite in this repo. That's nice.

PR #1490, opened 2021-12-03
do not recreate log_dir if it is already a symlink

level-a:

do not recreate log_dir if it is already a symlink

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -201,10 +201,11 @@ class nginx::config {
   }
 
   file { $log_dir:
-    ensure => directory,
-    mode   => $log_mode,
-    owner  => $log_user,
-    group  => $log_group,
+    ensure  => directory,
+    mode    => $log_mode,
+    owner   => $log_user,
+    group   => $log_group,
+    replace => no, # log_dir can be a symlink
   }
 
   if $client_body_temp_path {
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -205,7 +205,7 @@ class nginx::config {
     mode    => $log_mode,
     owner   => $log_user,
     group   => $log_group,
-    replace => no, # log_dir can be a symlink
+    replace => $nginx::manage_log_dir,
   }
 
   if $client_body_temp_path {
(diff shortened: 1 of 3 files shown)


smortex:
Hum… I would precisely want the module to replace the symlink with a directory if it existed :thinking:…

Maybe add a parameter like `Boolean $manage_logdir = true` and conditionally manage this resource?

manifests/config.pp
```diff
@@ -201,10 +201,11 @@
...
   file { $log_dir:
-    ensure => directory,
-    mode   => $log_mode,
-    owner  => $log_user,
-    group  => $log_group,
+    ensure  => directory,
+    mode    => $log_mode,
+    owner   => $log_user,
+    group   => $log_group,
+    replace => $manage_log_dir,
```
smortex, line 208: Should probably be:

```suggestion
    replace => $nginx::manage_log_dir,
```

… that or assign to a local variable like done already too many times above?  :clown_face:  

This module is not the finest in term of elegance :shrug:

smortex (changes requested):

manifests/config.pp
level-a: :man_facepalming:, fixed

smortex:
The unit tests take forever to run but maybe it's worth updating them to check this new parameter in `spec/classes/nginx_spec.rb`?

level-a:
`bundle exec rake parallel_spec` finished for 11m for me
not sure if I know how to properly add it to `nginx_spec.rb`
smth like this to `RedHat` and `Debian` section?
```
context 'manage_log_dir => true' do
  let(:params) { { manage_log_dir: true } }
  it { is_expected.to contain_file('/var/log/nginx') }
end
```

smortex:
@level-a near line 342 / 352 add the `replace` parameter to ensure it has the proper `false` default value, then add a test like above with a `true` value and check that the non-default value is reflected in the file resource parameters.

---

When done, you can combine your commits in a single one (squash):

From your working directory:

```
git rebase -i origin/master # Interactively rewrite history
```

This will bring your editor.  Keep the first line unchanged, and replace `pick` with `squash` for all subsequent lines.  Save and quit, your editor will pop again and allow you to combine the commit messages in a single commit message.

Then push your updated branch to update this PR:

```
git push -f            # Send the changes (-f is required because we re-wrote history)
```

root-expert (commented):
Maybe I'm missing something here, but why do we have to duplicate the spec on both Debian and RedHat?

Can't we add it to a common place? (for all supported OSes)

root-expert (approved):
LGTM

PR #1491, opened 2021-12-04
add server/location_acl argument

level-a:
- add ability to include conf.d/name.acl ACLs to server/location blocks to not repeat allow/deny rules multiple times
- allows to combine common allow list (in .acl file) with allow/deny rules in specific block (deny rule will be the last)

### DIFF
--- a/manifests/resource/location.pp
+++ b/manifests/resource/location.pp
@@ -17,6 +17,8 @@
 #   Locations to allow connections from.
 # @param location_deny
 #   Locations to deny connections from.
+# @param location_acl
+#   Specifies location ACL name in nginx::conf_dir/conf.d/${location_acl}.acl.
 # @param www_root
 #   Specifies the location on disk for files to be read from. Cannot be set in
 #   conjunction with $proxy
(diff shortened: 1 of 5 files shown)


level-a:
ping

level-a:
ping

manifests/resource/location.pp
```diff
@@ -269,6 +271,7 @@
   Optional[Enum['any', 'all']] $location_satisfy                   = undef,
   Optional[Array] $location_allow                                  = undef,
   Optional[Array] $location_deny                                   = undef,
+  Optional[String] $location_acl                                   = undef,
```
root-expert, line 274: I suppose the name can't be an empty string 
```suggestion
  Optional[String[1]] $location_acl                                = undef,
```

Also maybe the user must be able to overwrite the default path `nginx::conf_dir/conf.d` ?

manifests/resource/server.pp
```diff
@@ -290,8 +294,10 @@
   Variant[Array[Stdlib::Absolutepath], Stdlib::Absolutepath] $listen_unix_socket = '/var/run/nginx.sock',
   Optional[String] $listen_unix_socket_options                                   = undef,
   Optional[Enum['any', 'all']] $location_satisfy                                 = undef,
+  Optional[String] $server_acl                                                   = undef,
```
root-expert, line 297: Same here:

```suggestion
  Optional[String[1]] $server_acl                                                = undef,
```

root-expert (changes requested):
Thanks for the PR!
Please add some units tests for this change

PR #1492, opened 2021-12-05
Test Arch Linux acceptance testing

bastelfreak:


smortex:
Do you confirm this is some WIP for future integration of Arch as a first class citizen in our modulesync config?  If so, may I suggest converting this to a draft until final review is ready?

bastelfreak:
the code in puppet-nginx is ready, but we should release the related gems first

PR #1498, opened 2022-02-25
Switch away from os.distro.codename fact

root-expert:
Releated info: https://github.com/voxpupuli/beaker-hostgenerator/pull/244

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -112,7 +112,7 @@ class nginx::params {
       }
     }
     'Debian': {
-      if ($facts['os']['name'] == 'ubuntu' and $facts['os']['distro']['codename'] == 'xenial') {
+      if ($facts['os']['name'] == 'Ubuntu' and $facts['os']['release']['major'] == '16.04') {
         $_module_os_overrides = {
           'manage_repo' => true,
           'daemon_user' => 'www-data',


manifests/params.pp
```diff
@@ -112,7 +112,7 @@
       }
     }
     'Debian': {
-      if ($facts['os']['name'] == 'ubuntu' and $facts['os']['distro']['codename'] == 'xenial') {
+      if ($facts['os']['name'] == 'ubuntu' and $facts['os']['release']['major'] == '16.04') {
```
ekohl, line 115: Slightly surprised this works. I'd expect it to be case sensitive.
```suggestion
      if ($facts['os']['name'] == 'Ubuntu' and $facts['os']['release']['major'] == '16.04') {
```

ekohl (approved):

manifests/params.pp
root-expert: Maybe it doesn't and our CI doesn't catch it :smile:

root-expert:
I guess we have to wait for my PR on puppetlabs-apt to get merged before merging this.

kenyon:
Anything related to EoL Ubuntu 16.04 could also just be removed I think.

manifests/params.pp
smortex: Puppet string comparison is case-insensitive:

https://puppet.com/docs/puppet/7/lang_expressions.html#comparison-equality

ekohl (approved):
Thanks!

ekohl:
> Anything related to EoL Ubuntu 16.04 could also just be removed I think.

In the short term, this improves things without having to do a major release.

PR #1502, opened 2022-03-21
modulesync 5.2.0

bastelfreak:


spec/classes/nginx_spec.rb
```diff
@@ -41,21 +41,24 @@
         when 'RedHat'
           context 'using defaults' do
             it { is_expected.to contain_package('nginx') }
+
             it do
               is_expected.to contain_yumrepo('nginx-release').with(
-                'baseurl'  => "https://nginx.org/packages/#{%w[CentOS VirtuozzoLinux].include?(facts[:operatingsystem]) ? 'centos' : 'rhel'}/#{facts[:operatingsystemmajrelease]}/$basearch/",
```
ekohl, line 46: Isn't there also a style that accepts it consistent? So either all aligned or all ` => `?

spec/defines/resource_upstream_spec.rb
```diff
@@ -242,34 +256,38 @@
...
                     with_content("# MANAGED BY PUPPET\nupstream #{title} {\n")
                 }
+
                 it {
                   is_expected.to contain_concat__fragment("#{title}_upstream_member_#{params[:members].keys[0]}").
                     with_content("  server #{params[:members].keys[0]}:80;\n")
                 }
 
-                if upstream_parameter[:match] != false
+                if upstream_parameter[:match] == false
```
ekohl, line 282: Would it make sense to not swap it but rather write it as
```suggestion
                if upstream_parameter[:match]
```

spec/classes/nginx_spec.rb
smortex: This was added in https://github.com/voxpupuli/voxpupuli-test/pull/77

PR #1507, opened 2022-08-12
Drop Ubuntu 16.04 (EOL)

smortex:
This PR also include:
* #1502


bastelfreak (approved):
Is there maybe some Ubuntu 16 code we can also drop?
at least manifests/params.pp has some code we can drop

bastelfreak (approved):

PR #1512, opened 2022-08-19
Move header escaping responsibility from the end user to Puppet

smortex:

More and more security headers are JSON formatted documents.  With the hardcoded quotes, we have to think hard not to break the generated NGINX configuration.

Rely on Puppet programmatic string representation to avoid the end user to have to worry about escaping quotes.

#### This change is backward incompatible

If you manually escaped header contents before passing them to the module, you need to remove the extra escaping, e.g.:

```diff
diff --git a/site-modules/profile/functions/security_headers.pp b/site-modules/profile/functions/security_headers.pp
index c6087a2..ba6434d 100644
--- a/site-modules/profile/functions/security_headers.pp
+++ b/site-modules/profile/functions/security_headers.pp
@@ -27,9 +27,9 @@ function profile::security_headers(
     x_content_type_options    => 'nosniff',
     referrer_policy           => 'strict-origin-when-cross-origin',
     permissions_policy        => 'accelerometer=(), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), cross-origin-isolated=(), display-capture=(), document-domain=(), encrypted-media=(), execution-while-not-rendered=(), execution-while-out-of-viewport=(), fullscreen=(), geolocation=(), gyroscope=(), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(), xr-spatial-tracking=()',
-    nel                       => '{\"report_to\":\"default\",\"max_age\":31536000,\"include_subdomains\":true}',
-    report_to                 => '{\"group\":\"default\",\"max_age\":31536000,\"endpoints\":[{\"url\":\"https://example.com/\"}],\"include_subdomains\":true}',
-    expect_ct                 => 'report-uri=\"https://example.com/\", max-age=86400',
+    nel                       => '{"report_to":"default","max_age":31536000,"include_subdomains":true}',
+    report_to                 => '{"group":"default","max_age":31536000,"endpoints":[{"url":"https://example.com"}],"include_subdomains":true}',
+    expect_ct                 => 'report-uri="https://example.com/", max-age=86400',
     x_xss_protection          => '0',
   } + $options.lest || { {} }
 

```

n/a

### DIFF
--- a/spec/defines/resource_location_spec.rb
+++ b/spec/defines/resource_location_spec.rb
@@ -460,7 +460,8 @@ describe 'nginx::resource::location' do
                 'add_header' => {
                   'header 1' => 'test value 1',
                   'header 2' => { 'test value 2' => 'tv2' },
-                  'header 3' => { '' => '\'test value 3\' tv3' }
+                  'header 3' => { '' => '\'test value 3\' tv3' },
+                  'header 4' => '{"foo": "bar"}',
                 }
               )
             end
@@ -472,6 +473,8 @@ describe 'nginx::resource::location' do
                 with_content(%r{^\s+add_header\s+"header 2"\s+"test value 2" tv2;$})
               is_expected.to contain_concat__fragment("server1-500-#{Digest::MD5.hexdigest('location')}").
                 with_content(%r{^\s+add_header\s+"header 3"\s+'test value 3' tv3;$})
+              is_expected.to contain_concat__fragment("server1-500-#{Digest::MD5.hexdigest('location')}").
+                with_content(%r(^\s+add_header\s+"header 4"\s+"{\\"foo\\": \\"bar\\"}";$))
             end
           end
         end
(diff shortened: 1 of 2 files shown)


ekohl (commented):
I asked this elsewhere too, but is the StringConverter a public API? Or can we convert the file to EPP and use more native methods?

smortex:
> I asked this elsewhere too, but is the StringConverter a public API? Or can we convert the file to EPP and use more native methods?

I added a similar fix to this PR: switching to EPP would be great but this is a massive change. So for now, just rely on Ruby's `String#inspect`.

ekohl (approved):
Just switching this file feels like a small change, but I didn't look at the details.

smortex:
> Just switching this file feels like a small change, but I didn't look at the details.

The backwards-incompatible is probably over-zealous, but people like me who had to manually escape things will have a broken config after this change, so better safe than sorry :smile:.  I updated the PR description above to describe the needed changes.  Will do the same for the other PR because it also has consequences for some use cases.

PR #1513, opened 2022-08-19
Allow to set log_format's "escape" parameter

smortex:

NGINX can output logs in JSONL format.  This requires to set the escape
parameter to NGINX log_format, which the module does not currently
support.

#### This change is backwards-incompatible

Most users will not be affected, but if you customized the `escape` parameter with a `log_format` containing closing and opening quotes in the form `"escape=foo' 'actual_format"`, an update is required. e.g.:

```diff
diff --git a/site-modules/profile/manifests/nginx.pp b/site-modules/profile/manifests/nginx.pp
index 1fa15f7..3d5192f 100644
--- a/site-modules/profile/manifests/nginx.pp
+++ b/site-modules/profile/manifests/nginx.pp
@@ -109,7 +108,10 @@ class profile::nginx (
     dynamic_modules           => $dynamic_modules,
     log_mode                  => '0755',
     log_format                => {
-      json_combined => "escape=json' '{${json_log.map |$key, $value| { "\"${key}\":${value}" }.join(',')}}",
+      json_combined => {
+        escape => 'json',
+        format => "{${json_log.map |$key, $value| { "\"${key}\":${value}" }.join(',')}}",
+      },
     },
     http_format_log           => 'json_combined buffer=10m flush=5m',
 

```

n/a

### DIFF
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -74,7 +74,11 @@ http {
   default_type  application/octet-stream;
 <% if @log_format -%>
 <% @log_format.sort_by{|k,v| k}.each do |key,value| -%>
-  log_format <%= key %> '<%= value %>';
+  <%- if value.is_a?(Hash) -%>
+  log_format <%= key %> <%= "escape=#{value['escape']} " if value['escape'] %><%= value['format'].inspect %>;
+  <%- else -%>
+  log_format <%= key %> <%= value.inspect %>;
+  <%- end -%>
 <% end -%>
 <% end -%>
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -122,7 +122,7 @@ class nginx (
   Enum['on', 'off'] $http_tcp_nopush                         = 'off',
   $keepalive_timeout                                         = '65s',
   $keepalive_requests                                        = '100',
-  $log_format                                                = {},
+  Hash[String[1], Nginx::LogFormat] $log_format              = {},
   Boolean $mail                                              = false,
   Variant[String, Boolean] $mime_types_path                  = 'mime.types',
   Boolean $stream                                            = false,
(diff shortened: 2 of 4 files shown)


templates/conf.d/nginx.conf.erb
```diff
@@ -74,7 +74,11 @@ http {
   default_type  application/octet-stream;
 <% if @log_format -%>
 <% @log_format.sort_by{|k,v| k}.each do |key,value| -%>
-  log_format <%= key %> '<%= value %>';
+  <%- if value.is_a?(String) -%>
+  log_format <%= key %> <%= Puppet::Pops::Types::StringConverter.convert(value, '%p') %>;
```
ekohl, line 78: What's this string converter? Looks more like it's some private type.

smortex: Good catch, I was at first trying to use EPP but this is a massive change that needs more time than I can afford ATM. `Puppet::Pops::Types::StringConverter` is what implements `String($value, '%p')` in the Puppet language, but since we are on the Ruby side, we can just `String#inspect` I suppose…

smortex: Seems to do the right thing! I amended the commit accordingly.

templates/conf.d/nginx.conf.erb
```diff
@@ -74,7 +74,11 @@ http {
   default_type  application/octet-stream;
 <% if @log_format -%>
 <% @log_format.sort_by{|k,v| k}.each do |key,value| -%>
-  log_format <%= key %> '<%= value %>';
+  <%- if value.is_a?(String) -%>
+  log_format <%= key %> <%= value.inspect %>;
+  <%- else -%>
+  log_format <%= key %> <%= "escape=#{value['escape']} " if value['escape'] %><%= value['format'].inspect %>;
```
ekohl, line 80: Should this be swapped around? Feels like `is_a?(Hash)` would be more robust, though it probably doesn't matter.

ekohl (approved):

templates/conf.d/nginx.conf.erb
smortex: No objection.  Just updated the PR accordingly.

ekohl (approved):

PR #1523, opened 2022-12-01
init: set package_ensure defaults to installed

### LINKED ISSUE #1522: Default package_ensure value is not compatible with current stdlib default for ensure_packages()

- Puppet: puppetserver-7.9.2-1.el7.noarch, puppet-agent-7.20.0-1.el7.x86_64
- Ruby: ruby-2.0.0.648-39.el7_9.x86_64
- Distribution: CentOS Linux release 7.9.2009 (Core)
- Module version: first encountered with puppet/nginx 1.1.0 + puppetlabs/stdlib 8.5.0, 
Also tested with latest versions in git, as of 2022-11-30: https://github.com/voxpupuli/puppet-nginx/tree/09562ecbbba3ea1e9cc27a5a6989363575d8b789 + https://github.com/puppetlabs/puppetlabs-stdlib/tree/e1977c552f8c013f0f3928c246a8e348952dc844

```
include nginx
ensure_packages(['nginx'])
```

Note: while the above snippet alone doesn't make much sense to use, it's a reduction of a much more complex setup, where one type of call is used by one module and the other used by some other module.

## What are you seeing
```
Error: Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: Package[nginx] is already declared at (file: /home/varesa/test-modules/nginx/manifests/package/redhat.pp, line: 106); cannot redeclare (file: /home/varesa/test.pp, line: 2) (file: /home/varesa/test.pp, line: 2, column: 1) on node test-node
```

`ensure_packages()` does not fail. 

This is something that used to work, but the stdlib defaults changed somewhere between 7.1.0 and 8.5.0 from `{ 'ensure' => 'present' }` to `{ 'ensure' => 'installed' }`. This means that the default of `$package_ensure = present` does not deduplicate with `ensure_packages()` and causes a conflicting resource declaration.

While I believe this would also be a breaking change for the nginx-module for the very same reason, maybe in some future major release it'd be a good idea to align with the current stdlib default, causing less friction when used with `ensure_packages()`?

Of course, as of now, there is the workaround of either overriding the ensure-parameter, either in the nginx class declaration, or the ensure_packages call. Assuming that at least one of these is not inside some third party / vendor module.

kenyon:
For compatibility with https://github.com/puppetlabs/puppetlabs-stdlib/pull/1196.

Fixes #1522.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1391,7 +1391,7 @@ Data type: `Any`
 
 
 
-Default value: `present`
+Default value: `installed`
 
 ##### <a name="-nginx--package_name"></a>`package_name`
 
@@ -1455,7 +1455,7 @@ Data type: `Any`
 
 
 
-Default value: `'present'`
+Default value: `installed`
 
 ##### <a name="-nginx--repo_source"></a>`repo_source`
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -193,7 +193,7 @@ class nginx (
   Optional[Enum['on', 'off']] $reset_timedout_connection     = undef,
 
   ### START Package Configuration ###
-  $package_ensure                                            = present,
+  $package_ensure                                            = installed,
   $package_name                                              = $nginx::params::package_name,
   $package_source                                            = 'nginx',
   $package_flavor                                            = undef,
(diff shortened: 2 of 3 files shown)


smortex (approved):
LGTM

PR #1527, opened 2022-12-20
apt::source: configure repo only for current architecture

bastelfreak:
Without this setting, we tell apt to download release information for all architectures apt knows. This includes i386 on amd64 systems. nginx doesn't publish i386 packages. because of that, it results in a warning during apt update:

N: Skipping acquire of configured file 'nginx/binary-i386/Packages' as repository 'https://nginx.org/packages/ubuntu jammy InRelease' doesn't support architecture 'i386'

This is fixed with this change.

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -29,10 +29,11 @@ class nginx::package::debian {
           default => $repo_source,
         }
         apt::source { 'nginx':
-          location => $stable_repo_source,
-          repos    => 'nginx',
-          key      => { 'id' => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62' },
-          release  => $release,
+          location     => $stable_repo_source,
+          repos        => 'nginx',
+          key          => { 'id' => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62' },
+          release      => $release,
+          architecture => $facts['os']['architecture'],
         }
       }
       'nginx-mainline': {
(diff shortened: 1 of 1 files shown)


bastelfreak:
This will look like:
```
Notice: /Stage[main]/Nginx::Package::Debian/Apt::Source[nginx]/Apt::Setting[list-nginx]/File[/etc/apt/sources.list.d/nginx.list]/content:
--- /etc/apt/sources.list.d/nginx.list	2022-12-20 10:44:43.029009469 +0100
+++ /tmp/puppet-file20221220-1147919-1fj17e7	2022-12-20 10:45:04.640920139 +0100
@@ -1,3 +1,3 @@
 # This file is managed by Puppet. DO NOT EDIT.
 # nginx
-deb https://nginx.org/packages/ubuntu jammy nginx
+deb [arch=amd64] https://nginx.org/packages/ubuntu jammy nginx

Notice: /Stage[main]/Nginx::Package::Debian/Apt::Source[nginx]/Apt::Setting[list-nginx]/File[/etc/apt/sources.list.d/nginx.list]/content: current_value '{sha256}827157dcc30574434a3993b56d28624f144469d9e55d7749aaa6aad338ee5749', should be '{sha256}bd83793d6907c903d5701add41c2f046bf70300f8d3c8590523dd6ccc6decef6' (noop)
Notice: Apt::Setting[list-nginx]: Would have triggered 'refresh' from 1 event
Notice: Apt::Source[nginx]: Would have triggered 'refresh' from 1 event
```

arjenz (approved):

smortex:
> Without this setting, we tell apt to download release information for all architectures apt knows.
"apt knows" or "apt is configured to use" :wink:.  I had similar "issue" at a customer running Ubuntu, and preferred to `dpkg --remove-architecture i386` because that made no sense on their system.

They could not say if the added it or if Ubuntu was shipping with this by default.

PR #1531, opened 2023-01-03
Enable subdirectory hierarchy for client_body and proxy temp paths

jplindquist:
Allow specification of up to three-level subdirectory hierarchy for both [client_body_temp_path](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_temp_path) and [proxy_temp_path](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_temp_path) options

This allows for the existing string option to specify just the path as it works now, and for an alternate type as an array with the first element as the path, and the up to 3 additional integers for each level

- Updated docs
- Updated tests

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -206,7 +206,13 @@ class nginx::config {
   }
 
   if $client_body_temp_path {
-    file { $client_body_temp_path:
+    if $client_body_temp_path.is_a(String) {
+      $_client_body_temp_path = [$client_body_temp_path]
+    } else {
+      $_client_body_temp_path = $client_body_temp_path
+    }
+
+    file { $_client_body_temp_path[0]:
       ensure => directory,
       owner  => $daemon_user,
       mode   => '0700',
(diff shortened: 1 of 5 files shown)


jcpunk:
Neat! Looks good to me, but I'll defer to folks with more experience on this module.

manifests/config.pp
```diff
@@ -206,15 +206,29 @@
   }
 
   if $client_body_temp_path {
-    file { $client_body_temp_path:
+    if $client_body_temp_path.is_a(String) {
+      $_client_body_temp_path = [$client_body_temp_path]
+    }
+    else {
```
ekohl, line 212: I'm never a fan of splitting this
```suggestion
    } else {
```

spec/classes/nginx_spec.rb
```diff
@@ -1124,7 +1146,9 @@
 
                   # if we have a _path attribute make sure we create the path
                   if param[:attr].end_with?('_path')
-                    if param[:value].is_a?(Hash)
+                    if (param[:attr] == 'client_body_temp_path' || param[:attr] == 'proxy_temp_path') && param[:value].is_a?(Array)
```
ekohl, line 1149: I'd write this as
```suggestion
                    if ['client_body_temp_path', 'proxy_temp_path'].include?(param[:attr]) && param[:value].is_a?(Array)
```

Slight nit: to keep the diff smaller I'd normally write it as `elsif`. Now you're also changing the `if` itself which makes the diff larger.

ekohl (commented):
Some trivial nits, but overall I think this looks valid. I'm not that familiar with the config option so say if this is useful or usable.

jplindquist:
Thank you @ekohl , updated!

bastelfreak (approved):

ekohl (approved):

jcpunk:
I'd love to see this squashed down to one commit.

PR #1537, opened 2023-02-16
Fix streamhost default logging

jaketus:


kenyon (approved):

bastelfreak:
@jaketus can you please rebase against our latest master?

lkck24:
How can we move on here? Can I help here?

TheMeier (changes requested):
the tests are failing but more important I think this is wrong. It produces something like this
```
access_log off stream_format;
```

PR #1540, opened 2023-03-14
Add RHEL 9 to supported OS

tuxmea:


bastelfreak:
merged this in https://github.com/voxpupuli/puppet-nginx/pull/1543

PR #1542, opened 2023-04-02
Add text/scss mime handling

bastelfreak:


bastelfreak:
mhm it's not listed in https://github.com/nginx/nginx/blob/master/conf/mime.types, maybe we should not add it here :thinking:

PR #1554, opened 2023-06-05
puppetlabs/stdlib: Require 9.x

hashworks:


bastelfreak:
implemented in https://github.com/voxpupuli/puppet-nginx/pull/1558

PR #1555, opened 2023-06-14
Relax dependencies version requirements

smortex:
- Allow puppetlabs-stdlib 9.x
- Allow puppetlabs-concat 9.x


bastelfreak:
implemented in https://github.com/voxpupuli/puppet-nginx/pull/1558

PR #1560, opened 2023-06-28
init.pp: Add missing datatypes

bastelfreak:


manifests/init.pp
```diff
@@ -107,13 +107,13 @@
...
-  $gzip_buffers                                              = undef,
-  $gzip_comp_level                                           = 1,
-  $gzip_disable                                              = 'msie6',
-  $gzip_min_length                                           = 20,
-  $gzip_http_version                                         = 1.1,
+  Optional[String] $gzip_buffers                             = undef,
+  Integer $gzip_comp_level                                   = 1,
+  String $gzip_disable                                       = 'msie6',
+  Integer $gzip_min_length                                   = 20,
+  Float $gzip_http_version                                   = 1.1,
```
alexjfisher, line 114: I know the default is `1.1`, but this doesn't look right to me.  Version 'numbers' should be strings, not floats (and in this case possibly an `Enum`)

manifests/init.pp
```diff
@@ -122,8 +122,8 @@
   Optional[Variant[Array[String], String]] $http_raw_append  = undef,
   Enum['on', 'off'] $http_tcp_nodelay                        = 'on',
   Enum['on', 'off'] $http_tcp_nopush                         = 'off',
-  $keepalive_timeout                                         = '65s',
-  $keepalive_requests                                        = '100',
+  String $keepalive_timeout                                  = '65s',
+  String $keepalive_requests                                 = '100',
```
alexjfisher, line 126: Even the the default is (questionably), a string, did Integers also use to work here?

manifests/init.pp
```diff
@@ -173,8 +173,8 @@
   Enum['on', 'off'] $ssl_stapling_verify                     = 'off',
   Stdlib::Absolutepath $snippets_dir                         = $nginx::params::snippets_dir,
   Boolean $manage_snippets_dir                               = true,
-  $types_hash_bucket_size                                    = '512',
-  $types_hash_max_size                                       = '1024',
+  Variant[Integer,String] $types_hash_bucket_size            = '512',
+  Variant[Integer,String] $types_hash_max_size               = '1024',
```
alexjfisher, line 177: Correct to allow strings, (ie `4k` is an allowable value). I wonder if an `Nginx::Size` type alias would be useful?
http://nginx.org/en/docs/syntax.html

alexjfisher: and I guess also `Nginx::Time_interval`??

manifests/init.pp
alexjfisher: Maybe
```suggestion
  Enum['1.0','1.1'] $gzip_http_version                       = '1.1',
```

alexjfisher: Possibly.  I'm guessing for very few users though.  I can't imagine many people setting this to `1.0` (and as a float).

I suppose you _could_ make it a `Variant[Float,Enum['1.0','1.1']]` and then in the code issue some deprecation warning `unless $gzip_http_version =~ String`, but _that_ is a lot of maintenance work for the benefit that, (I'm _assuming_), will be low in this case.

manifests/init.pp
alexjfisher: The docs have `keepalive_requests` down as a `number`, so we should definitely be allowing `Integer` here.

Not sure if suffixes are allowed on `numbers`.  eg. `'1k'` instead of `1000`.  Can't see this in the docs, but would have to test to be 100% sure.

manifests/init.pp
bastelfreak: I'm fine with anything.  My goal is to enable the parameter types puppet-lint plugin. What about using `Variant[String,Float]` and raise another PR as breaking change that switches to the enum and we keep that open until the next major release? I can follow the same pattern for the other odd types

alexjfisher: Personally, I wouldn't bother with the plugin if it means creating bad data types though.  IMO, the default being a float` here is just a bug.

manifests/init.pp
alexjfisher: @bastelfreak This one definitely needs updating for this PR.  People _will_ have used `Integers` (which is correct for something nginx describes as a `number`).

manifests/init.pp
```diff
@@ -88,12 +88,12 @@
...
   Enum['on', 'off'] $accept_mutex                            = 'on',
-  $accept_mutex_delay                                        = '500ms',
-  $client_body_buffer_size                                   = '128k',
+  String $accept_mutex_delay                                 = '500ms',
+  String $client_body_buffer_size                            = '128k',
   String $client_max_body_size                               = '10m',
-  $client_body_timeout                                       = '60s',
-  $send_timeout                                              = '60s',
-  $lingering_timeout                                         = '5s',
+  String $client_body_timeout                                = '60s',
```
alexjfisher, line 94: I think there's a reasonable chance that people have overridden these timeouts with `Integers` since the `s` for seconds is optional in the config file??

manifests/init.pp
smortex: Looks like `Nginx::Size` and `Nginx::Time` already exist

manifests/init.pp
smortex: Tried to set it to `1k`, it fails with:

```
nginx: [emerg] "keepalive_requests" directive invalid number in /etc/nginx/nginx.conf:36
```

Let's use an Integer!

smortex:
FYI, a companion PR was opened in #1566 to remove the compatibility data types.

manifests/init.pp
alexjfisher: @smortex Great spot! :)

PR #1566, opened 2023-07-05
Remove legacy data types

smortex:
This PR remove backwards-compatible data types which where bogus.

It is intended to be rebased on top of master and merged when we want to cut a new major version and people had the time to update their code to not rely on these.


TheMeier:
rebased

TheMeier:
@smortex  what is missing, or is this ready for review?

smortex:
> @smortex what is missing, or is this ready for review?

I think it is fine.  I think I marked it as draft to avoid merging it too fast. but if we are ready for a new major version this is fine!

PR #1568, opened 2023-07-07
Convert some templates to EPP

jay7x:


ekohl (commented):
I played around with this a bit and I wrote https://gist.github.com/ekohl/28443e6ba178c931bb69285e976568eb. It's not perfect, but it gets you a long way. The biggest thing to be aware of is that you need to change `if x` into `if x {`, but it converts `end` into `}` already as well as `@x` into `$x`. I'd suggest to give it a try and see.

jay7x:
Conversion is mostly completed. Need to recheck whitespace differences though.

smortex (commented):
Looks like you forgot to add `templates/prepend_append.epp` :wink:

kenyon (approved):

smortex (commented):
Impressive!

spec/default_module_facts.yml
```diff
@@ -1,2 +1,4 @@
 ---
 ipaddress6: '::'
```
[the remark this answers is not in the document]
jay7x: This legacy fact is still used in another ERB template.. I'll convert it to EPP later and drop the legacy fact as part of that PR better.

spec/default_module_facts.yml
```diff
@@ -1,2 +1,4 @@
...
+networking:
+  ip6: '::'
```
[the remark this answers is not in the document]
jay7x: Nice suggestion!

kenyon (approved):

smortex (approved):

templates/mailhost/mailhost.epp
```diff
@@ -0,0 +1,49 @@
...
+  <%- } -%>
+<%- } -%>
+<%= $mailhost_common -%>
+
+<%- if versioncmp($nginx_version, '1.15.0') < 0 { -%>
+  ssl                   off;
+<% } %>
+  starttls              <%= $starttls %>;
+
+<% if $starttls == 'on' or $starttls == 'only' { %>
```
ekohl, line 45:
```suggestion
<% if $starttls != 'off' { %>
```

templates/mailhost/mailhost_common.epp
```diff
@@ -0,0 +1,61 @@
...
+<%- } -%>
+<%- } -%>
+<%- if $protocol == 'smtp' { -%>
+<%- if $smtp_auth { -%>
+  smtp_auth          <%= $smtp_auth %>;
+<%- } -%>
+<%- if $smtp_capabilities { -%>
+  smtp_capabilities  <%= $smtp_capabilities.join(" ") %>;
+<%- } -%>
+<%- } -%>
```
ekohl, line 61: Should this be rewritten to a `case $protocol`? Or perhaps even split into `mailhost_$protocol.epp` files.

templates/mailhost/mailhost_ssl.epp
```diff
@@ -0,0 +1,45 @@
...
+# MANAGED BY PUPPET
+server {
+<%= $mailhost_prepend -%>
+<%- if $listen_ip =~ Array { -%>
+  <%- $listen_ip.each |$ip| { -%>
+  listen       <%= $ip %>:<%= $ssl_port %><% if versioncmp($nginx_version, '1.15.0') >= 0 { %> ssl<% } %>;
+  <%- } -%>
+<%- } else { -%>
+  listen                <%= $listen_ip %>:<%= $ssl_port %><% if versioncmp($nginx_version, '1.15.0') >= 0 { %> ssl<% } %>;
+<%- } -%>
```
ekohl, line 24: Alternative to calling `Array()` elsewhere, you can also use
```suggestion
<%- Array($listen_ip).each |$ip| { -%>
  listen       <%= $ip %>:<%= $ssl_port %><% if versioncmp($nginx_version, '1.15.0') >= 0 { %> ssl<% } %>;
<%- } -%>
```

ekohl (commented):
Feel free to ignore my suggestions. They're aimed at making the template even easier.

I notice 1.15.0 is used quite often in these templates which made me check, but sadly Debian 10 still uses nginx 1.14.2.

jay7x:
All suggestions looks reasonable.. I guess I'll implement it this Friday

manifests/resource/mailhost.pp
```diff
@@ -210,17 +266,46 @@
   if $ssl_port == undef or $listen_port != $ssl_port {
     concat::fragment { "${name}-header":
       target  => $config_file,
-      content => template('nginx/mailhost/mailhost.erb'),
       order   => '001',
+      content => epp('nginx/mailhost/mailhost.epp', {
+          has_ipaddress6        => $has_ipaddress6,
+          ipv6_enable           => $ipv6_enable,
+          ipv6_listen_ip        => $ipv6_listen_ip,
```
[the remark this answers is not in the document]
jay7x: If only this would be that easy :)
```
➜  puppet apply -e '$a = "1.2.3.4"; notice(Array($a))'
Notice: Scope(Class[main]): [1, ., 2, ., 3, ., 4]
```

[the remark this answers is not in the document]
jay7x: Aha! I found!
```
➜  puppet apply -e '$a = "1.2.3.4"; $b = ["1.2.3.4", "1.1.1.1"]; notice(Array.new($a, true), Array.new($b, true))'
Notice: Scope(Class[main]): [1.2.3.4] [1.2.3.4, 1.1.1.1]
```

spec/default_module_facts.yml
```diff
@@ -1,2 +1,4 @@
 ---
-ipaddress6: '::'
+ipaddress6: '2001:db8::c0:ffee'
+networking:
+  ip6: '2001:db8::c0:ffee'
```
ekohl, line 4: Nested don't facts work and you'll break things because it doesn't do a deep merge. We have a helper `override_facts`: https://github.com/voxpupuli/voxpupuli-test#fact-handling

I would really like to get rid of this file. If we need custom facts, just call it as native Ruby.

Something like this may work:
```ruby
add_custom_fact :ipaddress6, '2001:db8::c0:ffee'
add_custom_fact: :networking, lambda { |_os, facts| facts[:networking]['ip6'] = facts[:ipaddress6] }
```

But really, I think that if a test cares about IPv6 it should use:
```ruby
context 'with IPv6' do
  let:(facts) { override_facts(super(), ipaddress6: '2001:db8::c0:ffee', networking: { ip6: '2001:db8::c0:ffee' }) }
end
```

ekohl (changes requested):

spec/default_module_facts.yml
jay7x: hmm.. I just realized it works fine without `networking.ip6` override.. I guess facterdb does the magic. So we can just drop it (and then drop the legacy `ipaddress6` in next batch)

ekohl: IMHO a good test doesn't rely on FacterDB to ensure there's an IPv6 address (or is absent) when asserting the IPv6 code works well.

jay7x: IPv6 address is added to the facts explicitly now. There is unit test to ensure behavior is good w/o any IPv6 defined.

manifests/resource/mailhost.pp
```diff
@@ -170,23 +170,38 @@
   Optional[Array] $pop3_capabilities             = undef,
   Optional[String] $smtp_auth                    = undef,
   Optional[Array] $smtp_capabilities             = undef,
-  Optional[Variant[Array, String]] $raw_prepend  = undef,
-  Optional[Variant[Array, String]] $raw_append   = undef,
-  Optional[Hash] $mailhost_cfg_prepend           = undef,
-  Optional[Hash] $mailhost_cfg_append            = undef,
```
bastelfreak, line 176: I'm fine with merging this, but should we mark it as backwards-incompatible because of the type changes?

smortex: Good question, as far as I am concerned, it looks like the previous data types where too relaxed and allowed to provide configuration that would cause nginx to fail to start.  Existing valid config should still be valid so I would not mark it as backwards incompatible, and if we find some corner cases that used to be valid and are now broken, we can quickly issue a new patch release to fix these.

PR #1575, opened 2023-09-29
Use EPP template in nginx::resource::map

jay7x:
Migrate the `nginx::resource::map` defined resource to use an EPP template.


manifests/resource/map.pp
```diff
@@ -65,7 +65,10 @@
 #
 define nginx::resource::map (
   String[2] $string,
-  Variant[Array, Hash] $mappings,
+  Variant[
+    Array[Struct[{ 'key' => String, 'value' => Any }]],
+    Hash[String, Any]
+  ] $mappings = [],
```
bastelfreak, line 71: should we introduce a custom datatype for this ? my personal rule of thumb is: do I need newlines in a datatype? worth migrating it out of the class.

jay7x: I guess yes, it's good time to make another type for this

jay7x: The only concerning question is.. do we need to reuse the type in the EPP also (and move Hash -> Array conversion there maybe)

templates/conf.d/map.epp
```diff
@@ -0,0 +1,31 @@
+<%- |
+  Optional[String] $default,
```
bastelfreak, line 2: I think optional doesn't make sense here unless you give the variable a default value on `undef`?

types/stringmappings.pp
```diff
@@ -0,0 +1,4 @@
+type Nginx::StringMappings = Variant[
```
bastelfreak, line 1: can you add a very brief summary for puppet-strings at the top?

manifests/resource/map.pp
jay7x: Added Nginx::StringMappings and moved Hash -> Array conversion to the EPP

types/stringmappings.pp
jay7x: Ah 🤦🏻‍♂️ sure

templates/conf.d/map.epp
jay7x: The default `undef` comes from the defined resource parameters.. So it is truly optional which is tested later in EPP (See `<% if $default { -%>` below)

jay7x: I can add `undef` here as default value too if it'll be less confusing

bastelfreak (approved):

PR #1581, opened 2023-10-31
Add Debian 12 support

bastelfreak:

PR #1582, opened 2023-10-31
Add OracleLinux support

bastelfreak:
also contains https://github.com/voxpupuli/puppet-nginx/pull/1581

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -55,6 +55,13 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


zilchms:
Seems like there is a new default config for debian12, which we need to address

TheMeier:
superseeded by https://github.com/voxpupuli/puppet-nginx/pull/1601

PR #1594, opened 2024-05-22
Allow specifying error_log severity for servers

Enrice:
improved fix for #1416

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -3322,6 +3322,7 @@ The following parameters are available in the `nginx::resource::server` defined 
 * [`include_files`](#-nginx--resource--server--include_files)
 * [`access_log`](#-nginx--resource--server--access_log)
 * [`error_log`](#-nginx--resource--server--error_log)
+* [`error_log_severity`](#-nginx--resource--server--error_log_severity)
 * [`passenger_cgi_param`](#-nginx--resource--server--passenger_cgi_param)
 * [`passenger_set_header`](#-nginx--resource--server--passenger_set_header)
 * [`passenger_env_var`](#-nginx--resource--server--passenger_env_var)
@@ -4177,9 +4178,16 @@ Default value: `undef`
 
 Data type: `Optional[Variant[String, Array]]`
 
-Where to write error log. May add additional options like error level to
-the end. May set to 'absent', in which case it will be omitted in this
+Where to write error log. May be set to 'absent', in which case it will be omitted in this
 server stanza (and default to nginx.conf setting)
+
+Default value: `undef`
+
+##### <a name="-nginx--resource--server--error_log_severity"></a>`error_log_severity`
+
+Data type: `Optional[Nginx::ErrorLogSeverity]`
+
+Optional error level
 
 Default value: `undef`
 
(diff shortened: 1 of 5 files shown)


TheMeier:
You need to regenerate REFERENCE.md, see https://voxpupuli.org/docs/how_to_run_tests/

Enrice:
Can sb plz hint me why the tests may fail? I must have missed sth.

TheMeier:
The tests for this module are broken, I will merge a fix soon. After I did you need to re-base

TheMeier:
@Enrice if you re-base now the tests should look much better

REFERENCE.md
```diff
@@ -4177,12 +4178,19 @@ Default value: `undef`
 
 Data type: `Optional[Variant[String, Array]]`
```
TheMeier, line 4179: Shouldn't this be switched to `Optional[String]`  now? I don't see any other options except the log level here https://nginx.org/en/docs/ngx_core_module.html#error_log

Enrice:
I was referring to the unit tests. Still no clue why they fail.

TheMeier:
You did not set the new option in the tests. I guess it would go here: https://github.com/voxpupuli/puppet-nginx/pull/1594/files#diff-26cc1a5b135079611a0a648adb7f94475b3eae5ca87348edb6933631ce8e79aeL72

But actually A dedicated test for the new parameter would be better

REFERENCE.md
Enrice: As far as I understood, this is to be able to specify multiple lines of error_log. Look at the templates.

Enrice:
That's what I did here e.g.:
https://github.com/voxpupuli/puppet-nginx/pull/1594/files#diff-26cc1a5b135079611a0a648adb7f94475b3eae5ca87348edb6933631ce8e79aeR361

Still the defaults from the define or the class don't seem to apply.

TheMeier:
Th part in line 361-372 look good. Simple revert the changes in lines 90, 350, 357-358. These are different tests...

Enrice:
I don't understand why this should work, because I changed the default here: https://github.com/voxpupuli/puppet-nginx/pull/1594/files#diff-8967bbea230d7e3438686906bca67bccaa8b7f25820b28faf3a1d29494812493R403

So this change should also be reflected in the tests.

However we can discuss if this breaking change makes sense. To me, it does.
Reason: we EXPLICITLY default the severity on class level to 'error' which would be the nginx default anyway. So it is legit to have it EXPLICITLY configured on the server resource as well, so I followed this pattern.

But maybe this should be split into separate features?

REFERENCE.md
TheMeier: Right I did overlook that. The template is kind of ugly, I think it would benefit from a refactoring some time. Maybe it would make sense to change `error_log` to a more specific type to aid in that. Something like this:

```
Variant[Undef, 'absent', Array[Hash[Stdlib::Absolutepath,Optional[Nginx::ErrorLogSeverity]]] = [
  {   '/foo/bar' => 'undef' },
  {   '/a/b/'       => 'warn' },
}
```
Just some random idea ;)

bastelfreak:
I think this is fine to merge as an enhancement because it doesn't change the current log level? I also had to rebase it to cleanup the git history.

bastelfreak (approved):

PR #1595, opened 2024-05-26
drop support for Debian 10

TheMeier:
there is already breaking changes on master, so next releease will be major

test have been broken for a while and will be fixed in https://github.com/voxpupuli/puppet-nginx/pull/1596

### DIFF
--- a/spec/acceptance/nginx_server_spec.rb
+++ b/spec/acceptance/nginx_server_spec.rb
@@ -11,7 +11,10 @@ describe 'nginx::resource::server define:' do
         ensure   => present,
         www_root => '/var/www/www.puppetlabs.com',
       }
-      host { 'www.puppetlabs.com': ip => '127.0.0.1', }
+      file_line {'mock hosts':
+        path => '/etc/hosts',
+        line => '127.0.0.1 www.puppetlabs.com'
+      }
       file { ['/var/www','/var/www/www.puppetlabs.com']: ensure => directory }
       file { '/var/www/www.puppetlabs.com/index.html': ensure  => file, content => 'Hello from www\n', }
       "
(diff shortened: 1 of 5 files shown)


spec/classes/nginx_spec.rb
```diff
@@ -190,7 +190,7 @@
 
             it { is_expected.to contain_package('nginx') }
 
-            if (facts.dig(:os, 'name') == 'Debian' && %w[10 11].include?(facts.dig(:os, 'release', 'major'))) ||
+            if (facts.dig(:os, 'name') == 'Debian' && %w[11].include?(facts.dig(:os, 'release', 'major'))) ||
                (facts.dig(:os, 'name') == 'Ubuntu' && %w[bionic focal jammy].include?(facts.dig(:os, 'distro', 'codename')))
```
ekohl, line 194: These all look like modern versions. Is there anything that still uses plain `passenger` or  can it be simplified to just `facts.dig(:os, 'family') == 'Debian'`?

ekohl (changes requested):
This includes https://github.com/voxpupuli/puppet-nginx/pull/1596, but IMHO it's the other way around: this is a prerequisite for dropping older nginx support.

kenyon (approved):

spec/acceptance/nginx_server_spec.rb
```diff
@@ -11,7 +11,10 @@ class { 'nginx': }
         ensure   => present,
         www_root => '/var/www/www.puppetlabs.com',
       }
-      host { 'www.puppetlabs.com': ip => '127.0.0.1', }
+      file_line {'mock hosts':
+        path => '/etc/hosts',
+        line => '127.0.0.1 www.puppetlabs.com'
+      }
```
smortex, line 17: Looks like `host_core` is not in fixtures.yml.

While this is indeed an extra dependency for CI, I like the idea of acceptance tests that follow best practices by using relevant modules: if someone wants to run nginx on windows (oh my), they will probably have to revert this.  But this is definitively not a blocker.

smortex (approved):

spec/acceptance/nginx_server_spec.rb
TheMeier: I actually did implement that with `host_core` and the removed it again because I could not find any way to add a fixture but **only** for acceptance tests.
I will open another issue for it

PR #1596, opened 2024-05-26
drop support for versions < 1.15.0

TheMeier:


manifests/init.pp
```diff
@@ -240,7 +240,7 @@
   Hash $nginx_upstreams                                   = {},
   Nginx::UpstreamDefaults $nginx_upstreams_defaults       = {},
   Boolean $purge_passenger_repo                           = true,
-  String[1] $nginx_version                                = pick(fact('nginx_version'), '1.6.0'),
+  String[1] $nginx_version                                = pick(fact('nginx_version'), '1.16.0'),
```
smortex, line 243: I am wondering if we can not just default to installed?

```suggestion
  String[1] $nginx_version                                = 'installed',
```

Putting an arbitrary (outdated) version does not look good, and if we do a breaking release maybe it is time to improve this?

TheMeier: Fine with me, I was just trying to change as little as possible

ekohl (changes requested):
Debian 10 ships nginx 1.14, so this would require dropping that from `metadata.json`.

EPEL7 contains nginx 1.20, so that one shouldn't cause any issues.

PR #1607, opened 2024-07-15
Fix use of legacy facts

silug:
Replace legacy facts to improve compatibility with Puppet 8.

N/A

### DIFF
--- a/lib/facter/nginx_version.rb
+++ b/lib/facter/nginx_version.rb
@@ -2,7 +2,7 @@
 
 Facter.add(:nginx_version) do
   confine { Facter.value(:kernel) != 'windows' }
-  confine { Facter.value(:operatingsystem) != 'nexus' }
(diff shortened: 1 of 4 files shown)


silug:
This fixes tests for #1606 for me.

lib/facter/nginx_version.rb
```diff
@@ -2,7 +2,7 @@
 
 Facter.add(:nginx_version) do
   confine { Facter.value(:kernel) != 'windows' }
-  confine { Facter.value(:operatingsystem) != 'nexus' }
+  confine { Facter.value(:os)['name'] != 'nexus' }
```
ekohl, line 5: I think you can use
```suggestion
  confine { Facter.value('os'.name') != 'nexus' }
```

While it's probably safe to assume `os` is present and there's probably no (measurable) performance difference, but , it's still a good practice to learn.

PR #1608, opened 2024-07-29
Fix http_raw_prepend/append indentation

matejzero:

`http_raw_prepend` and `http_raw_append` have wrong indentation. Example with current template:
```
  ssl_stapling_verify       off;

    real_ip_header X-Forwarded-For;
    set_real_ip_from 1.2.3.4;

  include /etc/nginx/conf.d/*.conf;
```

This MR removes 2 extra spaces.


kenyon (approved):
Since this would cause a restart of the nginx service, IMO we should wait to merge this until we do a major release.

matejzero:
I can live with that, I think it's only a cosmetic fix.

PR #1612, opened 2024-09-01
feat: support disabling "worker_rlimit_nofile"

### LINKED ISSUE #1611: Support disabling "worker_rlimit_nofile"

- Puppet: 8.8.1
- Ruby: 3.1.2p20
- Distribution: Debian GNU/Linux 12 (bookworm)
- Module version: 6.0.0

```puppet
  class { 'nginx':
    worker_rlimit_nofile => undef,
  }
```

## What are you seeing

row `worker_rlimit_nofile 1024;` inside `/etc/nginx/nginx.conf` file

Config without `worker_rlimit_nofile`

## Output log

None? Maybe I missed something

I don't want to use "worker_rlimit_nofile" inside the config and try disable it from rendering from template.
Also I try to use false or empty string and get errors when validating.

As a potential solution:
Add `> 0` inside template

```erb
<% if @worker_rlimit_nofile * > 0 * -%>
worker_rlimit_nofile <%= @worker_rlimit_nofile %>;
<% end -%>
```

DoctorFTB:

Add support for disabling "worker_rlimit_nofile" by checking for a value greater than zero

Fixes #1611

### DIFF
--- a/spec/classes/nginx_spec.rb
+++ b/spec/classes/nginx_spec.rb
@@ -448,6 +448,12 @@ describe 'nginx' do
                 match: 'worker_rlimit_nofile 10000;'
               },
               {
+                title: 'should not set worker_rlimit_nofile',
+                attr: 'worker_rlimit_nofile',
+                value: 0,
+                notmatch: %r{worker_rlimit_nofile}
+              },
+              {
                 title: 'should set pcre_jit',
                 attr: 'pcre_jit',
                 value: 'on',
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -13,7 +13,7 @@ daemon <%= @daemon %>;
 user <%= @daemon_user %><% if @daemon_group -%> <%= @daemon_group %><% end -%>;
 <% end -%>
 worker_processes <%= @worker_processes %>;
-<% if @worker_rlimit_nofile -%>
+<% if @worker_rlimit_nofile > 0 -%>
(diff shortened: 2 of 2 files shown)


kenyon (commented):
This seems OK, but I think it would be better if it defaulted to `undef`, so that you don't have to specify a special value to disable it. I guess that would be a breaking change though. Not sure why the default is 1024 when there is no default upstream: https://nginx.org/en/docs/ngx_core_module.html#worker_rlimit_nofile

DoctorFTB:
> This seems OK, but I think it would be better if it defaulted to `undef`, so that you don't have to specify a special value to disable it. I guess that would be a breaking change though. Not sure why the default is 1024 when there is no default upstream: https://nginx.org/en/docs/ngx_core_module.html#worker_rlimit_nofile

Yeah! This module has not a little options with values ​​different from the default values ​​in the nginx docs.

For example:
*name* -> *current module* vs *nginx docs*
`keepalive_timeout` -> `65s` vs `75s`
`keepalive_requests` -> `100` vs `1000`
`client_max_body_size` -> `10m` vs `1m`

Also has a default value as value, not as undef, like options: `client_body_timeout`, `send_timeout`

May be its `best practice` or .. ?

deric:
@DoctorFTB Default value for `keepalive_requests`  changed in Nginx `1.19.10`. In this case such change should be backward compatible. But generally it's hard to maintain different default values for different package version.

https://nginx.org/en/docs/http/ngx_http_core_module.html
> Prior to version 1.19.10, the default value was 100.

PR #1626, opened 2024-11-18
fix handling for arm64 vs aarch64 on debian/ubuntu apt repositories

### LINKED ISSUE #1625: Wrong plattform type for ubuntu 22.04 on arm when using nginx repository

- Puppet: 7
- Ruby: -
- Distribution: Ubuntu 22.04 LTS
- Module version: validated for > 5.0

The fact [os][architecture] for Ubuntu 22.04 running on an ARM machine ist aarch64.
The nginx repository expects arm64 in this case. (See: https://nginx.org/packages/ubuntu/dists/jammy/nginx/) 

There should be a logic in the package::debian.pp to handle this correctly and set arm64 in the nginx.list apt repository configuration instead of aarm64.

bk-lg:
Add logic for using arm64 instead of aarch64 on ubuntu/debian systems.
The system fact reported by puppet is aarch64 but the repositories expecting arm64. 

Fixes #1625

### DIFF
--- a/manifests/package/debian.pp
+++ b/manifests/package/debian.pp
@@ -28,12 +28,23 @@ class nginx::package::debian {
           undef => "https://nginx.org/packages/${distro}",
           default => $repo_source,
         }
-        apt::source { 'nginx':
-          location     => $stable_repo_source,
-          repos        => 'nginx',
-          key          => { 'id' => '573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62' },
-          release      => $release,
-          architecture => $facts['os']['architecture'],
+        if $facts['os']['architecture'] == 'aarch64' {
+          apt::source { 'nginx':
+            location     => $stable_repo_source,
+            repos        => 'nginx',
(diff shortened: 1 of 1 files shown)


bk-lg:
@kenyon thank you for your fast response. Is this correctly done like you wanted?

bk-lg:
For notice: Debian 12 check fails due to a timeout of the ci worker against the regular debian archives

kenyon (approved):
Yes, that's what I was thinking.

bk-lg:
@kenyon do I need to do anything else for getting this merged or is my worke done here and it will be merged before the next module release or something like that?

kenyon:
What version of puppetlabs-apt are you using? Seems like this should have been fixed in puppetlabs-apt v8.4.0 by https://github.com/puppetlabs/puppetlabs-apt/pull/1012.

bk-lg:
I am using puppetlabs-apt in version 8.5.0 in the affected environment. But then this seems to be out of scope for the puppet-nginx module and I need to start further investigations and eventualy will file an issue in the puppetlabs-apt project if I find no cause on my side. 
Thank you for pointing out this older issue in the apt module.

PR #1632, opened 2025-04-15
Add ssl parameter to IPv6 listen directive

Cicco0:

This change ensures that the `ssl` parameter is also applied to the IPv6 listen directive ([::]:993) when the Nginx version is ≥ 1.15.0. Previously, only the IPv4 directive (*:993 ssl) included the ssl parameter.


TheMeier:
@Cicco0 can you please add a test for that condition and rebase.

Cicco0:
@TheMeier @smortex 
I just wanted to follow up on my pull request. Please let me know if there's anything I can do or update to help move it forward.

PR #1634, opened 2025-04-28
Disable mail relay in nginx 1.14

jstraw:
While redhat installs the mail packages by default, arch doesn't and therefore you need to be able to define that package for installation.

### DIFF
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -194,6 +194,17 @@ define nginx::resource::mailhost (
     package { $nginx::mail_package_name:
       ensure => 'installed',
     }
+    $mail_load_content = $facts['os']['family'] ? {
+      'ArchLinux' => "load_module /usr/lib/nginx/modules/ngx_mail_module.so;\n",
+      'RedHat'    => "load_module /usr/lib64/nginx/modules/ngx_mail_module.so;\n",
+    }
+    file { '/etc/nginx/modules-enabled/mail.conf':
+      ensure  => 'file',
+      owner   => 'root',
+      mode    => '0644',
+      content => $mail_load_content,
+      require => File['/etc/nginx/modules-enabled'],
+    }
   }
 
   # Add IPv6 Logic Check - Nginx service will not start if ipv6 is enabled
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -189,7 +189,10 @@ define nginx::resource::mailhost (
     fail('You must include the nginx base class before using any defined resources')
   } elsif versioncmp($facts.get('nginx_version', $nginx::nginx_version), '1.15.0') < 0 {
     fail('This module does not support nginx 1.14')
+  } elsif ! $nginx::mail {
+    fail('nginx mail proxy requires the nginx::mail flag to be set true')
   }
+
   if $nginx::mail_package_name {
     package { $nginx::mail_package_name:
       ensure => 'installed',
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -188,6 +188,11 @@ define nginx::resource::mailhost (
   if ! defined(Class['nginx']) {
     fail('You must include the nginx base class before using any defined resources')
   }
+  if $nginx::mail_package_name {
+    package { $nginx::mail_package_name:
+      ensure => 'installed',
+    }
+  }
 
   # Add IPv6 Logic Check - Nginx service will not start if ipv6 is enabled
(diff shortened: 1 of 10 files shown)


manifests/resource/mailhost.pp
```diff
@@ -188,6 +188,11 @@
   if ! defined(Class['nginx']) {
     fail('You must include the nginx base class before using any defined resources')
   }
+  if $nginx::mail_package_name {
+    package { $nginx::mail_package_name:
+      ensure => present,
```
bastelfreak, line 193:
```suggestion
      ensure => 'installed',
```

bastelfreak (approved):

manifests/resource/mailhost.pp
```diff
@@ -187,6 +187,24 @@
...
     fail('You must include the nginx base class before using any defined resources')
+  } elsif versioncmp($facts.get('nginx_version', $nginx::nginx_version), '1.15.0') < 0 {
+    fail('This module does not support nginx 1.14')
+  }
+  if $nginx::mail_package_name {
+    package { $nginx::mail_package_name:
+      ensure => 'installed',
+    }
+    $mail_load_content = $facts['os']['family'] ? {
+      'ArchLinux' => 'load_module /usr/lib/nginx/modules/ngx_mail_module.so;',
```
bastelfreak, line 198:
```suggestion
      'ArchLinux' => "load_module /usr/lib/nginx/modules/ngx_mail_module.so;\n",
```

manifests/resource/mailhost.pp
```diff
@@ -187,6 +187,24 @@
...
+      'RedHat'    => 'load_module /usr/lib64/nginx/modules/ngx_mail_module.so;',
```
bastelfreak, line 199:
```suggestion
      'RedHat'    => "load_module /usr/lib64/nginx/modules/ngx_mail_module.so;\n",
```

manifests/resource/mailhost.pp
jstraw: good catch, I'll get these when I squash a bunch of the extra commits

manifests/resource/mailhost.pp
```diff
@@ -187,6 +187,24 @@
 ) {
   if ! defined(Class['nginx']) {
     fail('You must include the nginx base class before using any defined resources')
+  } elsif versioncmp($facts.get('nginx_version', $nginx::nginx_version), '1.15.0') < 0 {
+    fail('This module does not support nginx 1.14')
```
ekohl, line 191:
```suggestion
    fail('The mail module requires nginx 1.15 or newer')
```

spec/acceptance/nginx_mail_spec.rb
```diff
@@ -5,15 +5,9 @@
 describe 'nginx::resource::mailhost define:' do
   has_recent_mail_module = true
 
-  if fact('os.family') == 'RedHat' && fact('os.release.major') == '8'
-    # EPEL had recent nginx-mod-mail package for CentOS 7 but not CentOS 8
-    # Stream.  The base packages use an older version of nginx that does not
-    # work with the acceptance test configuration.
-    has_recent_mail_module = false
-  end
+  has_recent_mail_module = false if fact('os.family') == 'RedHat' && fact('os.release.major') == '8'
```
ekohl, line 8: You can trim it down to a single statement
```suggestion
  has_recent_mail_module = fact('os.family') != 'RedHat' || fact('os.release.major') != '8'
```

bastelfreak:
Do you know which operating systems still ship nginx 1.14? And to me this looks like a bugfix and not a breaking change? Is there any changed behaviour that justifies the backwards-incompatible label that I missed?

jstraw:
El8 uses 1.14.1 in epel.  

Fir the backwards incompatible flag, it's if you were actually using it on arch or el where there's a module for mail, we now load the module ourselves where they would have had to do it before... that could cause an unexpected failure of their code where they are loading the module themselves.

templates/conf.d/nginx.conf.erb
```diff
@@ -6,6 +6,7 @@ load_module "<%= mod_item -%>";
 load_module "modules/<%= mod_item -%>.so";
   <%- end -%>
 <%- end -%>
+
```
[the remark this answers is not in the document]
kenyon: When this has happened before, and we attempted to restore the whitespace, people just get mad again because of yet another whitespace change, so probably best to just leave it now ([example](https://github.com/puppetlabs/puppetlabs-docker/pull/983)). It was already a major version, so a service restart shouldn't be too unexpected.

PR #1639, opened 2025-06-11
Update README and metadata

nmburgan:
Update the readme and metadata to show we can no longer support PE or Puppet Core

n/a


genebean (approved):
LGTM. FWIW, there is an archive of that page in case they pull it or change it:
https://web.archive.org/web/20250416004958/https://www.puppet.com/blog/open-source-puppet-updates-2025

bastelfreak (changes requested):
We also need a PR title that reflects this change. And we need to discuss if this is a major change or not. We usually did major releases for dropping a end of life puppet version.. And it would help if we would actually test against our facter version.

nmburgan:
Good call. Yeah, we should figure out if we need to major rev it, and hopefully we can get a facter-new gem out soon.

bastelfreak:
Next step for facter: finding a new name: https://github.com/OpenVoxProject/facter/issues/12

Afterwards we can release it. That shouldn't take too long.

metadata.json
```diff
@@ -18,10 +18,6 @@
     }
   ],
   "requirements": [
-    {
```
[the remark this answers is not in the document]
nmburgan: Note to reviewers: I will remove this before merging for now, until we've got the Facter fork ready and we're comfortable removing it.

jhoblitt (approved):
The warning language is well phrased and appropriate.

tuxmea:
Do we really want to annoy users by specifying that a module is no longer "supported" to work on PE/PC?
I would prefer to say that we "**can no longer verify functionality on PE/PC**" and that additional testing is required.

It sound less harsh and might motivate people to move from PE/PC to OpenVox.
If we just say "no support", people might not doing updates anymore or build something by themselves.

binford2k:
I suppose we could even add a link.

> Before using this module with Puppet Core, we suggest [testing it yourself](https://www.puppet.com/docs/pdk/latest.x/pdk_testing.html).

tuxmea:
please use correct link: https://www.puppet.com/docs/pdk/latest/pdk_testing.html

binford2k:
that's weird. It must have been an artifact of their docs system changes, because I copied that URL from a browser page.

README.md
```diff
@@ -9,6 +9,12 @@
 
 This module was migrated from James Fryman <james@frymanet.com> to Vox Pupuli.
 
+## WARNING: This module is no longer supported for Puppet Enterprise or Puppet Core
```
binford2k, line 12:
```suggestion
## WARNING: This module is no longer tested on Puppet Enterprise or Puppet Core
```

I think that I've come around to @tuxmea's suggestion. "Supported" is such an overloaded word, in some cases we use it to mean "this works on," in some cases we use it to mean "we've tested it on," in some cases we use it to mean "there's commercial support for," and so on.

Let's just be specific and say the accurate thing we're getting at.

binford2k (changes requested):

README.md
nmburgan: Oops, I intended to but neglected to change the header :)

nmburgan: Fixed

binford2k:
So here's a fun fact for y'all. If you remove `puppet` from `metadata.json`, the PDK shits its pants.

```
$ pdk validate
pdk (ERROR): Module metadata does not contain a "puppet" requirement.
$ pdk test unit
pdk (ERROR): Module metadata does not contain a "puppet" requirement.
```

And if you dare try `pdk build` then it tries to completely regenerate the metadata. I propose we add proper bounds instead. I'm ok with it in this PR or in a separate PR.

```json
    {
      "name": "puppet",
      "version_requirement": ">= 7.0.0 <= 8.10.0"
    },
```

jhoblitt:
Could we start using the string `openvox` instead of `puppet`, since we will no longer be testing with "puppet"?

binford2k:
@jhoblitt for the modules I'm working on right now, I added both. I updated my comment to represent that.

PR #1645, opened 2025-06-19
Allow `ssl_verify_client` when only `ssl_trusted_cert` is set

### LINKED ISSUE #1644: Support `ssl_verify_client` when only `ssl_trusted_cert` is set
Currently, `ssl_client_cert` must be set for `ssl_verify_client` to take effect. As of TLS 1.1, and nginx 1.28 (possibly older), this is no longer required.

This is a very simple proposed patch to allow `ssl_verify_client` when either `ssl_client_cert` or `ssl_trusted_cert` is set, as both cases are valid configurations:

```diff
--- a/modules/nginx/templates/server/server_ssl_settings.erb
+++ b/modules/nginx/templates/server/server_ssl_settings.erb
@@ -16,6 +16,8 @@
 <% end -%>
 <% if defined? @ssl_client_cert -%>
   ssl_client_certificate    <%= @ssl_client_cert %>;
+<% end -%>
+<% if defined? ssl_verify_client && ( defined? @ssl_client_cert || defined? @ssl_trusted_cert ) -%>
   ssl_verify_client         <%= @ssl_verify_client %>;
 <% end -%>
 <% if defined? @ssl_dhparam -%>
```

ltning:
Fixes #1644

### DIFF
--- a/templates/server/server_ssl_settings.erb
+++ b/templates/server/server_ssl_settings.erb
@@ -14,11 +14,18 @@
   ssl_certificate_key       <%= key %>;
 <% end -%>
 <% end -%>
-<% if defined? @ssl_client_cert -%>
+<% if scope.call_function('versioncmp', [scope['nginx::nginx_version'], '1.27.2']) >= 0 -%>
+<%   if defined? @ssl_client_cert -%>
   ssl_client_certificate    <%= @ssl_client_cert %>;
-<% end -%>
-<% if defined? ssl_verify_client && ( defined? @ssl_client_cert || defined? @ssl_trusted_cert ) -%>
+<%   end -%>
+<%   if ( defined? @ssl_verify_client ) && ( @ssl_client_cert.is_a?(String) || @ssl_trusted_cert.is_a?(String) ) -%>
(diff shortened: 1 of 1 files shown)


ltning (commented):
Updated with version check and corrected logic

kenyon (approved):

ltning:
I have found an issue with this PR, and I'm unsure how to solve it. In short, the global default for `ssl_verify_client` is `on`, which means that it will now be switched on for anyone who uses `ssl_trusted_certificates` for OCSP purposes, for example.

My suggestion: Change the global default to `unset`. This will require people to be explicit if they want `ssl_verify_client`. In my opinion this is the correct solution.
We could modify the template to set it explicitly to 'on' if `ssl_client_certificate` is set AND `ssl_verify_client` is unset. This would keep the current behaviour for what I think is the majority of cases.

Alternatively, change the global default to `no`. This *will* require changes for everyone who relies on the default, which is probably bad.

I'll work on and propose a PR with my suggested change, but I don't have the infrastructure to test it thoroughly. I'll need help assessing whether this might have broader implications than I am bargaining for.

PR #1647, opened 2025-06-23
Fix ssl verify client compat

### LINKED ISSUE #1646: Fix for #1644 (`ssl_verify_client` with only `ssl_trusted_certificate`) breaks common use case
Also posted as a comment on #1645:

I have found an issue with this PR (yes, it is mine..), and I'm unsure how to solve it. In short, the global default for `ssl_verify_client` is `on`, which means that it will now be switched on for anyone who uses `ssl_trusted_certificates` for OCSP purposes, for example.

My suggestion: Change the global default to `unset`. This will require people to be explicit if they want `ssl_verify_client`. In my opinion this is the correct solution.
We could modify the template to set it explicitly to 'on' if `ssl_client_certificate` is set AND `ssl_verify_client` is unset. This would keep the current behaviour for what I think is the majority of cases.

Alternatively, change the global default to `no`. This *will* require changes for everyone who relies on the default, which is probably bad.

I'll work on and propose a PR with my suggested change, but I don't have the infrastructure to test it thoroughly. I'll need help assessing whether this might have broader implications than I am bargaining for.

ltning:
Fixes a problem introduced with #1645. The global default for `ssl_verify_client` is `on`, which means that it will be switched on for anyone who uses `ssl_trusted_certificates` for OCSP purposes, for example.

This commit changes the global default to `unset`. This will require people to be explicit if they want `ssl_verify_client`.

The `server_ssl_settings.erb` template now sets `ssl_verify_client` explicitly to 'on' if `ssl_client_certificate` is set AND `ssl_verify_client` is unset. This keeps the current behaviour for what I think is the majority of cases.

Tested to the best of my ability, and I could not find any other cases that expect a value to be present for `ssl_verify_client`.

Fixes #1646

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -3747,11 +3747,11 @@ Default value: `undef`
 
 ##### <a name="-nginx--resource--server--ssl_verify_client"></a>`ssl_verify_client`
 
-Data type: `String`
+Data type: `Optional[String]`
 
 Enables verification of client certificates.
 
-Default value: `'on'`
+Default value: `undef`
 
 ##### <a name="-nginx--resource--server--ssl_crl"></a>`ssl_crl`
 
--- a/manifests/resource/server.pp
+++ b/manifests/resource/server.pp
@@ -305,7 +305,7 @@ define nginx::resource::server (
   Boolean $ssl_listen_option                                                     = true,
   Optional[Variant[String, Boolean, Array[String]]] $ssl_cert                    = undef,
   Optional[String] $ssl_client_cert                                              = undef,
-  String $ssl_verify_client                                                      = 'on',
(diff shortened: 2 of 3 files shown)


kenyon (changes requested):
Can you please rebase and resolve the merge conflicts? Should be good then 👍

kenyon:
Can you please rebase to remove the merge commit? Thanks.

ltning:
> Can you please rebase to remove the merge commit? Thanks.

I am afraid my git-fu is failing me. :( I'm not sure how to go about doing that..

kenyon:
Also your current commits are not based on the current `master`, so make sure you do `git fetch upstream` first.

kenyon (approved):

PR #1650, opened 2025-08-06
Change proxy_redirect parameter to Optional[Variant[Array[String],Str…

rgarifullin-bank131:
Several proxy_redirect directives can be specified on the same level

### DIFF
--- a/templates/conf.d/nginx.conf.erb
+++ b/templates/conf.d/nginx.conf.erb
@@ -184,8 +184,8 @@ http {
 <% if @client_body_buffer_size -%>
   client_body_buffer_size <%= @client_body_buffer_size %>;
 <% end -%>
-<% if @proxy_redirect -%>
-  proxy_redirect          <%= @proxy_redirect %>;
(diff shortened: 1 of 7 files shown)


templates/conf.d/nginx.conf.erb
```diff
@@ -184,7 +184,11 @@ http {
...
   client_body_buffer_size <%= @client_body_buffer_size %>;
 <% end -%>
-<% if @proxy_redirect -%>
+<% if @proxy_redirect.is_a?(Array) -%>
+<% @proxy_redirect.each do |value| -%>
+  proxy_redirect          <%= value %>;
+<% end -%>
+<% elsif @proxy_redirect -%>
   proxy_redirect          <%= @proxy_redirect %>;
 <% end -%>
```
jay7x, line 193: Should be able to simplify this by casting 
```
<% Array(@proxy_redirect).each do |value| -%>
  proxy_redirect ...
<% end -%>
```

PR #1672, opened 2026-01-19
Add `reload` param to allow not reloading nginx

### LINKED ISSUE #1166: Feature request:  Make service notify configurable

- Puppet: 5.3.3-1
- Ruby: -
- Distribution: Debian Jessie
- Module version: 0.9.0

change a config (e.g. an upstream member) and see the puppet run log.

## What are you seeing
Nginx service is reloading after config changes

## What behavior did you expect instead
The service notify parameter should be configurable. (enable / disable)
sometimes you don't want to auto refresh the prod config of the Nginx service.
e.g. we're auto-adding upstream members to our main proxy's (over puppet export / puppetdb).
but we don't want that the member change to be active / deployed, without any manual check, during the next puppet run.

## Output log
` Notice: /Stage[main]/Nginx::Service/Service[nginx]: Would have triggered 'refresh' from 1 event`

bwitt:

Add `reload` boolean so that you can opt to not reload nginx on config changes.
Defaults to true.

Fixes https://github.com/voxpupuli/puppet-nginx/issues/1166


manifests/init.pp
```diff
@@ -416,6 +419,7 @@
   Optional[String] $service_restart = undef,
   String $service_name = 'nginx',
   Boolean $service_manage = true,
+  Variant[Type, Undef] $reload = Class['nginx::service'],
```
smortex, line 422: That looks wrong as IIRC, passing `undef` to a parameter with a default value will use the default value in the end.

I guess you can accept an array of classes to notify that has reasonable defaults but can be set to an empty array?


```suggestion
  Array[Type] $reload = [Class['nginx::service']],
```

manifests/init.pp
```diff
@@ -458,6 +462,10 @@
...
   # and preserve the relationship to the implementation classes through
   # a transitive relationship to the composite class.
-  Class['nginx::package'] -> Class['nginx::config'] ~> Class['nginx::service']
-  Class['nginx::package'] ~> Class['nginx::service']
+  if $reload {
+    Class['nginx::package'] -> Class['nginx::config'] ~> Class['nginx::service']
+    Class['nginx::package'] ~> Class['nginx::service']
+  } else {
+    Class['nginx::package'] -> Class['nginx::config'] -> Class['nginx::service']
+  }
```
smortex, line 470: Can't we also use the variable here:


```suggestion
  Class['nginx::package'] -> Class['nginx::config'] ~> $nginx::reload
  Class['nginx::package'] ~> $nginx::reload
```

smortex (changes requested):
See inline suggestions, but I think this is not working as expected the way it is now.

I feel like I already saw such pattern but can't remember where ATM.  It may make sense to use consistent parameter names for this pattern if multiple modules implement it. @neomilium does it ring  a bell, was it in one of our modules?

An alternative pattern we already implement in some vox modules is having a `manage_<someting>` parameter (where `<something>` is usually one of `package`, `service`, or `repo`), wrap the relevant resources in the `<module>::<something>` class with `if $module::manage_someting { ... }` and require / notify the class normally.

Last (but not least), changing configuration that mandate a service restart but skipping it is weird, each time I see something like it, it is needed for dubious reason (e.g. "my service will crash if I restart mysql"). The use case for the original issue feel dubious (don't use configuration management to orchestrate changes), do you have a practical use case for it?

PR #1703, opened 2026-06-17
Fix default IPv6 listen_options clash

### LINKED ISSUE #1293: Duplicate listen options when using IPv6 with www. redirect

- Puppet: 4.10.10
- Ruby: 2.1.9p490 (2016-03-30 revision 54437) [x86_64-linux]
- Distribution: Ubuntu Server 16.04
- Module version: 0.15.0

```puppet
include nginx

nginx::resource::server { 'example.com':
  ipv6_enable            => true,
  rewrite_www_to_non_www => true,
}
```

## What are you seeing

Nginx fails to load after the puppet run due to duplicate listen options. The `default` option is added to the ipv6 listen directive in both the `example.com` server block and the `www.example.com` server block.

The `ipv6only` and `default` options can only be set once, including them in both server blocks will break.

## Output log
```
Error: /Stage[main]/Nginx::Service/Service[nginx]/ensure: change from stopped to running failed: Systemd start for nginx failed!
journalctl log for nginx:
-- Logs begin at Tue 2018-12-25 23:13:37 CST, end at Wed 2018-12-26 10:09:21 CST. --
Dec 26 10:07:39 linuxdaemon-desktop systemd[1]: Starting nginx - high performance web server...
Dec 26 10:07:39 linuxdaemon-desktop nginx[5663]: nginx: [emerg] duplicate listen options for [::]:80 in /etc/nginx/sites-enabled/example.com.conf:12
Dec 26 10:07:39 linuxdaemon-desktop systemd[1]: nginx.service: Control process exited, code=exited status=1
Dec 26 10:07:39 linuxdaemon-desktop systemd[1]: Failed to start nginx - high performance web server.
Dec 26 10:07:39 linuxdaemon-desktop systemd[1]: nginx.service: Unit entered failed state.
Dec 26 10:07:39 linuxdaemon-desktop systemd[1]: nginx.service: Failed with result 'exit-code'.
Dec 26 10:08:02 linuxdaemon-desktop systemd[1]: Starting nginx - high performance web server...
Dec 26 10:08:02 linuxdaemon-desktop nginx[6296]: nginx: [emerg] a duplicate default server for [::]:80 in /etc/nginx/sites-enabled/example.com.conf:12
Dec 26 10:08:02 linuxdaemon-desktop systemd[1]: nginx.service: Control process exited, code=exited status=1
Dec 26 10:08:02 linuxdaemon-desktop systemd[1]: Failed to start nginx - high performance web server.
Dec 26 10:08:02 linuxdaemon-desktop systemd[1]: nginx.service: Unit entered failed state.
Dec 26 10:08:02 linuxdaemon-desktop systemd[1]: nginx.service: Failed with result 'exit-code'.
Dec 26 10:09:13 linuxdaemon-desktop systemd[1]: Starting nginx - high performance web server...
Dec 26 10:09:13 linuxdaemon-desktop nginx[7008]: nginx: [emerg] duplicate listen options for [::]:80 in /etc/nginx/sites-enabled/example.com.conf:12
Dec 26 10:09:13 linuxdaemon-desktop systemd[1]: nginx.service: Control process exited, code=exited status=1
Dec 26 10:09:13 linuxdaemon-desktop systemd[1]: Failed to start nginx - high performance web server.
Dec 26 10:09:13 linuxdaemon-desktop systemd[1]: nginx.service: Unit entered failed state.
Dec 26 10:09:13 linuxdaemon-desktop systemd[1]: nginx.service: Failed with result 'exit-code'.
Dec 26 10:09:21 linuxdaemon-desktop systemd[1]: Starting nginx - high performance web server...
Dec 26 10:09:21 linuxdaemon-desktop nginx[7619]: nginx: [emerg] duplicate listen options for [::]:80 in /etc/nginx/sites-enabled/example.com.conf:12
Dec 26 10:09:21 linuxdaemon-desktop systemd[1]: nginx.service: Control process exited, code=exited status=1
Dec 26 10:09:21 linuxdaemon-desktop systemd[1]: Failed to start nginx - high performance web server.
Dec 26 10:09:21 linuxdaemon-desktop systemd[1]: nginx.service: Unit entered failed state.
Dec 26 10:09:21 linuxdaemon-desktop systemd[1]: nginx.service: Failed with result 'exit-code'.

Notice: Applied catalog in 0.96 seconds
```

`ipv6only` and `default` listen options can only be set once according to nginx documentation.

### LINKED ISSUE #1700: `ipv6_listen_options` forces ipv6only on all vhosts
When setting up multiple vhosts/servers with IPv6 on a device with `listen [::]:80` it prevents nginx from starting because of duplicate entries in the listen section:

`nginx: [emerg] duplicate listen options for [::]:80 in /etc/nginx/sites-enabled/default-redirect.conf`

Config snippet:
```
server {
  listen *:80 default_server;
  listen [::]:80 default_server ipv6only=on;
```

This seems related to #1293. Where a workaround has been suggested by using `ipv6_listen_options => ''`. Which used to work and allow deployments to run. However since change #1670 it is forced to fill in a string because of the `String[1]`, preventing the earlier workaround from working.

It is also something mentioned in the nginx docs to only be set once on start.

smortex:
When not explicicly set, `$ipv6_listen_options` defaults to `$listen_options` with `"ipv6only=on"` appended to it.

The nginx documentation gives two important pieces of information regarding the `ipv6only` parameter:

1. "This parameter is turned on by default";
2. "It can only be set once on start".

Source: https://nginx.org/en/docs/http/ngx_http_core_module.html#listen

Because we explicitly set this parameter on each Virtual Host, this leads to broken configuration if multiple virtual hosts are configured.

While one could previously explicitly set `ipv6_listen_options => ''` to avoid this, the data type was changed in 4ad7bc0e4508368f667460be3220d558bec850c7 an empty String is not allowed anymore, making the module unusable in such a situation.

This commit adjust the module to not add `ipv6only=on` by default anymore, relying on the default behavior of nginx to achieve the same result.

It also allows to pass an explicit empty String as `$ipv6_listen_options` to avoid the default value which match `$listen_options`.

Last, it adjust the templates to avoid unexpeted spaces in the generated config.

Fixes: #1293
Fixes: #1700
Overcomes: #1701

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -2965,13 +2965,11 @@ Default value: `$listen_port`
 
 ##### <a name="-nginx--resource--mailhost--ipv6_listen_options"></a>`ipv6_listen_options`
 
-Data type: `Optional[String[1]]`
+Data type: `Optional[String[0]]`
 
-Extra options for listen directive like 'default' to catchall. Defaults to
-'ipv6only=on'. If listen_options is set, those options are inherited with
-'ipv6only=on' appended.
+Extra options for listen directive like 'default' to catchall.
 
-Default value: `undef`
+Default value: `$listen_options`
 
 ##### <a name="-nginx--resource--mailhost--ssl"></a>`ssl`
 
@@ -3726,13 +3724,11 @@ Default value: `$listen_port`
 
 ##### <a name="-nginx--resource--server--ipv6_listen_options"></a>`ipv6_listen_options`
 
-Data type: `Optional[String[1]]`
+Data type: `Optional[String[0]]`
 
-Extra options for listen directive like 'default' to catchall. Defaults to
-'ipv6only=on'. If listen_options is set, those options are inherited with
-'ipv6only=on' appended.
+Extra options for listen directive like 'default' to catchall.
 
-Default value: `undef`
+Default value: `$listen_options`
 
 ##### <a name="-nginx--resource--server--add_header"></a>`add_header`
 
@@ -4972,13 +4968,11 @@ Default value: `$listen_port`
 
 ##### <a name="-nginx--resource--streamhost--ipv6_listen_options"></a>`ipv6_listen_options`
 
-Data type: `Optional[String[1]]`
+Data type: `Optional[String[0]]`
 
-Extra options for listen directive like 'default' to catchall. Defaults to
-'ipv6only=on'. If listen_options is set, those options are inherited with
-'ipv6only=on' appended.
+Extra options for listen directive like 'default' to catchall.
 
-Default value: `undef`
+Default value: `$listen_options`
 
 ##### <a name="-nginx--resource--streamhost--proxy"></a>`proxy`
 
--- a/manifests/resource/mailhost.pp
+++ b/manifests/resource/mailhost.pp
@@ -17,9 +17,7 @@
 # @param ipv6_listen_port
 #   Default IPv6 Port for NGINX to listen with this server on.
 # @param ipv6_listen_options
-#   Extra options for listen directive like 'default' to catchall. Defaults to
-#   'ipv6only=on'. If listen_options is set, those options are inherited with
-#   'ipv6only=on' appended.
+#   Extra options for listen directive like 'default' to catchall.
 # @param ssl
 #   Indicates whether to setup SSL bindings for this mailhost.
 # @param ssl_cert
@@ -139,7 +137,7 @@ define nginx::resource::mailhost (
   Boolean $ipv6_enable = false,
   Variant[Array[String], String] $ipv6_listen_ip = '::',
   Stdlib::Port $ipv6_listen_port = $listen_port,
-  Optional[String[1]] $ipv6_listen_options = undef,
+  Optional[String[0]] $ipv6_listen_options = $listen_options,
   Boolean $ssl = false,
   Optional[String] $ssl_cert = undef,
   Optional[String] $ssl_ciphers = $nginx::ssl_ciphers,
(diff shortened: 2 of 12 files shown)


jay7x:
Looks like a breaking change 🤔

smortex:
> Looks like a breaking change 🤔

I really consider this a bugfix: the removed parameter was set to its default value, but that value can only be set a single time for a host/port in the configuration.  It was bad (and has been bad for a long time), but a workaround was possible until recently, and we cannot use this workaround anymore.  So a proper fix was required :grin: 

But maybe I am missing something, and I don't mind to do a major release after merging this if it makes sense.  Can you elaborate about what makes you think this should be considered a breaking change @jay7x ?

jay7x:
I'm looking at changes in test files and that looks like behaviour is changed. I had no look at the rest of the code around though.. does it affect existing setups with default settings?

smortex:
> does it affect existing setups with default settings

Yes, it fix them* :wink:  When updating to the latest release, I _had_ to remove the workaround `ipv6_listen_options => ''` because it was not allowed anymore, and as a result nginx refused to start because `ipv6only=on` was set multiple times for the same address+port :scream:.
With this change, the workaround is not needed anymore and can be safely removed, but if kept this does not change anything for the user (in fact, I am not 100% sure that allowing `ipv6_listen_options => ''` makes sense, but it reverts the backward compatibility with older versions of the module, that is the reason I allowed it).


As the removed config is the default one, and cannot be repeated, it feels better to me to not explicitly add it: if the users _need_ to set it to a specific value, they _will_ set it to that value.


*: I do have a bunch of nginx servers with really basic configuration, in reality I believe that this change is very safe, but my use case are quite simple and for this reason I cannot be 100% affirmative that it will not cause any kind of trouble with fancy configurations.

jay7x:
I see.. actually, existing installations was my main concern. Read it as "shall we release this change as major or minor release" :)

jay7x (approved):

kenyon (approved):

